PR #26, opened 2022-01-17
Add support for Flax/Jax models

versae:
This PR adds support for benchmarking/finetuning models in Flax/Jax format (marked with the "JAX" tag in the 🤗 /Hub) via on-the-fly conversion to PyTorch. In the future, Huggingface might add native Jax support in the Trainer, for now conversion is the quickest way to use Flax/Jax models

### DIFF
--- a/scandeval/benchmarks/abstract/base.py
+++ b/scandeval/benchmarks/abstract/base.py
@@ -251,11 +251,16 @@ class BaseBenchmark(ABC):
                 task = model_metadata['task']
 
         # Ensure that the framework is installed
+        from_flax = False
         try:
-            if framework == 'pytorch':
+            if framework in ('pytorch', 'jax'):
                 import torch
                 import torch.nn as nn
                 from torch.nn import Parameter
+                if framework == 'jax':
+                    from_flax = True
+                    import jax
+                    framework = 'pytorch'
             elif framework == 'spacy':
                 import spacy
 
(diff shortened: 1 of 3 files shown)


scandeval/benchmarks/abstract/base.py
```diff
@@ -791,15 +797,15 @@ def benchmark(self,
...
 
-        if framework in ['pytorch']:
+        if framework in ['pytorch', 'jax']:
 
             # Set platform-dependent random seeds
             if framework == 'pytorch':
                 import torch
                 torch.manual_seed(4242)
                 torch.cuda.manual_seed_all(4242)
                 torch.backends.cudnn.benchmark = False
```
saattrupdan: We still need to set random seeds if `framework=='jax'`, so this shouldn't be conditional on the `framework` variable.

setup.py
```diff
@@ -3,6 +3,7 @@
 from bump_version import get_current_version
 
 PYTORCH_REQUIRES = ['torch>=1.6.0']
+JAX_REQUIRES = PYTORCH_REQUIRES + ['jaxlib>=0.1.75', 'jax>=0.2.26', 'flax>=0.3.6']
 SPACY_REQUIRES = ['spacy>=3.2.0', 'spacy-transformers>=1.1.0']
 ALL_REQUIRES = (PYTORCH_REQUIRES +
                 SPACY_REQUIRES)
```
saattrupdan: `ALL_REQUIRES` should also include `JAX_REQUIRES`.

saattrupdan (changes requested):
Thanks for the PR, appreciate it!

Looks great. I've marked a couple of minor comments.

Furthermore, can you add that Jax is now supported to the readme, at the place where frameworks are mentioned? And can you update the `_load_model` docstring as well? Lastly, can you change the destination branch to `dev` instead of `main`, as I'm running package versioning from `dev`. I'll update the package (and thus merge it into `main`) after this merge.

scandeval/benchmarks/abstract/base.py
versae: That's right! Fixed.

setup.py
versae: Added.

PR #27, opened 2022-01-20
Add support for model revisions

versae:
This PR add the option of specifying a model revision to download tokenizer, config, and model files from. It is optional, but it can be specified by appending a suffix `@` followed by a commit hash, tag name, or branch name:
```bash
$ scandeval -m versae/roberta-base-ncc-512@2d735b9ff9a28c36170ea0a2f950d9b4714fdad5 -d norne-nb
```
It defaults to `main` and it only works for Hugging Face models.


saattrupdan:
Amazing, looks great! I'll update the package straight away 👌

PR #53, opened 2023-03-27
option to fix embedding / ability to override model config

peter-sk:
These changes add two optional parameters to scandeval. These parameters allow to use popular GPT-2-based models, as well as to provide local models (directories resulting from save_pretrained) that are not available on HF Hub (yet).

--fix-embedding
    Whether to increase the size of the model embedding to the size of the
    tokenizer vocabulary, if needed.

--override-model-config
    Override model config from HF Hub with revision, framework, task and
    languages. This parameter is expected to be a dictionary mapping
    a string defining the model id to a dictionary mapping revision, framework,
    and task to appropriate strings and languages to a list of strings
    with language codes.

### DIFF
--- a/src/scandeval/benchmark_config_factory.py
+++ b/src/scandeval/benchmark_config_factory.py
@@ -32,7 +32,7 @@ def build_benchmark_config(
         model_language (None, str or sequence of str, optional):
             The language codes of the languages to include for models. If specified
             then this overrides the `language` parameter for model languages.
-        model_framework (None, str, optional):
+        model_framework (str, optional):
             The model framework to use. Only relevant if `model-id` refers to a local path.
             Otherwise, the framework will be set automatically.
         dataset_language (None, str or sequence of str, optional):
--- a/src/scandeval/benchmark_config_factory.py
+++ b/src/scandeval/benchmark_config_factory.py
@@ -20,6 +20,8 @@ def build_benchmark_config(
     progress_bar: bool,
     save_results: bool,
     verbose: bool,
+    fix_embedding: bool,
+    override_model_config: Optional[dict],
 ) -> BenchmarkConfig:
     """Create a benchmark configuration.
 
@@ -58,6 +60,15 @@ def build_benchmark_config(
             'scandeval_benchmark_results.json'.
         verbose (bool, optional):
             Whether to output additional output.
+        fix_embedding (bool, optional):
+            Whether to increase the size of the model embedding to the size of the
+            tokenizer vocabulary, if needed.
+        override_model_config (dict, optional):
+            Override model config from HF Hub with revision, framework, task and
+            languages. This parameter is expected to be a dictionary mapping
+            a string defining the model id to a dictionary mapping revision, framework,
+            and task to appropriate strings and languages to a list of strings
+            with language codes.
     """
     # Prepare the languages
     languages = prepare_languages(language=language)
@@ -90,6 +101,8 @@ def build_benchmark_config(
         progress_bar=progress_bar,
         save_results=save_results,
         verbose=verbose,
+        fix_embedding=fix_embedding,
+        override_model_config=override_model_config,
     )
 
 
--- a/src/scandeval/benchmark_config_factory.py
+++ b/src/scandeval/benchmark_config_factory.py
@@ -32,6 +32,9 @@ def build_benchmark_config(
         model_language (None, str or sequence of str, optional):
             The language codes of the languages to include for models. If specified
             then this overrides the `language` parameter for model languages.
+        model_framework (None, str, optional):
+            The model framework to use. Only relevant if `model-id` refers to a local path.
+            Otherwise, the framework will be set automatically.
         dataset_language (None, str or sequence of str, optional):
             The language codes of the languages to include for datasets. If
             specified then this overrides the `language` parameter for dataset
@@ -59,15 +62,6 @@ def build_benchmark_config(
             'scandeval_benchmark_results.json'.
         verbose (bool, optional):
             Whether to output additional output.
-        fix_embedding (bool, optional):
-            Whether to increase the size of the model embedding to the size of the
-            tokenizer vocabulary, if needed.
-        override_model_config (dict, optional):
-            Override model config from HF Hub with revision, framework, task and
-            languages. This parameter is expected to be a dictionary mapping
-            a string defining the model id to a dictionary mapping revision, framework,
-            and task to appropriate strings and languages to a list of strings
-            with language codes.
     """
     # Prepare the languages
     languages = prepare_languages(language=language)
--- a/src/scandeval/benchmark_config_factory.py
+++ b/src/scandeval/benchmark_config_factory.py
@@ -20,8 +20,7 @@ def build_benchmark_config(
     progress_bar: bool,
     save_results: bool,
     verbose: bool,
-    fix_embedding: bool,
-    override_model_config: Optional[dict],
+    model_framework: Optional[str],
 ) -> BenchmarkConfig:
     """Create a benchmark configuration.
 
@@ -101,8 +100,7 @@ def build_benchmark_config(
         progress_bar=progress_bar,
         save_results=save_results,
         verbose=verbose,
-        fix_embedding=fix_embedding,
-        override_model_config=override_model_config,
+        model_framework=model_framework,
     )
 
 
--- a/src/scandeval/benchmark_dataset.py
+++ b/src/scandeval/benchmark_dataset.py
@@ -93,7 +93,7 @@ def get_local_model_config(
                                    "local model `{model_id}`! "
                                    "Please use the --model-framework option.")
         logger.info(
-            f"Assuming 'pytroch' as the framework for local model `{model_id}`! "
+            f"Assuming 'pytorch' as the framework for local model `{model_id}`! "
             "If this is in error, please use the --model-framework option to override."
         )
         framework = "pytorch"
--- a/src/scandeval/benchmark_dataset.py
+++ b/src/scandeval/benchmark_dataset.py
@@ -482,7 +482,7 @@ class BenchmarkDataset(ABC):
                     from_flax=model_config.framework == "jax",
                     use_auth_token=self.benchmark_config.use_auth_token,
                     cache_dir=self.benchmark_config.cache_dir,
-                    fix_embedding=elf.benchmark_config.fix_embedding,
+                    fix_embedding=self.benchmark_config.fix_embedding,
                 )
 
             # Initialise compute_metrics function
--- a/src/scandeval/benchmark_dataset.py
+++ b/src/scandeval/benchmark_dataset.py
@@ -1,6 +1,7 @@
 """Abstract benchmarking dataset class."""
 
 import logging
+import os
 import warnings
 from abc import ABC, abstractmethod
 from collections import defaultdict
@@ -45,6 +46,15 @@ from .utils import (
 
 # Set up logger
 logger = logging.getLogger(__name__)
+
+def _local_model_config(model_id: str, benchmark_config: BenchmarkConfig):
+    framework = benchmark_config.model_framework
+    if framework is None:
+        if benchmark_config.raise_errors:
+            raise RuntimeError(f"No framework specified for local model `{model_id}`!")
+        framework = "pytorch"
+    model_config = ModelConfig(model_id=model_id, revision="local", framework=framework, task="unknown", languages=[])
+    return model_config
 
 
 class BenchmarkDataset(ABC):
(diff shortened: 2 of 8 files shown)


saattrupdan:
Hi @peter-sk and thanks for the PR!

First comment I have is about the choice of flags. ScandEval should be as uniform as possible and *not* flexible, to ensure consistent evaluations of models (akin to the mindset of `black`, for instance). Therefore, I don't see why `fix-embedding` shouldn't always be true, for instance. Do you have a use case where this would be a bad idea?

As for the `override-model-config`, the revision can already be selected using "model-id@revision-hash" as the model ID, as specified in the readme. Do you have a use case where it is relevant to override the config? Because ideally, I think it would be more streamlined if it just does these things automatically, and not assume that the user applied the correct flags, if that makes sense.

peter-sk:
Hi @saattrupdan and thanks for the comments 👍 

Regarding fix-embedding, I do not see why the default should not be True. I could either remove the flag or make the reverse flag "--no-fix-embedding" - any preferences from your side? Personally, I think it is a behaviour that it might be nice to be able to switch off (and get an error message instead).

Regarding override-model-config, the use case is models that are not (yet) on HF Hub but only available as a local folder. I am open to finding a better way than providing a full model config. Basically, I want to be able to evaluate a model on downstream tasks WHILE I train it - without having to upload a new model to the HF Hub each time.

peter-sk:
I updated the pull request to make fix embedding the default behaviour.

src/scandeval/model_loading.py
```diff
@@ -405,10 +406,13 @@ def align_model_and_tokenizer(
     # the vocab size according to the model, we raise an error
     if hasattr(model.config, "vocab_size") and hasattr(tokenizer, "vocab_size"):
         if model.config.vocab_size < tokenizer.vocab_size:
-            raise InvalidBenchmark(
-                "The vocab size of the tokenizer is larger than the vocab size of the "
-                "model. This is not supported."
-            )
+            if fix_embedding:
+                model.resize_token_embeddings(tokenizer.vocab_size+1)
```
saattrupdan:
```suggestion
                model.resize_token_embeddings(new_num_tokens=tokenizer.vocab_size+1)
```

src/scandeval/cli.py
```diff
@@ -147,6 +147,25 @@
...
     with scores lying in the 'scandeval_benchmark_results.jsonl' file.""",
 )
+@click.option(
+    "--fix-embedding/--no-fix-embedding",
+    "-f/-nf",
+    default=True,
+    show_default=True,
+    help="""Whether to increase the size of the model embedding to the size of the
+            tokenizer vocabulary, if needed."""
+)
```
saattrupdan: I'm not sure what the benefit is of having this as a flag. Do you foresee any downsides of having this always enabled? If not then I think it's more in the opinionated spirit of scandeval to remove this flag and always fix the embeddings.

src/scandeval/model_loading.py
```diff
@@ -405,10 +406,13 @@ def align_model_and_tokenizer(
...
+            else:
+                raise InvalidBenchmark(
```
saattrupdan: If `fix_embedding` is removed then we could use the `raise_errors` flag here, and raise this error instead of fixing the embeddings if `--raise-errors` is set.

src/scandeval/cli.py
```diff
@@ -147,6 +147,25 @@
...
+    "--override-model-config",
+    "-o",
+    default=None,
+    show_default=True,
+    help="""Override model config from HF Hub with revision, framework, task and
+            languages. This parameter is expected to be a dictionary mapping
+            a string defining the model id to a dictionary mapping revision, framework,
+            and task to appropriate strings and languages to a list of strings
+            with language codes."""
+)
```
saattrupdan: Maybe this should be renamed to `--model-config`. Also, I think it would be a bit more intuitive to use `type=(str, str)` and `multiple=True` as [described here](https://stackoverflow.com/a/72108720/3154226). This would allow the following:

```
scandeval --model-config framework pytorch
```

saattrupdan: Also, if this is purely for being able to use `scandeval` with local models, I don't think it's necessary to specify `task`, `language` or even `revision`? So maybe this option is not needed at all?

saattrupdan (changes requested):
I've added a few comments, where I'd like to discuss if we need these flags at all, or if the issues could be solved automatically.

src/scandeval/cli.py
peter-sk: I don't think there is any downside. Let's not have a flag :-)

src/scandeval/model_loading.py
peter-sk: Sounds like a great idea. Default behaviour is to just continue in the best way possible. --raise-errors stops us on issues - I like the philosophy!

src/scandeval/cli.py
peter-sk: True. Language can be set separately. I'll update the documentation for --language when I update the pull request. Same goes for task. Revision could be "local" - what do you think?

src/scandeval/cli.py
peter-sk: The flag has been removed.

src/scandeval/model_loading.py
peter-sk: Done.

src/scandeval/cli.py
peter-sk: Ok, after reviewing the code, I found out that the languages and task attributes of modelconfig are not used at all. I just set them to [] and "unknown" for local models.

peter-sk:
I made a new pull request that is much more minimal. The only option added is "--model--framework" aka "-mf", which allows specifying "pytorch" or "jax" for local models to avoid an error when "--raise-errors" is set.

Given the use of framework right now is limited to a special case for "jax", this option will have limited utility.

In general, the ModelConfig class carries information that is never used (or, as far as I can tell, dumped to the results file). This might be something to refactor out - in a different PR :-)

peter-sk:
One minor idea: the default option for the framework could be "pytorch", i.e., it would only need to be specified for local non-pytorch models.

src/scandeval/cli.py
```diff
@@ -57,6 +57,16 @@
     specified. If "all" then all models will be benchmarked. If not specified then this
     will use the `language` value.""",
 )
+@click.option(
+    "--model-framework",
+    "-mf",
+    default=None,
+    show_default=True,
+    multiple=True,
```
saattrupdan: I don't see how multiple frameworks make sense here?

src/scandeval/benchmark_dataset.py
```diff
@@ -46,6 +47,15 @@
 # Set up logger
 logger = logging.getLogger(__name__)
 
+def _local_model_config(model_id: str, benchmark_config: BenchmarkConfig):
+    framework = benchmark_config.model_framework
+    if framework is None:
+        if benchmark_config.raise_errors:
+            raise RuntimeError(f"No framework specified for local model `{model_id}`!")
+        framework = "pytorch"
```
saattrupdan: Maybe we could do something more intelligent if `framework` is `None`. For instance, PyTorch models usually end in `.bin` and JAX models end in `.msgpack`. Maybe we could check the directory for existence of such?

src/scandeval/model_loading.py
```diff
@@ -351,7 +352,7 @@ def setup_model_for_question_answering(model: PreTrainedModel) -> PreTrainedMode
 
 
 def align_model_and_tokenizer(
-    model: PreTrainedModel, tokenizer: PreTrainedTokenizer
+    model: PreTrainedModel, tokenizer: PreTrainedTokenizer, raise_errors: bool = False
```
saattrupdan, line 357: Needs to include `raise_errors` in the docstring.

src/scandeval/model_loading.py
```diff
@@ -37,6 +37,7 @@ def load_model(
     from_flax: bool,
     use_auth_token: Union[bool, str],
     cache_dir: str,
+    raise_errors: bool,
```
saattrupdan: Needs to include `raise_errors` in the docstring.

src/scandeval/benchmark_dataset.py
```diff
@@ -46,6 +47,15 @@
...
+    model_config = ModelConfig(model_id=model_id, revision="local", framework=framework, task="unknown", languages=[])
```
saattrupdan: Just using the same defaults as for fresh models here:

```suggestion
    model_config = ModelConfig(model_id=model_id, revision="main", framework=framework, task="fill-mask", languages=[])
```

src/scandeval/benchmark_dataset.py
```diff
@@ -46,6 +47,15 @@
 # Set up logger
 logger = logging.getLogger(__name__)
 
+def _local_model_config(model_id: str, benchmark_config: BenchmarkConfig):
```
saattrupdan:
```suggestion
def get_local_model_config(model_id: str, benchmark_config: BenchmarkConfig) -> ModelConfig:
```

saattrupdan: Also, missing a docstring.

src/scandeval/config.py
```diff
@@ -81,6 +81,8 @@ class BenchmarkConfig:
     Attributes:
         model_languages (sequence of Language objects):
             The languages of the models to benchmark.
+        model_framework (str):
+            The framework of the models to benchmark.
```
saattrupdan:
```suggestion
        model_framework (str or None, optional):
            The framework of the models to benchmark. If None then the framework will be inferred.
            Defaults to None.
```

src/scandeval/config.py
```diff
@@ -121,6 +123,7 @@ class BenchmarkConfig:
     save_results: bool
     verbose: bool
     testing: bool = False
+    model_framework: str = None
```
saattrupdan:
```suggestion
    model_framework: Optional[str] = None
```

src/scandeval/benchmark_config_factory.py
```diff
@@ -31,6 +32,9 @@ def build_benchmark_config(
         model_language (None, str or sequence of str, optional):
             The language codes of the languages to include for models. If specified
             then this overrides the `language` parameter for model languages.
+        model_framework (None, str, optional):
+            The model framework to use. Only relevant if `model-id` refers to a local path.
+            Otherwise, the framework will be set automatically.
```
saattrupdan: This assumes that `model_framework` has a default value of `None`, which is not the case?

src/scandeval/benchmark_dataset.py
```diff
@@ -113,7 +123,8 @@ def benchmark(  # noqa
                 If the extracted framework is not recognized.
         """
         # Fetch the model config
-        model_config = get_model_config(
+        model_configurator = _local_model_config if os.path.isdir(model_id) else get_model_config
```
saattrupdan:
```suggestion
        model_configurator = get_local_model_config if os.path.isdir(model_id) else get_model_config
```

saattrupdan (changes requested):
Looks much better! I have a handful of suggested changes, many of them minor 🙂

src/scandeval/cli.py
peter-sk: Ok. Removed the multiple=True.

src/scandeval/model_loading.py
peter-sk: Done.

src/scandeval/benchmark_dataset.py
peter-sk: Done.

src/scandeval/benchmark_dataset.py
peter-sk: Done.

src/scandeval/benchmark_dataset.py
peter-sk: I made some basic guesses based on file extensions. This should work for most models.

src/scandeval/benchmark_dataset.py
peter-sk: Done.

src/scandeval/benchmark_config_factory.py
peter-sk: Done.

src/scandeval/benchmark_dataset.py
peter-sk: framework = benchmark_config.model_framework
    if framework is None:
        try:
            exts = {f.split('.')[-1] for f in os.listdir(model_id)}
            if "bin" in exts:
                framework = "pytorch"
            elif "msgpack" in exts:
                framework = "jax"
            elif "whl" in exts:
                raise InvalidBenchmark("SpaCy models are not supported.")
            elif "h5" in exts:
                raise InvalidBenchmark("TensorFlow/Keras models are not supported.")
        except OSError as e:
            logger.info(
                f"Cannot list files for local model `{model_id}`!"
            )
            if benchmark_config.raise_errors:
                raise e
    if framework is None:
        if benchmark_config.raise_errors:
            raise InvalidBenchmark(f"Cannot recognize framework automatically for "
                                   "local model `{model_id}`! "
                                   "Please use the --model-framework option.")
        logger.info(
            f"Assuming 'pytorch' as the framework for local model `{model_id}`! "
            "If this is in error, please use the --model-framework option to override."
        )
        framework = "pytorch"

peter-sk:
I made all the suggested fixes and doc strings. I also added a basic extension-based recognition of frameworks.

saattrupdan (approved):
Looks great! I also tested it with a local model, and it works wonders. Thanks for the contribution @peter-sk 🎉

PR #55, opened 2023-04-27
left-padding-is-broken in tokenizers

peter-sk:
question answering does not work with left padding due to an upstream issue in the fast tokenizers
[tokenizers issue #1229](https://github.com/huggingface/tokenizers/issues/1229)
this workaround should prevent scandeval from encountering this issue


peter-sk:
Following my bug report, left padding has been fixed in the tokenizers library.

PR #60, opened 2023-07-12
feat: Allows parallel benchmarks

### LINKED ISSUE #59: Cache interference issues when testing two models at once
Seems like the cache systems interferes with other running sessions of scandeval as it seems to delete the checkpoint of another runs.
Thus does not allow for workflow like:
```
CUDA_VISIBLE_DEVICES=1 scandeval --model-id MODEL1 --language da --raise-errors
# on a separate instance:
CUDA_VISIBLE_DEVICES=2 scandeval --model-id MODEL2 --language da --raise-errors
```

Naturally I could use multiple `--model-id` flags, but I wanted them to run each on their own GPUs.

A simple solution is to naturally do it in different folders and thus create two scandeval caches (which is what I am doing now).
However I wanted to suggest to nest the cache folder such that the structure is:

```
.scandeval_cache/model_id/checkpoint-270
```

as opposed to:

```
.scandeval_cache/checkpoint-270
```

I might have missed other important considerations

KennethEnevoldsen:
Previously it was not possible due to cache interference. 

Note that I have kept til cache to the base cache dir for datasets and metrics. Though it might be cleaner to nest them into their own folders so instead of e.g. having `.scandeval_cache/f1` you have ``.scandeval_cache/metrics/f1`.

Fixes #59

Edit: The tests fail, but it seems like it is an issue with the CI?


src/scandeval/utils.py
```diff
@@ -21,6 +21,21 @@
 from .exceptions import InvalidBenchmark
 
 
+def create_model_cache_dir(cache_dir: str, model_id: str) -> str:
+    """Create cache directory for a model.
+
+    Args:
+        cache_dir: The cache directory.
+        model_id: The model ID.
```
KennethEnevoldsen: Intentionally didn't add typing here to avoid documenting the typing twice. This also avoids cases where edits in the future can lead to disagreements on the type. I can add it in if you wish.

saattrupdan: Just had a re-read of Google's formatting guidelines, and you're absolutely right - types are only required in doc strings if the information isn't included as type hints. All good, and maybe I'll do a cleanup of the rest of the code base in the near future 🙂

src/scandeval/utils.py
```diff
@@ -21,6 +21,21 @@
...
+
+    Returns:
+        The path to the cache directory.
+    """
+    # to avoid nesting due to models name containing '/'
+    _model_id = model_id.replace("/", "--")
+    return os.path.join(cache_dir, "model_cache", _model_id)
```
saattrupdan: For the sake of consistency in the code base, can you change this to using `Path` objects instead?

src/scandeval/benchmark_dataset.py
```diff
@@ -523,6 +532,9 @@ def _benchmark_single_iteration(
                 `train` and `test`. If an exception is raised, then the exception is
                 returned.
         """
+        model_cache_dir = create_model_cache_dir(
+            cache_dir=self.benchmark_config.cache_dir, model_id=model_config.model_id
+        )
```
saattrupdan: Instead of re-initialising this here, we could just include the model cache dir as an argument to benchmark single, just like you did with get training args

src/scandeval/utils.py
saattrupdan: For the sake of consistency.

```suggestion
        cache_dir: 
            The cache directory.
        model_id: 
            The model ID.
```

saattrupdan (changes requested):
Looks good! Just got a few small changes. The CI error is mysterious, I'll look into that.

src/scandeval/utils.py
KennethEnevoldsen: It is also a lot easier to maintain 👌

saattrupdan:
Also, as you might have seen, I've opened a PR into your branch which fixes the CI bug:
https://github.com/KennethEnevoldsen/ScandEval/pull/1

src/scandeval/benchmark_dataset.py
KennethEnevoldsen: used the output dir from the training args

KennethEnevoldsen:
Perfect - I have merged it in. Everything should work as intended now.

saattrupdan (approved):
LGTM! Feel free to merge 🙂

PR #66, opened 2023-10-28
Feat/enable multi gpu

saattrupdan:
This PR does the following:

1. Sets up the package for multi-GPU benchmarking, where only the main process is logging. A guide to how to run scandeval with multiple GPUs is included in the readme. Basically, you simply replace `scandeval` with `accelerate launch -m scandeval` when benchmarking. This is possible due to the added `__main__.py` module. If you want to benchmark very large models it is beneficial to also use `deepspeed` - simply `pip install deepspeed` and call `accelerate launch --use_deepspeed --zero_rank 3 -m scandeval` instead of `scandeval` in this case.
2. When not in a distributed setting, only the first GPU will be used. This is mentioned to the user, so that they're able to swap to the `accelerate` setup instead, if needed.

This closes https://github.com/saattrupdan/ScandEval/issues/37 and depends on #69 and #70.


saattrupdan:
@peter-sk noted that the multi-GPU functionality doesn't yet work for decoder-models, and he's correct! Turned the PR into a draft and will look into this.

PR #68, opened 2023-11-27
enable mistral

peter-sk:
mistral needs flash attention and a decent dtype to be finetuned on a 40GB GPU


peter-sk:
I agree that --use-flash-attention-2 would be the optimal way. Where would you pass it? Is it part of the model config? Or where would it go?

PR #69, opened 2023-11-28
Fix/do not use auto batch size

saattrupdan:
Previously the `auto_find_batch_size` parameter was used, which reduces batch size by 2x when CUDA OOM is encountered. The problem is that the corresponding `gradient_accumulation` parameter doesn't get doubled in this case, so we revert back to doing this manually instead.


KennethEnevoldsen (approved):
Looks good

PR #70, opened 2023-11-28
feat: Add 8-bit AdamW optimizer

saattrupdan:
This changes the optimizer from the standard AdamW one to the 8-bit equivalent. This is only available when CUDA is available, so we choose the optimizer based on this.


CHANGELOG.md
```diff
@@ -40,6 +40,8 @@ and this project adheres to
 - Now uses the new `auto_find_batch_size` argument in `TrainingArguments`, rather than
   manually doing this, as the underlying `accelerate` built-in version is more robust.
 - Now reports all model parameters, rather than just the trainable ones.
+- Now uses 8-bit AdamW optimizer when CUDA is available rather than the default AdamW,
```
KennethEnevoldsen, line 55: Does this influence smaller models negatively?

KennethEnevoldsen: (if yes then it might be worth looking into differentiating based on memory usage).

saattrupdan: I'll run some experiments to check 🙂

saattrupdan: Ran some experiments, testing the models `Maltehb/aelaectra-danish-electra-small-cased` and `KennethEnevoldsen/dfm-sentence-encoder-large-1` on the `angry-tweets` dataset.

The smaller model achieved 32.19% ± 2.92% with 8-bit AdamW and 32.56% ± 3.23% without, and the larger model achieved 53.69% ± 1.49% with 8-bit AdamW and 53.74% ± 1.41% without. So, all in all, there's no real performance difference, and I think we can include this without it significantly affecting any of the scores.

KennethEnevoldsen: Solid! should definitely just use that in the future then!

KennethEnevoldsen (approved):

PR #71, opened 2023-11-28
Add NLG benchmark support

saattrupdan:
This PR adds the following:

1. Adds new general tasks for generative models. These are:
  a. `TEXT_TO_TEXT`, being tasks where the output is pure text, such as abstractive summarisation, abstractive question answering and translation. The `text_to_text` module has also been added for this task.
  b. `MULTIPLE_CHOICE`, being tasks where the output has to be one out of several text completions. No module has been added for this yet - I'll do that in another PR.
  c. `RAW_TEXT`, where the model is asked to reproduce text (these are evaluated using perplexity). No module has been added for this yet - I'll do that in another PR.
2. Adds the NordjyllandNews as an example of a `TEXT_TO_TEXT` dataset, as well as its builder script.
3. Move `GenerativeModel` and `Tokenizer` to a separate `protocols` module, to avoid circular imports.
4. Remove part of the Github CI where `torch` was forced to be version `2.0.0`. This has been fixed in recent `torch` versions, so this is not needed anymore.

This closes https://github.com/saattrupdan/ScandEval/issues/42


src/scandeval/dataset_tasks.py
```diff
@@ -110,6 +110,103 @@ def get_all_dataset_tasks() -> dict[str, DatasetTask]:
...
+    supertask="text-to-text",
+    metrics=[
+        MetricConfig(
+            name="bertscore",
+            pretty_name="BERTScore",
+            huggingface_id="bertscore",
+            results_key="f1",
+            postprocessing_fn=lambda raw_score: (raw_score, f"{raw_score:.2%}"),
+            compute_kwargs=dict(model_type="microsoft/mdeberta-v3-base", device="cpu"),
+        ),
```
KennethEnevoldsen, line 146: would probably rewrite to (in the following as well):
```suggestion
    postprocessing_fn=lambda raw_score: (raw_score, f"{raw_score:.2%}"),
    metrics=[
        MetricConfig(
            name="bertscore",
            pretty_name="BERTScore",
            huggingface_id="bertscore",
            results_key="f1",
            postprocessing_fn= postprocessing_fn,
            compute_kwargs=dict(model_type="microsoft/mdeberta-v3-base", device="cpu"),
        ),
```

src/scandeval/dataset_tasks.py
```diff
@@ -110,6 +110,103 @@ def get_all_dataset_tasks() -> dict[str, DatasetTask]:
...
+            huggingface_id="character",
+            results_key="cer_score",
+            postprocessing_fn=lambda raw_score: (raw_score, f"{raw_score:.2%}"),
+        ),
+    ],
+    labels=[],
+)
+
+
+RAW_TEXT = DatasetTask(
```
KennethEnevoldsen: Hmm raw_text seems odd. PERPLEXITY, CLM, CAUSAL_LM, AUTOREGRESSIVE_LM, PREDICTIVE_LM might be better (causal always seems a bit odd as there is no causality).

src/scandeval/text_to_text.py
```diff
@@ -0,0 +1,236 @@
...
+        special_token_metadata = get_special_token_metadata(tokenizer=tokenizer)
+        has_cls_token = special_token_metadata["has_cls_token"]
+        has_sep_token = special_token_metadata["has_sep_token"]
+        cls_token = special_token_metadata["cls_token"]
+        sep_token = special_token_metadata["sep_token"]
+
+        def tokenise(examples: dict) -> BatchEncoding:
+            # If the tokenizer is not adding special tokens, then we add them manually.
+            # We don't need this when performing few-shot evaluations, so in that case
+            # we don't add the special tokens.
```
KennethEnevoldsen: Hmm, couldn't we imagine that some models don't require special tokens (e.g. no sep token)?

src/scandeval/text_to_text.py
```diff
@@ -0,0 +1,236 @@
...
+    ):
+        """Load the data collator used to prepare samples during finetuning.
+
+        Args:
+            tokenizer:
+                A pretrained tokenizer. Can be None if the tokenizer is not used in the
+                initialisation of the data collator. Defaults to None.
+            model:
+                A pretrained model. Can be None if the model is not used in the
+                initialisation of the data collator. Defaults to None.
```
KennethEnevoldsen: I would just specify that it is not used.

KennethEnevoldsen: assuming it is there due to an ABC

src/scandeval/text_to_text.py
```diff
@@ -0,0 +1,236 @@
...
+        model_outputs_and_labels: tuple[np.ndarray, np.ndarray],
+        id2label: list[str],
+    ) -> dict[str, float]:
+        """Compute the metrics needed for evaluation.
+
+        Args:
+            model_outputs_and_labels:
+                The first sequence contains the model outputs and the second sequence
+                contains the true labels.
+            id2label (list of str):
```
KennethEnevoldsen:
```suggestion
            id2label:
```

src/scandeval/text_to_text.py
```diff
@@ -0,0 +1,236 @@
...
+                initialisation of the data collator. Defaults to None.
+
+        Returns:
+            The data collator.
+        """
+        return DataCollatorWithPadding(tokenizer, padding="longest")
+
+    def _compute_metrics(
+        self,
+        model_outputs_and_labels: tuple[np.ndarray, np.ndarray],
```
KennethEnevoldsen: You could add a type hint as tuple[Predictions, Labels]

Where you define Predictions and Labels as ndarrays. I find that it helps me keep track of what is what.

src/scandeval/text_to_text.py
```diff
@@ -0,0 +1,236 @@
...
+                Conversion of indices to labels.
+
+        Returns:
+            A dictionary with the names of the metrics as keys and the metric values as
+            values.
+        """
+        model_outputs, labels = model_outputs_and_labels
+
+        model_output_dtype = np.asarray(model_outputs).dtype
+        if model_output_dtype in [np.float16, np.float32, np.float64]:
```
KennethEnevoldsen:
```suggestion
        output_is_prob = model_output_dtype in [np.float16, np.float32, np.float64]
        if output_is_prob:
```
I assume this is why

src/scandeval/text_to_text.py
```diff
@@ -0,0 +1,236 @@
...
+            predictions = np.asarray(model_outputs).argmax(axis=-1)
+        else:
+            predictions = model_outputs
+
+        results: dict[str, float] = dict()
+        for cfg in self.dataset_config.task.metrics:
+            metric = self._metrics[cfg.name]
+            score_dict: dict[str, float] | None = metric.compute(
```
KennethEnevoldsen, line 108: When would metric return None?

src/scandeval/text_to_text.py
```diff
@@ -0,0 +1,236 @@
...
+        """
+        shuffled_train = train_dataset.shuffle(seed=random_seed)
+        num_few_shots = self.dataset_config.num_few_shot_examples
+        few_shot_examples: list[dict[str, Any]] = list()
+        while len(few_shot_examples) < num_few_shots:
+            example = shuffled_train.select(range(1))[0]
+            few_shot_examples.append(example)
+            shuffled_train = shuffled_train.filter(
+                lambda x: x["text"] != example["text"]
+            )
```
KennethEnevoldsen, line 150: Any reason not to do `shuffled_train.select(range(num_few_shots))`?

src/scandeval/text_to_text.py
```diff
@@ -0,0 +1,236 @@
...
+                **cfg.compute_kwargs,
+            )
+            if score_dict is not None:
+                scores = score_dict[cfg.results_key]
+                if isinstance(scores, list):
+                    scores = sum(scores) / len(scores)
+                results[cfg.name] = scores
+        return results
+
+    def _extract_few_shot_examples(
```
KennethEnevoldsen, line 123: We could do better using "similar" documents (though that is model dependent) so I would suspect that this is the best naive approach.

src/scandeval/text_to_text.py
```diff
@@ -0,0 +1,236 @@
...
+            example = shuffled_train.select(range(1))[0]
+            few_shot_examples.append(example)
+            shuffled_train = shuffled_train.filter(
+                lambda x: x["text"] != example["text"]
+            )
+        random.seed(random_seed)
+        random.shuffle(few_shot_examples)
+        return few_shot_examples
+
+    def _apply_few_shot_prompt(
```
KennethEnevoldsen, line 156: Generally, with a few of these things, I would suspect it is model-dependent. Don't think it is worth restructuring anything now, but believe you might already have had some issues with this in the other benchmarks.

I tried to solve it in SEB by creating a Model object that contains all the assumptions about the model.

src/scandeval/text_to_text.py
```diff
@@ -0,0 +1,236 @@
...
+            The examples with the few-shot prompt applied.
+        """
+        # Build the few-shot part of the prompt
+        few_shot_prompts = [
+            self.dataset_config.prompt_template.format(
+                text=example["text"].replace("\n", " ").strip(),
+                target_text=example["target_text"].replace("\n", " ").strip(),
+            )
+            for example in few_shot_examples
+        ]
```
KennethEnevoldsen, line 177: Would probably make it a function:
```suggestion
        few_shot_prompts = [self.format_prompt(example["text"], example["target_text"])
        for example in few_shot_examples]
```

src/scandeval/text_to_text.py
```diff
@@ -0,0 +1,236 @@
...
+        Args:
+            examples:
+                The examples to apply the prompt to.
+            few_shot_examples:
+                The examples to be included in the few-shot prompt.
+
+        Returns:
+            The examples with the few-shot prompt applied.
+        """
+        # Build the few-shot part of the prompt
```
KennethEnevoldsen, line 170: would probably make it a function:

```
build_few_shot_prompt(...)
```

KennethEnevoldsen: Maybe that is what this function already is - maybe it is the apply_few_shot prompt that makes me suspect it is being forwarded through the model

KennethEnevoldsen (approved):
Looks good. A few minor things.

src/scandeval/dataset_tasks.py
saattrupdan: The `postprocessing_fn` is dependent on the metric though, as some metrics output numbers between 0 and 1, and others between 0 and 100 (and perplexity neither), so having a single postprocessing function doesn't quite work, as far as I can see.

src/scandeval/dataset_tasks.py
saattrupdan: Yeah I didn't quite like it either. I'm hesitant at calling it `PERPLEXITY`, since that's just a metric, and we might include other metrics that measure "knowledge about the given text". `CAUSAL_LM` and `AUTOREGRESSIVE_LM` can maybe be misleading, since (I'm guessing) it should be possible benchmark seq2seq models on these tasks as well. `PREDICTIVE_LM` could maybe work, but it doesn't really seem to capture what we're trying to measure.

Maybe something like `RAW_COMPREHENSION`? Or maybe simply `KNOWLEDGE`, albeit that's a bit vague...

src/scandeval/text_to_text.py
saattrupdan: That is indeed the case for some models, but IIRC some tasks need these tokens (I think it was the QA tasks), so I needed to include these. But for these generative-only tasks we might be fine leaving them out. I'll leave them out for now, then we can always add them back in in the future if necessary.

src/scandeval/text_to_text.py
saattrupdan: When the metric computation isn't run on the main process; i.e., only when multi-GPU is used. So it's not really relevant until multi-GPU support is in, but I've added a comment explaining this now in any case.

src/scandeval/text_to_text.py
saattrupdan: Yep, since we're bootstrapping, so there will be duplicates in the training dataset - this ensures that the few-shot examples are not duplicated. I've added a comment explaining this now.

src/scandeval/text_to_text.py
saattrupdan: Yeah it has to work with every model out there. And again, the point of ScandEval is to evaluate each model fairly and consistently, not to get the maximal performance out of each one

src/scandeval/text_to_text.py
saattrupdan: You mean the prompt template is model-dependent? Because this function simply concatenates the few-shot examples with the final example. The prompt templates are all located in the `dataset_configs` module.

src/scandeval/text_to_text.py
saattrupdan: But it's already a function, no? More specifically, the `self.dataset_config.prompt_template.format` function? It seems like all this does is renaming that function to `self.format_prompt`?

src/scandeval/text_to_text.py
saattrupdan: What do you mean by "forwarded through the model" here?

src/scandeval/dataset_tasks.py
KennethEnevoldsen: Understandable. `RAW_COMPREHENSION` is a bit misleading are we measuring comprehension? (or knowedge) Maybe `TEXT_MODELLING`?

src/scandeval/text_to_text.py
KennethEnevoldsen: Ahh perfect! makes sense

src/scandeval/dataset_tasks.py
saattrupdan: I like `TEXT_MODELLING`! Let's say that.

src/scandeval/text_to_text.py
KennethEnevoldsen: Def. could be (e.g. system prompts). It is the case in SEB:
https://github.com/KennethEnevoldsen/scandinavian-embedding-benchmark/issues/18

PR #74, opened 2023-11-29
Fix/overlap between splits in qa datasets

saattrupdan:
A bug have caused an overlap between the dataset splits in the ScandiQA datasets. This fixes that, and the corresponding evaluations have to be redone as well, unfortunately.


KennethEnevoldsen (approved):
Looks good!

PR #80, opened 2023-12-04
Enable Flash Attention 2 and rename TEXT_TO_TEXT task to SUMMARIZATION

saattrupdan:
This PR:

1. Adds a `--use-flash-attention` flag, which for instance allows benchmarking Mistral-based models. If the model has "mistral" in its name (case-insensitive) then this is automatically set.
2. Renames the `TEXT_TO_TEXT` task to `SUMMARIZATION`, while keeping 'text-to-text' as supertask. This makes it slightly more intuitive, e.g., `scandeval --dataset-task summarization -m gpt2`. If we add an abstractive question-answering task in the future it would also use the 'text-to-text' supertask, making it use the same evaluation script, but also allowing different metrics and few-shot prompts, for instance.

This PR replaces the PR #68

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,11 +2,15 @@
 name = "ScandEval"
 version = "8.0.0"
 description = "Evaluation of pretrained language models on mono- or multilingual language tasks."
-authors = ["Dan Saattrup Nielsen <saattrupdan@gmail.com>"]
+authors = ["Dan Saattrup Nielsen <dan.nielsen@alexandra.dk>"]
+maintainers = [
+    "Dan Saattrup Nielsen <dan.nielsen@alexandra.dk>",
+    "Kenneth Enevoldsen <kenneth.enevoldsen@cas.au.dk>",
+]
 readme = "README.md"
 license = "MIT"
 homepage = "https://scandeval.github.io"
-repository = "https://github.com/saattrupdan/ScandEval"
+repository = "https://github.com/ScandEval/ScandEval"
 
 [tool.poetry.dependencies]
 python = ">=3.10,<4.0"


src/scandeval/benchmarker.py
```diff
@@ -130,6 +132,7 @@ def __init__(
         verbose: bool = False,
         trust_remote_code: bool = False,
         load_in_4bit: bool | None = None,
+        use_flash_attention: bool = False,
```
KennethEnevoldsen, line 135: Any reason why this is not `True`

KennethEnevoldsen (approved):
Looks good - Naturally it would be best are FA2 was default, though I assume it is to avoid a bothersome additional dependency?

src/scandeval/benchmarker.py
saattrupdan: It's because it requires `flash-attn` to be built, which currently can only be installed via `pip` with the slightly hacky `--no-build-isolation` flag, and furthermore requires CUDA to be installed. Since there are so few models requiring it currently, I don't see the issue atm. As it gets more prevalent, and `flash-attn` gets easier to install more broadly, this might change for sure.

KennethEnevoldsen: I assumed as much. We might accidentally bias the benchmark by requiring flash attention for some models, but I if the package is problematic to build I would keep it as is. Pytorch has also integrated flash attention v1 so I would imagine that v2 will also become native at some point.

saattrupdan:
@KennethEnevoldsen I've also added you as a maintainer now, given all that you've helped with these PRs - is that okay with you, or would you prefer to stay informal?

KennethEnevoldsen:
Thanks @saattrupdan, that is perfectly fine with me

PR #81, opened 2023-12-12
Question answering bug

saattrupdan:
This fixes a bug causing an error when benchmarking some models on question answering datasets.

The error occurs when the answer is a proper subset of the first token in the context, and is fixed by simply using the entire first token when this occurs.

### DIFF
--- a/src/scandeval/question_answering.py
+++ b/src/scandeval/question_answering.py
@@ -420,9 +420,13 @@ def prepare_train_examples(
                 ):
                     token_start_index += 1
                 tokenized_examples.start_positions.append(token_start_index - 1)
-                while token_end_index > 0 and offsets[token_end_index][1] >= end_char:
+                while (
+                    token_end_index > token_start_index
+                    and offsets[token_end_index][1] >= end_char
+                ):
                     token_end_index -= 1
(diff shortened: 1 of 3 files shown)


src/scandeval/question_answering.py
```diff
@@ -420,7 +420,7 @@ def prepare_train_examples(
                 ):
                     token_start_index += 1
                 tokenized_examples.start_positions.append(token_start_index - 1)
-                while offsets[token_end_index][1] >= end_char:
+                while token_end_index > 0 and offsets[token_end_index][1] >= end_char:
```
KennethEnevoldsen: I am having a hard time judging whether this actually solves the problem - I assume that you already checked - but it might be good idea to write a test on such that we don't reintroduce the bug in the future?

KennethEnevoldsen (approved):

src/scandeval/question_answering.py
saattrupdan: Added a unit test now, which fails without the fix and is passed with it 🙂

KennethEnevoldsen: Wonderful exactly what I was looking for!

PR #82, opened 2023-12-12
Layernorm and fp16 bug

saattrupdan:
This fixes a bug where a few models encountered an error when benchmarking them on CUDA, due to their layernorm implementation not supporting fp16.

The error happens since `torch_dtype` is set when loading the model based on *either* the HuggingFace model configuration, if it has been specified in there, or the data type of the first weight of the model. The `torch_dtype` is then the data type that the model will be loaded in as. This is completely fine in the cases where `torch_dtype` is set in the model configuration, such as with the Mistral model, but we encounter the above issue when this is not set in the model configuration and the model has been stored in mixed precision.

The fix is simply to check if `torch_dtype` is set in the model configuration, and only setting it to "auto" during model initialisation if this is the case - otherwise we set it to None, which defaults to loading in float32.

Added a unit test to check for this.

### DIFF
--- a/tests/test_model_setups/test_hf.py
+++ b/tests/test_model_setups/test_hf.py
@@ -1,0 +1,31 @@
+"""Unit tests for the `model_setups.hf` module."""
+
+import pytest
+import torch
+from scandeval.model_setups.hf import HFModelSetup
+
+
+@pytest.mark.parametrize(
+    argnames=["model_id", "expected"],
+    argvalues=[
+        ("microsoft/mdeberta-v3-base", None),
+        ("jonfd/electra-small-nordic", None),
+        ("mistralai/Mistral-7B-v0.1", torch.bfloat16),
+    ],
+    ids=[
+        "mdeberta-v3-base",
+        "electra-small-nordic",
+        "mistral-7b",
+    ],
+)
+def test_torch_dtype_is_set_correctly(benchmark_config, model_id, expected):
+    model_setup = HFModelSetup(benchmark_config=benchmark_config)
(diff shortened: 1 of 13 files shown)


tests/test_model_setups/test_hf.py
```diff
@@ -0,0 +1,31 @@
...
+import torch
+from scandeval.model_setups.hf import HFModelSetup
+
+
+@pytest.mark.parametrize(
+    argnames=["model_id", "expected"],
+    argvalues=[
+        ("microsoft/mdeberta-v3-base", None),
+        ("jonfd/electra-small-nordic", None),
+        ("mistralai/Mistral-7B-v0.1", torch.bfloat16),
```
KennethEnevoldsen, line 13: Seems like quite large models to download for tests. I would probably:

```suggestion
        ("microsoft/deberta-v3-xsmall", None),
        ("jonfd/electra-small-nordic", None),
        ("mistralai/Mistral-7B-v0.1", torch.bfloat16),
```
Note it is not multilingual.

Is there a smaller alternative for the mistral (I couldn't find one). Probably a good point going forward though: To always make and release very small models for tests.

KennethEnevoldsen (approved):
Looks good!

tests/test_model_setups/test_hf.py
saattrupdan: It's only downloading the model configs, not the models, so it should be alright :)

KennethEnevoldsen: Ah perfect, then just keep it as is.

PR #83, opened 2023-12-12
Add new datasets

saattrupdan:
This PR adds the followings:

1. Adds 8 new datasets, 4 of which are German, 3 of which are Dutch and the last one being Icelandic. With this, almost all of the Germanic languages are supported in all the tasks. Icelandic is missing a sentiment dataset, Dutch is missing an extractive QA dataset, and Faroese is missing a sentiment, extractive QA and summarisation dataset.
2. Fixed a couple of issues regarding sampling of few-shot examples.
3. Changed the unit tests of the benchmarks, to just not raise errors when evaluating, rather than having a particular score. This is changed since almost all scores were 0%, as I reduced the number of iterations to train for to 2, to speed up unit testing.
4. Added `isort` sorting of imports to `ruff`, which is why there are so many files changed 😅


src/scandeval/named_entity_recognition.py
```diff
@@ -145,12 +145,35 @@ def _compute_metrics(
...
+        predictions_all_zero = all(
+            all(ner_tag == "o" for ner_tag in prediction_list)
+            for prediction_list in predictions
         )
-        results_no_misc = self._metrics["micro_f1_no_misc"].compute(
-            predictions=predictions_no_misc, references=labels_no_misc
+        labels_all_zero = all(
+            all(ner_tag == "o" for ner_tag in label_list) for label_list in labels
         )
+        if predictions_all_zero and labels_all_zero:
```
AJDERS, line 158: Maybe add a comment explaining that its to avoid errors from the micro_f1 metric.

AJDERS (approved):
Nitpicking. Otherwise LGTM!

PR #84, opened 2023-12-14
feat: Add English datasets

saattrupdan:
This PR adds the following English datasets:

1. `sst5`, a sentiment classification dataset, where "very positive" and "very negative" has been converted into "positive" and "negative"
2. `scala-en`, a linguistic acceptability dataset created just like the other ScaLA datasets, based on the [GUM universal dependencies treebank](https://github.com/UniversalDependencies/UD_English-GUM/tree/master)
3. `squad`, an extractive QA dataset based on the SQuAD-v2 dataset, where samples without answers have been removed, as with the other ScandEval QA datasets
4. `cnn-dailymail`, a summarisation dataset with summaries of news articles.

The reason for adding English datasets is to be able to distinguish NLU tasks from non-English tasks, to (for instance) see if generative models are bad at NLU tasks in general, or if they're just bad at non-English tasks.

### DIFF
--- a/src/scandeval/dataset_configs.py
+++ b/src/scandeval/dataset_configs.py
@@ -366,7 +366,7 @@ CONLL_EN_CONFIG = DatasetConfig(
     pretty_name="the truncated version of CoNLL 2003",
     huggingface_id="ScandEval/conll-en-mini",
     task=NER,
-    languages=[NL],
+    languages=[EN],
     prompt_prefix="Below are sentences and JSON dictionaries with the named "
     "entities that occur in the given sentence.",
     prompt_template="Sentence: {text}\nNamed entities: {label}",
--- a/src/scandeval/dataset_configs.py
+++ b/src/scandeval/dataset_configs.py
@@ -2,7 +2,7 @@
 
 from .config import DatasetConfig
 from .dataset_tasks import LA, NER, QA, SENT, SPEED, SUMMARIZATION
-from .languages import DA, DE, FO, IS, NB, NL, NN, SV, get_all_languages
+from .languages import DA, DE, EN, FO, IS, NB, NL, NN, SV, get_all_languages
 
 
 def get_all_dataset_configs() -> dict[str, DatasetConfig]:
@@ -59,7 +59,6 @@ SWEREC_CONFIG = DatasetConfig(
     max_generated_tokens=3,
 )
 
-
 ANGRY_TWEETS_CONFIG = DatasetConfig(
     name="angry-tweets",
     pretty_name="the truncated version of AngryTweets",
@@ -75,7 +74,6 @@ ANGRY_TWEETS_CONFIG = DatasetConfig(
     num_few_shot_examples=12,
     max_generated_tokens=3,
 )
-
 
 NOREC_CONFIG = DatasetConfig(
     name="norec",
@@ -93,7 +91,6 @@ NOREC_CONFIG = DatasetConfig(
     max_generated_tokens=3,
 )
 
-
 ISREC_CONFIG = DatasetConfig(
     name="isrec",
     pretty_name="the truncated version of IsReC",
@@ -109,7 +106,6 @@ ISREC_CONFIG = DatasetConfig(
     num_few_shot_examples=12,
     max_generated_tokens=3,
 )
-
 
 FOREC_CONFIG = DatasetConfig(
     name="forec",
@@ -127,7 +123,6 @@ FOREC_CONFIG = DatasetConfig(
     max_generated_tokens=3,
 )
 
-
 SB10K_CONFIG = DatasetConfig(
     name="sb10k",
     pretty_name="the truncated version of SB10k",
@@ -144,7 +139,6 @@ SB10K_CONFIG = DatasetConfig(
     max_generated_tokens=3,
 )
 
-
 DUTCH_SOCIAL_CONFIG = DatasetConfig(
     name="dutch-social",
     pretty_name="the truncated version of Dutch Social",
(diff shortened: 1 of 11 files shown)


KennethEnevoldsen (approved):
Looks very reasonable. I thought that English was also gonna be a part of the TrustLLM project?

src/scandeval/dataset_configs.py
```diff
@@ -355,6 +361,29 @@ def get_dataset_config(dataset_name: str) -> DatasetConfig:
     max_generated_tokens=128,
 )
 
+CONLL_EN_CONFIG = DatasetConfig(
+    name="conll-en",
+    pretty_name="the truncated version of CoNLL 2003",
+    huggingface_id="ScandEval/conll-en-mini",
+    task=NER,
+    languages=[NL],
```
KennethEnevoldsen:
```suggestion
    languages=[EN],
```

saattrupdan:
> Looks very reasonable. I thought that English was also gonna be a part of the TrustLLM project?
> 
> [Oh not done with the review still ongoing]

Ready for review! And English is not included in TrustLLM nopes.

src/scripts/create_cnn_dailymail.py
```diff
@@ -0,0 +1,65 @@
+"""Create the CNN-DailyMail-mini summarisation dataset."""
```
KennethEnevoldsen, line 1: For Scandinavian datasets there haven't been much of a discussion about which datasets to use and why, but it might be ideal to discuss it here where we have multiple datasets to choose from. We even have the option to balance the dataset across a wide variety of domains, which might give us a more representative estimate. 

CNN is fairly similar to the other summarization datasets (newslike), but are we actually interested in news summarization or summarization in general (I would say in general).

That being said It might be worth simply creating an issue and merging this PR as is.

KennethEnevoldsen: E.g. for DaNE it might also be an idea to shuffle in parts of DANSK.

src/scripts/create_sst5.py
```diff
@@ -0,0 +1,84 @@
+"""Create the SST5-mini sentiment dataset and upload it to the HF Hub."""
```
KennethEnevoldsen, line 1: sst5 vs sst2?

KennethEnevoldsen (commented):
Seems like there is a error on the language tag (NL -> EN) but otherwise it looks good. Would create an issue for discussing how to select/construct dataset. However right now I think it is very reasonable to simply choose reasonable datasets and continue from there.

saattrupdan:
> > And English is not included in TrustLLM nopes.
> 
> 
> 
> Oh really thought it was Germanic.

Yeah English is a weird language. It's partly Germanic, but most of its influence is from French and Latin, from the Normans. And a small part is from the native Old English, which isn't Germanic either.

src/scripts/create_sst5.py
saattrupdan: SST2 and SST5 is really the same dataset, but SST2 only has two labels (positive and negative). To maintain consistency with the other sentiment datasets I went for "SST3", being SST5 where we merge the "very positive/negative" with "positive/negative".

src/scripts/create_cnn_dailymail.py
saattrupdan: Yeah you're right that the choice is currently somewhat arbitrary (mostly chosen based on similarity of domains and format). 

They can definitely be changed. I would suggest that such a change will be done in one PR, which would result in a major change to the package. Also, the leaderboard would need to be redone after such a change (probably 1-3 months of GPU hours needed for this).

KennethEnevoldsen:
> > > And English is not included in TrustLLM nopes.
> > 
> > 
> > Oh really thought it was Germanic.
> 
> Yeah English is a weird language. It's partly Germanic, but most of its influence is from French and Latin, from the Normans. And a small part is from the native Old English, which isn't Germanic either.

Ahh right. Yea, it gets muddy. However, you are completely right you often want it as a point of comparison.

src/scripts/create_sst5.py
KennethEnevoldsen: Yea, all stem from the sentiment tree structure. 

Ahh right to make it more similar to the rest of the datasets.

src/scripts/create_cnn_dailymail.py
KennethEnevoldsen: > Yeah you're right that the choice is currently somewhat arbitrary (mostly chosen based on similarity of domains and format).

Exactly, but I think it would require more discussion than this PR allows. There is a tradeoff between making the two languages comparable (Should NE for Danish should be similar to NE for Norwegian or is the goal making the best estimator for NE for Danish).

> They can definitely be changed. I would suggest that such a change will be done in one PR, which would result in a major change to the package. Also, the leaderboard would need to be redone after such a change (probably 1-3 months of GPU hours needed for this).

Wouldn't you rather do it a dataset at a time? Incremental changes are easier to manage and it doesn't really change the compute time (you only need to rerun that specific dataset to update the benchmark). Of course, the man-hours might be a reason to do it differently.

saattrupdan: You're right that doing it one dataset at a time is easiest if compute is flexible. But if we have to benchmark big generative models with, e.g., Peter's compute, then I'm guessing it would be easier to do it all at once. But maybe I'm wrong here!

As for the goal of the benchmark, I'll create a separate discussion here on this repo for that 🙂

KennethEnevoldsen: @peter-sk ?

saattrupdan: Created a discussion here: https://github.com/ScandEval/ScandEval/discussions/85. I'll create separate discussions for the development of the new versions of (some of the) datasets.

saattrupdan: Here's a discussion regarding a new Danish NER dataset: https://github.com/ScandEval/ScandEval/discussions/86

PR #88, opened 2023-12-21
Feat/add model cache

saattrupdan:
### Added
- Now caches the completions of open source generative models, which effectively makes
  benchmarking of these 10x faster. We cannot store all logits for storage reasons (it
  quickly gets >100GB in that case), so we instead store the top-100 logits. We thus
  assume that (a) these are the only logits needed, and (b) that the generations don't
  change. We argue that (a) is the case since we only use the logits in classification
  tasks, in which case we only use the first token anyway. Further, since we're using
  a temperature of 0 anyway, the generations will be as close to deterministic as
  possible (up to small rounding fluctuations of logits, which is negligible). This is
  a breaking change, since it is not compatible with the previous way we cached OpenAI
  model outputs.
- Added a new `--clear-model-cache` flag, which removes the cached models after
  finishing the benchmarking of each model, to save disk space. This doesn't remove the


src/scandeval/generation.py
```diff
@@ -21,6 +25,15 @@
 logger = logging.getLogger(__package__)
 
 
+@dataclass
+class CachedModelOutput:
+    completion: list[int]
+    completion_str: str
```
KennethEnevoldsen: Any reason why we do not simply cache the string? If the temperature is 0 anyway?

src/scandeval/sequence_classification.py
```diff
@@ -298,7 +298,7 @@ def _extract_labels_from_generation(
 
 
 def get_closest_logprobs_labels(
-    generation_logprobs: tuple[torch.Tensor],
+    generation_logprobs: tuple[torch.Tensor] | torch.Tensor,
```
KennethEnevoldsen: This seems like something that should have been standardized before this function.

src/scandeval/generation.py
```diff
@@ -165,79 +192,215 @@ def generate_single_iteration(
...
     """
-    # Tokens used in generation to know when generation is finished
-    stopping_criteria = get_generation_stopping_criteria(
-        tokenizer=tokenizer, model=model
-    )
+    # Load the model output cache
+    with cache_path.open() as f:
+        json_cache = json.load(f)
+        cache: dict[str, dict[str, CachedModelOutput]] = defaultdict(dict)
+        if not hasattr(sys, "_called_from_test"):
```
KennethEnevoldsen: This seems a bit hacky way of not using the cache while testing. Any reason why you don't use a flag. An alternative would be to input the cache as a argument. E.g. have a ModelOutputCache object with a method like `get_cache()`. Then during testing you can pass a DummyCache (dependency injection).

src/scandeval/generation.py
```diff
@@ -165,79 +192,215 @@ def generate_single_iteration(
...
 
     Returns:
         A list of dictionaries containing the scores for each metric.
     """
-    # Tokens used in generation to know when generation is finished
-    stopping_criteria = get_generation_stopping_criteria(
-        tokenizer=tokenizer, model=model
-    )
+    # Load the model output cache
+    with cache_path.open() as f:
```
KennethEnevoldsen: Seems like the loading should be extracted from this method

src/scandeval/generation.py
```diff
@@ -165,79 +192,215 @@ def generate_single_iteration(
...
+        cache: dict[str, dict[str, CachedModelOutput]] = defaultdict(dict)
+        if not hasattr(sys, "_called_from_test"):
+            for key in json_cache:
+                for sub_key in json_cache[key]:
+                    cache[key][sub_key] = CachedModelOutput(**json_cache[key][sub_key])
+
+    # Initialise the cache for the current max tokens, if it doesn't exist already
+    max_tokens_str = str(dataset_config.max_generated_tokens)
+    if max_tokens_str not in cache:
+        cache[max_tokens_str] = dict()
```
KennethEnevoldsen: Not sure I get why it is created for the max token str?

src/scandeval/generation.py
```diff
@@ -165,79 +192,215 @@ def generate_single_iteration(
...
+    # Split up the prepared dataset into a cached and non-cached part
+    unique_non_cached_ids: set[int] = set()
+    for example_idx, example in enumerate(prepared_dataset):
+        cached_texts = list(cache[max_tokens_str].keys())
+        cached_texts += prepared_dataset.select(unique_non_cached_ids)["text"]
+        if example["text"] not in cached_texts:
+            unique_non_cached_ids.add(example_idx)
+    cached_ids = set(range(len(prepared_dataset))) - unique_non_cached_ids
+    non_cached_dataset = prepared_dataset.select(unique_non_cached_ids)
+    cached_dataset = prepared_dataset.select(cached_ids)
```
KennethEnevoldsen: Seems like it should be a function?

src/scandeval/generation.py
```diff
@@ -165,79 +192,215 @@ def generate_single_iteration(
...
+        # Generate the completions for the non-cached examples
+        for batch_idx, batch in enumerate(tqdm(dataloader, leave=False)):
+            # Generate the completions of the documents in the batch
+            with warnings.catch_warnings():
+                warnings.simplefilter("ignore", category=UserWarning)
+                inputs = batch["input_ids"].to(model.device)
+                with torch.inference_mode():
+                    model_output: ModelOutput = model.generate(
+                        inputs=inputs, generation_config=generation_config
+                    )
```
KennethEnevoldsen: I believe you can do:
```
            with (
              warnings.catch_warnings(),
              torch.inference_mode()
              ):
              ...
```
Due to:
https://docs.python.org/3.10/whatsnew/3.10.html#parenthesized-context-managers

src/scandeval/generation.py
```diff
@@ -21,6 +25,15 @@
 logger = logging.getLogger(__package__)
 
 
+@dataclass
+class CachedModelOutput:
```
KennethEnevoldsen: Probably rename it to GenerativeModelOutput()?

src/scandeval/generation.py
```diff
@@ -165,79 +192,215 @@ def generate_single_iteration(
...
+            with cache_path.open("w") as f:
+                dumpable_cache: dict[str, dict[str, dict]] = defaultdict(dict)
+                for key, value in cache.items():
+                    for sub_key, sub_value in value.items():
+                        dumpable_cache[key][sub_key] = asdict(sub_value)
+                json.dump(dumpable_cache, f, indent=4)
+                del dumpable_cache
+
+    # Fetch the cached predictions for the cached examples
+    if len(cached_dataset) > 0:
```
KennethEnevoldsen, line 314: I would also extract this out in a method

KennethEnevoldsen (commented):
General idea is great, added a few suggestions - notably for separating out the tests from the code

src/scandeval/generation.py
saattrupdan: It was simply if the same prompt had two different completions, with different amounts of max tokens. For instance, if the same sample is used for two different tasks (say, sentiment and NER).

src/scandeval/generation.py
saattrupdan: It's mostly to maintain consistency with the rest of the code base, since I need it in a handful of functions. Seems a bit simpler than juggling around different flags everywhere 🙂

saattrupdan: But I've refactored many of the code blocks into separate functions now, so the cache functionality can still be tested

src/scandeval/generation.py
saattrupdan: Good point. Removed now

KennethEnevoldsen (approved):
Thanks for the changes, I believe it is much more readable now. It seems like the PR is still being worked on with the tests.

src/scandeval/generation.py
saattrupdan: After reconsidering, I ended up removing it after all, since I completely forgot about the prompt creation, which includes things like "\nSentiment:" at the end, for instance. This means that if the same sample is used for two different tasks, then their inputs would be different as well anyway.

KennethEnevoldsen: Makes sense

Issue #93, opened 2023-12-24
Add HellaSwag


saattrupdan:
This will be part of a new common sense reasoning task, framed as a text classification task. Could be machine translated initially.

Issue #92, opened 2023-12-24
Add MMLU


saattrupdan:
This will constitute a new "knowledge task", and can be framed as a text classification task, making it suitable for both encoders and decoders. 

We can add a machine translated version of MMLU to start with. Also, we probably need a way to select a smaller subset of it.

saattrupdan:
@peterbjorgensen I have formatted and uploaded the `m_mmlu` dataset that you showed me - it's now available at [Hugging Face Hub](https://huggingface.co/datasets/alexandrainst/m_mmlu). Same thing with [HellaSwag](https://huggingface.co/datasets/alexandrainst/m_hellaswag), [ARC](https://huggingface.co/datasets/alexandrainst/m_arc) and [TruthfulQA](https://huggingface.co/datasets/alexandrainst/m_truthfulqa).

I've implemented MMLU, HellaSwag and ARC in the framework now, as a common new "multiple choice" task. Not sure if we should add TruthfulQA yet, but we can chat about that. I'll tag you in a PR in a minute.

In any case, we still need the translations that you're working on, for Norwegian, Icelandic and ideally also Faroese, if possible.

Issue #91, opened 2023-12-23
Add NorQuAD


saattrupdan:
The NorQuAD dataset is a gold-standard Norwegian question answering dataset. It should probably replace the ScandiQA-no dataset. 

Link: https://github.com/ltgoslo/NorQuAD

Issue #89, opened 2023-12-23
Add FoNE


saattrupdan:
The FoNE dataset is a gold-standard Faroese NER dataset, which should replace the current WikiANN-fo, as the latter is automatically generated and too easy. 

Link: https://huggingface.co/datasets/vesteinn/sosialurin-faroese-ner/discussions/1

PR #95, opened 2023-12-27
Feat/add more datasets

### LINKED ISSUE #94: Add ARC
This will be part of the "knowledge" task, similar to MMLU. Maybe it shouldn't be included if MMLU is sufficient.

saattrupdan:
### Added
- Added the following new datasets:
    - `fone`, a Faroese NER dataset, which replaces the previous `wikiann-fo` dataset.
    - `dansk`, a Danish NER dataset, which complements the previous `dane` dataset.
    - `norquad`, a Norwegian question answering dataset, which replaces the previous `scandiqa-no` dataset.
    - `mmlu-da`, `mmlu-sv`, `mmlu-de`,  and `mmlu-nl`, being GPT-3.5-turbo translated versions of MMLU, the multiple choice dataset testing the knowledge of generative models.
    - `arc-da`, `arc-sv`, `arc-de`,  and `arc-nl`, being GPT-3.5-turbo translated versions of HellaSwag, the multiple choice dataset testing the knowledge of grade-school science questions.
    - `hellaswag-da`, `hellaswag-sv`, `hellaswag-de`,  and `hellaswag-nl`, being GPT-3.5-turbo translated versions of HellaSwag, the multiple choice dataset testing the common sense reasoning of generative models.

This closes #94, #93, #92, #91 and #89.


CHANGELOG.md
```diff
@@ -23,13 +23,32 @@ and this project adheres to
...
   cached model outputs or datasets.
+- Added the following new datasets:
+    - `fone`, a Faroese NER dataset, which replaces the previous `wikiann-fo` dataset.
+    - `dansk`, a Danish NER dataset, which complements the previous `dane` dataset.
+    - `norquad`, a Norwegian question answering dataset, which replaces the previous
+      `scandiqa-no` dataset.
+    - `mmlu-da`, `mmlu-sv`, `mmlu-de`,  and `mmlu-nl`, being GPT-3.5-turbo translated
+      versions of MMLU, the multiple choice dataset testing the knowledge of generative
+      models.
+    - `arc-da`, `arc-sv`, `arc-de`,  and `arc-nl`, being GPT-3.5-turbo translated
```
KennethEnevoldsen: Is 3.5 turbo the best model for translation? I could imagine that this might introduce a bias, which is not ideal.

src/scripts/create_dansk.py
```diff
@@ -0,0 +1,176 @@
...
+        "NORP": "MISC",
+        "ORDINAL": "O",
+        "ORGANIZATION": "ORG",
+        "PERCENT": "O",
+        "PERSON": "PER",
+        "PRODUCT": "MISC",
+        "QUANTITY": "O",
+        "TIME": "O",
+        "WORK OF ART": "MISC",
+    }
```
KennethEnevoldsen, line 57: Why is it that we map to the conll2003 format? It seems simplifying without adding any benefits (in fact, differing from the 'standard' might be good at measuring weaknesses with LLMs, which perform poorly on multilevel classification, especially if there are a lot of labels).

src/scripts/create_fone.py
```diff
@@ -0,0 +1,114 @@
...
+        8: "O",  # Original: I-Date
+        9: "I-LOC",  # Original: I-Location
+        10: "I-MISC",  # Original: I-Miscellaneous
+        11: "O",  # Original: I-Money
+        12: "I-ORG",  # Original: I-Organization
+        13: "O",  # Original: I-Percent
+        14: "I-PER",  # Original: I-Person
+        15: "O",  # Original: I-Time
+        16: "O",  # Original: O
+    }
```
KennethEnevoldsen, line 52: same question as for dansk

CHANGELOG.md
saattrupdan: I completely agree regarding the bias. This was primarily to have _something_ to enable us to evaluate the models on these kinds of tasks, which should be replaced by gold standard datasets later.

And I didn't choose gpt-3.5, as we didn't do the translation - it was simply already freely available from the University of Oregon, so it was a start. 

We're currently looking into translation of the remaining Germanic languages (Norwegian, Icelandic and Faroese), as well as validation of the translations from native speakers. So see this as a start, and I think we should wait with a new ScandEval release until the translations are done and validated. 

Does that sound alright with you?

src/scripts/create_dansk.py
saattrupdan: Hmm a really good point actually. Initially it was to be able to compare across languages, but since it seems like we're moving away from that anyway, I don't see why we can't change this too. This is also the case for several other NER datasets in that case (Swedish, Norwegian and Faroese all use different tagging systems).

src/scripts/create_fone.py
saattrupdan: Moving the discussion to that comment 🙂

CHANGELOG.md
KennethEnevoldsen: Sounds very reasonable. I would add that justification somewhere (changelog maybe?) such that it is made clear somewhere

src/scripts/create_dansk.py
KennethEnevoldsen: Yea, I believe it is great to standardize e.g. when training a model but in our case keeping the original labels is probably better. Unsure if the prompts for NER are written specifically for conll2003?

saattrupdan: Also, I think the NER module needs to change to reflect this change, somehow. Currently the label scheme is hardcoded, and the label scheme is also used in the few shot prompts. So not sure how to deal with it

KennethEnevoldsen: I can't find a full overview of the prompt, but I would probably create the label section and then either hardcode it pr. dataset or add a description pr. label and then parse it in automatically. 

the label scheme should ideally not be hardcoded. Thought it might be ideal to commit this PR for now and just create an issue for this.

CHANGELOG.md
saattrupdan: Added it to the changelog now.

src/scripts/create_dansk.py
saattrupdan: Added issue on it now: https://github.com/ScandEval/ScandEval/issues/96

saattrupdan:
@KennethEnevoldsen Noticed there was an issue with the DANSK script, as it didn't allow for NER tags spanning multiple tokens - this has been fixed now.

AJDERS (approved):
LGTM!

KennethEnevoldsen:
> @KennethEnevoldsen Noticed there was an issue with the DANSK script, as it didn't allow for NER tags spanning multiple tokens - this has been fixed now.

The dataset or the code?

saattrupdan:
> > @KennethEnevoldsen Noticed there was an issue with the DANSK script, as it didn't allow for NER tags spanning multiple tokens - this has been fixed now.
> 
> The dataset or the code?

Both - I fixed both dataset and code. I removed all DANSK evaluations and started from scratch again after that fix.

KennethEnevoldsen:
Ahh so not an error in the original dataset

PR #101, opened 2024-01-02
Fix/update transformers

saattrupdan:
### Changed
- Now compatible with`transformers >= 4.36.2`, and this is required now as they have changed their generation API in a breaking manner.

### Fixed
- Now only sets `torch_dtype` is CUDA is available, as otherwise errors are caused.


AJDERS (approved):
LGTM!

KennethEnevoldsen (approved):

Issue #96, opened 2023-12-30
Allow different label schemes for each dataset


saattrupdan:
Currently a task has a set label scheme, and we enforce that each dataset with that task follows that label scheme. This involves things like mapping OntoNotes NER tags to CoNLL NER tags.

The point of this was originally to make the benchmarking on a task the same for each language, but as argued in [this discussion](https://github.com/ScandEval/ScandEval/discussions/85), this is less of a concern now. We would thus like to "free the labels" of all affected datasets, which involves (a) making it possible for the framework to handle different label schemes per dataset, and (b) to re-benchmark all the affected (model, dataset) combinations.

PR #102, opened 2024-01-02
100 add dbrd dataset

AJDERS:
This PR replaces the `dutch_social`-dataset with the [`dbrd`-dataset](https://huggingface.co/datasets/dbrd). 
Note that the `dbrd`-dataset is **binary and not trinary**, hence why the config is commented out for now. We can wait with merging this until #96 is implemented, if necessary.


src/scripts/create_dbrd.py
```diff
@@ -1,59 +1,57 @@
...
-    val_df["label"] = val_df["label"].map(label_mapping)
     test_df["label"] = test_df["label"].map(label_mapping)
 
-    # Create validation split
-    val_size = 256
-    val_df = val_df.sample(n=val_size, random_state=4242)
+    # Create new splits
+    _, train_df = train_test_split(
+        train_df, test_size=2048, random_state=703, stratify=train_df.label
+    )
```
saattrupdan: The other datasets have been split using a (1024, 256, 2048) split. Further, since the test set is not big enough for both the validation and test split, we can grab the validation split from the training set here:

```suggestion
    full_train_df, val_df = train_test_split(
        train_df, test_size=256, random_state=703, stratify=train_df.label
    )
     _, train_df = train_test_split(
        full_train_df, test_size=1024, random_state=703, stratify=full_train_df.label
    )
```

src/scripts/create_dbrd.py
```diff
@@ -1,59 +1,57 @@
...
 
-    # Create test split
-    test_size = 1024
-    test_df = test_df.sample(n=test_size, random_state=4242)
+    _, test_df = train_test_split(
+        test_df, test_size=1024, random_state=703, stratify=test_df.label
+    )
```
saattrupdan: The other datasets have been split using a (1024, 256, 2048) split:

```suggestion
    _, test_df = train_test_split(
        test_df, test_size=2048, random_state=703, stratify=test_df.label
    )
```

src/scripts/create_dbrd.py
```diff
@@ -1,59 +1,57 @@
...
 
-    # Create train split
-    train_size = 1024
-    train_df = train_df.sample(n=train_size, random_state=4242)
+    test_df, val_df = train_test_split(
+        test_df, test_size=512, random_state=703, stratify=test_df.label
+    )
```
saattrupdan: Removing since the validation split was defined in the earlier suggestion.

```suggestion
```

saattrupdan (changes requested):
Looks good! I've got a few small changes. And you need to replace Dutch Social with DBRD in `tests/test_sequence_classification.py` too (but comment it out until `DBRD_CONFIG` is in properly.

saattrupdan:
@AJDERS Also, please update the changelog with the new dataset 🙂

saattrupdan (commented):
Looks good, but since this PR needs #96 to work, we should probably wait until that has been implemented.

BramVanroy:
Out of curiosity, what was the problem with dutch-social?

saattrupdan:
> Out of curiosity, what was the problem with dutch-social?

Supposedly that the quality of the samples were not great, but I'm not 100% sure.

PR #109, opened 2024-01-04
feat: Add `--only-validation-split` flag

saattrupdan:
### Added
- Added `--only-validation-split` flag, which only benchmarks the model on the validation split, which is 5-10x smaller than the test split (depending on the dataset). This is especially useful with paid models like OpenAI models. The value of this flag is stored in the benchmark results, so this will be visible on leaderboards.


AJDERS (approved):
LGTM!

KennethEnevoldsen (approved):
LGTM!

PR #110, opened 2024-01-04
fix: Set ignore_mismatched_sizes if needed

saattrupdan:
### Fixed
- Now sets `ignore_mismatched_sizes` when loading models if the model cannot be loaded otherwise. This previously caused some issues when loading certain models.

### DIFF
--- a/src/scandeval/model_setups/hf.py
+++ b/src/scandeval/model_setups/hf.py
@@ -291,8 +291,8 @@ class HFModelSetup:
                                 "flash-attn`."
                             )
                     except (KeyError, RuntimeError) as e:
-                        if not ignore_mismatched_sizes:
-                            ignore_mismatched_sizes = True
+                        if not model_kwargs["ignore_mismatched_sizes"]:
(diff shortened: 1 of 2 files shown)


src/scandeval/model_setups/hf.py
```diff
@@ -291,8 +291,8 @@ def load_model(
                                 "flash-attn`."
                             )
                     except (KeyError, RuntimeError) as e:
-                        if not ignore_mismatched_sizes:
-                            ignore_mismatched_sizes = True
+                        if not model_kwargs["ignore_mismatched_sizes"]:
```
KennethEnevoldsen, line 294: Add a warning here

KennethEnevoldsen (approved):
I would add a warning otherwise fine

saattrupdan:
> I would add a warning otherwise fine

Changed the warning to a debug log, so that it shows if `--verbose` is set. This is because this will show in each of the 10 iterations, yielding a lot of repetitive output. Hope that sounds good to you.

PR #111, opened 2024-01-04
feat: DANSK replaces DaNE

saattrupdan:
The [DANSK dataset](https://openreview.net/forum?id=_ONCiIHe-wU) is a new Danish NER dataset, which has been more thoroughly annotated and which cover a more diverse amount of Danish text (from gigaword). A tentative performance correlation analysis on 50+ models shows that there is a ~93% correlation between the two benchmark datasets, meaning that it doesn't make sense keeping both.

When #96 gets implemented, this further allows for even more NER tags, making the task more challenging.

Screenshot 2024-01-04 at 15 51 10

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   cached model outputs or datasets.
 - Added the following new datasets:
     - `fone`, a Faroese NER dataset, which replaces the previous `wikiann-fo` dataset.
-    - `dansk`, a Danish NER dataset, which complements the previous `dane` dataset.
+    - `dansk`, a Danish NER dataset, which replaces the previous `dane` dataset.
     - `norquad`, a Norwegian question answering dataset, which replaces the previous
       `scandiqa-no` dataset.
     - Danish, Swedish, German and Dutch versions of the MMLU, ARC and HellaSwag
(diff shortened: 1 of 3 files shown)


KennethEnevoldsen:
Sounds very reasonable

CHANGELOG.md
```diff
@@ -24,7 +24,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   cached model outputs or datasets.
 - Added the following new datasets:
     - `fone`, a Faroese NER dataset, which replaces the previous `wikiann-fo` dataset.
-    - `dansk`, a Danish NER dataset, which complements the previous `dane` dataset.
+    - `dansk`, a Danish NER dataset, which replaces the previous `dane` dataset.
```
KennethEnevoldsen, line 27: Oh do we replace it dane? Hmm then that might be worth merging the two (dansk + dane_plus have the same labels). That however seems like a dataset thing. I have added and issue [here](https://huggingface.co/datasets/chcaa/DANSK/discussions/2)

KennethEnevoldsen (approved):
looks good

CHANGELOG.md
saattrupdan: Sure thing. I'm quite sure that it won't change the evaluations significantly anyway, given the high correlation, but if you merge them then feel free to open a PR with the new merged dataset 🙂

KennethEnevoldsen: No I think I agree. It is def. not the highest on my todo list for that reason.

PR #117, opened 2024-01-05
Fix/multiple choice format

saattrupdan:
This fixes a few things related to the multiple choice format of MMLU, ARC and HellaSwag:

1. Newlines are preserved, to separate the options properly.
2. Made the option letters lower case in the prompt, as the answers are automatically lower cased - this is to maintain consistency between the prompt and the label.


AJDERS (approved):
LGTM!

PR #123, opened 2024-01-05
fix: Store and load `validation_split` values correctly

saattrupdan:
There was an error relating to the saving and loading of the `validation_split` entry in the `scandeval_benchmark_results.jsonl` file., causing this attribute to not get logged. This is fixed here.


KennethEnevoldsen (approved):
looks good

PR #124, opened 2024-01-06
Minor fixes

saattrupdan:
This PR does the following:

1. Fixes bug regarding checking if a model has already been benchmarked. This didn't previously check for the `validation_split` argument. I.e., if you have benchmarked a model previously on the full test set, then you _have not_ benchmarked the same model on the validation split.
2. Adds a `few_shot` argument, which is only relevant for generative models. This currently does nothing, but anticipates the time when we add finetuning of generative models. We also check for this argument when looking for duplicates.
3. Adds `ruff` checking of docstrings, which includes fixing a tonne of docstrings throughout the project.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -97,8 +97,13 @@ exclude = '''
 
 [tool.ruff]
 target-version = "py311"
-extend-select = ["I"]
+extend-select = [
+    "I",
+    "D",
+]
 
+[tool.ruff.pydocstyle]
+convention = "google"
 
 [tool.ruff.extend-per-file-ignores]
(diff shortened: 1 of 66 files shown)


pyproject.toml
```diff
@@ -97,8 +97,13 @@ exclude = '''
 
 [tool.ruff]
```
KennethEnevoldsen, line 98: Good to see. On other project I have in fact replaced isort and black with ruff as well.

KennethEnevoldsen (approved):
Looks good. Of course always hard to see when there is changes to so many files (but that of course had to happen in this case)

PR #125, opened 2024-01-06
fix: NER tags were not correct in FoNE

saattrupdan:
For some reason the FoNE dataset had inconsistent NER tags: Some samples only had I-LOC tags without a beginning B-LOC, and so on. This fixes those samples. The dataset has been updated now and I'm not experiencing the same error anymore.

This fixes https://github.com/ScandEval/ScandEval/issues/118.

### DIFF
--- a/src/scripts/create_fone.py
+++ b/src/scripts/create_fone.py
@@ -53,6 +53,36 @@ def main():
         lambda ner_tags: [ner_conversion_dict[ner_tag] for ner_tag in ner_tags]
     )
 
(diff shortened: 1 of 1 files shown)


KennethEnevoldsen (approved):
> For some reason the FoNE dataset had inconsistent NER tags: Some samples only had I-LOC tags without a beginning B-LOC, and so on. This fixes those samples. The dataset has been updated now and I'm not experiencing the same error anymore.

Sounds like we should create an issue on the dataset repo as well on this. Otherwise the change looks good.

PR #126, opened 2024-01-08
Bug/flash attn

### LINKED ISSUE #119: Installation of flash attention logging
Currently, if flash attention is not built, then we are logging that the user should run `pip install flash-attn --no-build-isolation` to install it. However, this seems to lead to errors during installation, and the correct command seems to be the following:

`pip install -U wheel && FLASH_ATTENTION_SKIP_CUDA_BUILD=TRUE pip install flash-attn --no-build-isolation`

The log message probably be updated with this instead.

AJDERS:
This PR closes #119


saattrupdan (approved):
LGTM!

PR #129, opened 2024-01-09
fix: Ensure that samples are small enough

saattrupdan:
## Summary
Some of the samples in our datasets were excessively long or excessively short. Especially the long ones caused problems with OOM errors, without them adding anything significant to the dataset. Since this is not a long-context benchmark per se, I've removed the excessively long top-5% samples from datasets, as well as adjusted the number of few-shot examples to ensure that the samples are at most ~3000 tokens long.

### Changed
- Samples with excessively short or long texts have been removed.
- Adjusted number of few-shot examples in datasets to ensure that the resulting prompt is at most ~3000 tokens long.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   previously.
 - Now doesn't force ASCII characters in the NER task for generative models, making the
   target JSON dictionary more consistent with the input text.
+- Samples with excessively short or long texts have been removed, being contexts whose
+  lengths are in the top-5% or bottom-5%.
+- Adjusted number of few-shot examples in datasets to ensure that the resulting prompt
+  is at most ~3000 tokens long.
 
 ### Fixed
 - Removed `text2text-generation` temporarily from the tags defining generative models,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,8 +52,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   previously.
 - Now doesn't force ASCII characters in the NER task for generative models, making the
   target JSON dictionary more consistent with the input text.
-- Samples with excessively short or long texts have been removed, being contexts whose
-  lengths are in the top-5% or bottom-5%.
+- Samples with excessively short or long texts have been removed.
 - Adjusted number of few-shot examples in datasets to ensure that the resulting prompt
   is at most ~3000 tokens long.
 
--- a/src/scripts/constants.py
+++ b/src/scripts/constants.py
@@ -1,0 +1,13 @@
+"""Constants used in the dataset creation scripts."""
+
+
+# Bounds on the size of texts in question answering datasets
+MIN_NUM_CHARS_IN_CONTEXT = 30
+MAX_NUM_CHARS_IN_CONTEXT = 5000
+MIN_NUM_CHARS_IN_QUESTION = 10
+MAX_NUM_CHARS_IN_QUESTION = 100
+
+
+# Bounds on the size of texts in summarisation datasets
+MIN_NUM_CHARS_IN_ARTICLE = 30
+MAX_NUM_CHARS_IN_ARTICLE = 6000
--- a/src/scandeval/dataset_configs.py
+++ b/src/scandeval/dataset_configs.py
@@ -498,7 +498,8 @@ SCALA_EN_CONFIG = DatasetConfig(
     huggingface_id="ScandEval/scala-en",
     task=LA,
     languages=[EN],
-    prompt_prefix="The following are sentences and whether they are grammatically correct.",
+    prompt_prefix="The following are sentences and whether they are grammatically "
+    "correct.",
     prompt_template="Sentence: {text}\nGrammatically correct: {label}",
     prompt_label_mapping=dict(correct="yes", incorrect="no"),
     num_few_shot_examples=12,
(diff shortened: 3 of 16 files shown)


CHANGELOG.md
```diff
@@ -52,6 +52,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   previously.
 - Now doesn't force ASCII characters in the NER task for generative models, making the
   target JSON dictionary more consistent with the input text.
+- Samples with excessively short or long texts have been removed, being contexts whose
+  lengths are in the top-5% or bottom-5%.
```
KennethEnevoldsen: While I understand that this is easy to reason about in a text I would much rather have a clearer minimum word boundary and a maximum word boundary. Like 10-3 000 tokens?

KennethEnevoldsen:
Could be OOM error be avoided by lowering the batch sizes? As I understand most of the software for "serving" language models only does it in batches of size 1. 

Essentially I would much rather avoid filtering on length too much (as we know that it can be a meaningful differentiate between models). I would much rather have an NA due to OOM than limit the datasets capabilities to handle differentiate between models. (a model should never make the dataset easier)

saattrupdan:
> Could be OOM error be avoided by lowering the batch sizes? As I understand most of the software for "serving" language models only does it in batches of size 1.
> 
> Essentially I would much rather avoid filtering on length too much (as we know that it can be a meaningful differentiate between models). I would much rather have an NA due to OOM than limit the datasets capabilities to handle differentiate between models. (a model should never make the dataset easier)

One problem where this is a problem is that we want to include a prefix prompt before the few-shot examples in the prompt, since this helps bring pretrained decoders up on par with instruction tuned decoders. If the few-shot examples are too long then truncation will remove this prefix prompt, which artificially favours the finetuned models. Some of the samples are true outliers, some being 15x longer than the median - these are the ones I want to remove. Some are artificially short, too (some QA contexts are only ~5 tokens, literally just being the answer), so these are removed too.

We can thus see this PR as dataset cleaning rather than helping the models.

KennethEnevoldsen:
> We can thus see this PR as dataset cleaning rather than helping the models.

If we want to clean I would prefer an upper or lower limit on n-char as it is more clear what we consider too long or too short. The current method would still be removed from a clean dataset.

> If the few-shot examples are too long then truncation will remove this prefix prompt, which artificially favours the finetuned models

I don't think I agree with it. It is after all a benefit of fine-tuning. So obscuring that benefit I would say artificially unfavouring fine-tuned models. The same thing applies to instruction-tuned vs not (the benchmark should reflect the pros and cons of different methods).

CHANGELOG.md
saattrupdan: Sure, we can set them as explicit numbers instead. We should probably do it in terms of number of characters, like what is done here. I had a look at the resulting boundaries from the quantiles, and the upper bound on the QA context is around 3000 characters, and around 80 characters for the question. We could conservatively set it to 5000 and 100 characters, respectively? 

To give some examples of outliers: The ScandiQA datasets have samples that are >100,000 characters, and the SQuAD-nl dataset has questions that are >1000 characters (and which are not questions at all).

saattrupdan: As for the lower bound, we could set the minimum context to have 30 characters, and minimum question to have 10 characters, say?

KennethEnevoldsen: > SQuAD-nl
> ScandiQA

Seems like these should be fixed in the source dataset as well. Just remove them completely?

Otherwise I think these boundaries sounds great

saattrupdan:
> I don't think I agree with it. It is after all a benefit of fine-tuning. So obscuring that benefit I would say artificially unfavouring fine-tuned models. The same thing applies to instruction-tuned vs not (the benchmark should reflect the pros and cons of different methods).

If we stay "pure" regarding the purpose of this benchmark then we wouldn't include any finetuned models at all (instruction tuned models are finetuned, too). Of course we cannot fully do this, since we want to compare the performance against closed-source models, which are all finetuned. To remedy this we can do a compromise where we allow benchmarking finetuned models, but where we don't offer any kind of benefit to them. The purpose of the benchmark is still to measure _pre-training performance_, as we [discussed previously](https://github.com/ScandEval/ScandEval/discussions/85#discussioncomment-7852680).

KennethEnevoldsen:
> If we stay "pure" regarding the purpose of this benchmark then we wouldn't include any finetuned models at all (instruction tuned models are finetuned, too). Of course we cannot fully do this, since we want to compare the performance against closed-source models, which are all finetuned. To remedy this we can do a compromise where we allow benchmarking finetuned models, but where we don't offer any kind of benefit to them. The purpose of the benchmark is still to measure pre-training performance, as we https://github.com/ScandEval/ScandEval/discussions/85#discussioncomment-7852680.

Hmm right, I agree with that point, let us roll with this.

A few thoughts:

From a purely research perspective we are interested in some kind of measure (consider it a measure of a latent factor) for how good a model can become given a certain number of examples (Chollet's paper on "on the measure of intelligence" might be a good place to start). I.e. both the speed of learning and the upper bound are interesting.

From an industry POV we want to be able to choose which model to use for a certain application (in which case training on a few 10-1000 examples can make quite the difference, even the difference between 0-10 examples can be quite meaningful for the models ability to handle novel task independently).

src/scripts/constants.py
```diff
@@ -0,0 +1,13 @@
+"""Constants used in the dataset creation scripts."""
+
+
+# Bounds on the size of texts in question answering datasets
+MIN_NUM_CHARS_IN_CONTEXT = 30
+MAX_NUM_CHARS_IN_CONTEXT = 5000
+MIN_NUM_CHARS_IN_QUESTION = 10
+MAX_NUM_CHARS_IN_QUESTION = 100
```
KennethEnevoldsen, line 8: A bit unsure about this one, but otherwise it looks good.

KennethEnevoldsen (approved):

PR #130, opened 2024-01-09
fix: Extract JSON dicts from generated output better

saattrupdan:
### Fixed
- When few-shot evaluating models on NER tasks, we are now more lenient towards the generated model output. Instead of taking the output as-is, we are now extracting the first dictionary (enclosed in curly brackets), as well as replacing all single apostrophes (') with double ones (").

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,6 +74,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   the model's maximum context length was really small (128). This has been fixed now.
 - Now sets `ignore_mismatched_sizes` when loading models if the model cannot be loaded
   otherwise. This previously caused some issues when loading certain models.
+- When few-shot evaluating models on NER tasks, we are now more lenient towards the
+  generated model output. Instead of taking the output as-is, we are now extracting the
+  first dictionary (enclosed in curly brackets), as well as replacing all single
+  apostrophes (') with double ones (").
 
 
 ## [v8.2.1] - 2023-12-20
--- a/poetry.lock
+++ b/poetry.lock
@@ -620,6 +620,16 @@ torch = ["torch"]
 vision = ["Pillow (>=6.2.1)"]
 
 [[package]]
+name = "demjson3"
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -74,6 +74,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   the model's maximum context length was really small (128). This has been fixed now.
 - Now sets `ignore_mismatched_sizes` when loading models if the model cannot be loaded
   otherwise. This previously caused some issues when loading certain models.
+- When few-shot evaluating models on NER tasks, we are now more lenient towards the
+  generated model output. Instead of taking the output as-is, we are now extracting the
+  first dictionary (enclosed in curly brackets), as well as replacing all single
+  apostrophes (') with double ones (").
```
KennethEnevoldsen, line 80: You might find (some aspects) of this function useful: 

https://github.com/KennethEnevoldsen/bag-of-tricks/blob/master/bagoftricks/llm_json_detection.py

KennethEnevoldsen (approved):
Added a suggestion, but otherwise looks good.

CHANGELOG.md
saattrupdan: Had a read through it, but most of the things doesn't seem applicable here. For instance, code blocks are automatically ignored since we find the dictionary present within it.  However, using `re.DOTALL` makes a lot of sense, so I'll definitely steal that 🙂 

I'm also hesitant to evaluating anything literally, since the LLM output can be malicious.

KennethEnevoldsen: Completely understand it - no we shouldn't evaluate arbitrary code. The demjson3, was by far the change that catches many of the cases I otherwise wouldn't catch so if you still have a few json that seems reasonable which isn't caught you might want to see if it handles those.

saattrupdan: Just tried `demjson3` out, and it works great! It doesn't catch if text exists outside the dictionary however, but a combination of the regex with `demjson3` seems to work really well - I've included it now 🙂

PR #131, opened 2024-01-09
Use vLLM

### LINKED ISSUE #116: Investigate whether vLLM can speed up evaluation
Try experimenting with vLLM for few-shot evaluating generative models, to see if we can speed up the evaluation a bit.

saattrupdan:
### Added
- Now uses vLLM as the underlying engine for few-shot evaluating generative models, which drastically improves the evaluation speed, as well as requiring less GPU memory.

### Changed
- If a model is stored in the Safetensors format on Hugging Face Hub, then we read out the number of parameters directly from those files. This results in more accurate parameter counts as opposed to loading in the model in 4-bit and counting manually.

### Fixed
- If a model is already pre-quantized then we will not attempt to quantize it as well.

This closes #116


KennethEnevoldsen (approved):
Looks good! Solid improvement!

PR #140, opened 2024-01-10
Fix/fp16 bug

### LINKED ISSUE #106: NaN validation loss when benchmarking
The models `vesteinn/ScandiBERT` and `mideind/IceBERT-mC4-is` get NaN validation loss when benchmarking.

We need to check if this is due to the framework or the models themselves. If it's the models then an `InvalidBenchmark` should be thrown instead of getting ~0% scores on everything.

saattrupdan:
### Fixed
- Fixed bug where some encoder models did not work properly when loaded in with FP16 mixed precision due to overflow. We now load in models with BF16 as these have a larger range, but fall back to FP16 if BF16 is not available. If both lead to overflow then we attempt again with full FP32, and lastly throw an informative error and block evaluation if the overflow persists.

With this PR (and #131), I think we're ready for an official release 🙂 

This PR depends on PR #131 

This closes #106


src/scandeval/exceptions.py
```diff
@@ -43,3 +43,17 @@ def __init__(self, message: str = "There is currently no internet connection."):
         """
         self.message = message
         super().__init__(self.message)
+
+
+class NaNValueInModelOutput(Exception):
```
KennethEnevoldsen, line 48: Great idea!

src/scandeval/finetuning.py
```diff
@@ -346,13 +387,19 @@ def get_training_args(
...
+        bf16:
+            Whether to use bf16 mixed precision training for the current iteration.
         batch_size:
             The batch size to use for the current iteration, or None if the batch size
             in the benchmark config should be used.
 
     Returns:
         The training arguments for the current iteration.
     """
+    assert not (fp16 and bf16), "Cannot use both fp16 and bf16 mixed precision training"
```
KennethEnevoldsen: As they are mutually exclusive wouldn't you rather have something like precision = Literal["fp16", "bf16", "fp32"] (or use an enum).

src/scandeval/generation.py
```diff
@@ -236,19 +237,28 @@ def generate_single_iteration(
             ]
         )
 
-        batch_size = (
-            1 if isinstance(model, OpenAIModel) else benchmark_config.batch_size
-        )
+        if isinstance(model, OpenAIModel):
+            batch_size = 1
+        elif isinstance(model, VLLMModel):
+            batch_size = len(torch_dataset)
```
KennethEnevoldsen, line 243: Hmm maybe I should take a closer look at vLMM seems very convenient

src/scandeval/model_setups/hf.py
```diff
@@ -240,94 +241,129 @@ def load_model(
...
                             )
-                    except (KeyError, RuntimeError) as e:
-                        if not model_kwargs["ignore_mismatched_sizes"]:
-                            logger.debug(
-                                f"{type(e).__name__} occurred during the loading "
-                                f"of the {model_id!r} model. Retrying with "
-                                "`ignore_mismatched_sizes` set to True."
+                        except ImportError as e:
+                            if "flash attention" in str(e).lower():
+                                raise InvalidBenchmark(
```
KennethEnevoldsen: InvalidBenchmark, is that the right error here?

src/scandeval/model_setups/hf.py
```diff
@@ -240,94 +241,129 @@ def load_model(
...
+                                model_kwargs["ignore_mismatched_sizes"] = True
+                                continue
+                            else:
+                                raise InvalidBenchmark(str(e))
+                        except (TimeoutError, RequestError):
+                            logger.info(
+                                f"Couldn't load the model {model_id!r}. Retrying."
                             )
-                            model_kwargs["ignore_mismatched_sizes"] = True
+                            sleep(5)
```
KennethEnevoldsen, line 345: So it just keeps trying forever?

src/scandeval/finetuning.py
saattrupdan: Good idea, I'll do that 🙂

KennethEnevoldsen:
> please review https://github.com/ScandEval/ScandEval/pull/131 first, since this one depends on that one?

Ah just reviewed it jointly here. I have added a LGTM over at the other thread as well as there were nothing that stood out to me in that section.

src/scandeval/model_setups/hf.py
saattrupdan: Added a `FlashAttentionNotInstalled` exception now

src/scandeval/model_setups/hf.py
saattrupdan: Good point, and yes, it does. I've added a maximum number of retries now (5).

PR #144, opened 2024-01-13
Feat/optimize cache memory

### LINKED ISSUE #143: Reduce memory consumption of `ModelCache`
The cache during generation currently takes up a lot of (regular) memory, which causes crashes at the end of iterations. We should reduce the memory consumption of the cache somehow.

This could involve:
1. Storing fewer scores;
2. Being more selective about when we should store scores (i.e., only when we're dealing with sequence-classification currently);
3. Not keeping all of the cache in memory, but instead working with a JSONL file directly;
4. Only keeping a registry of the keys (i.e., input prompts) in memory, potentially even just a hash of them, and then fetching them from disk when needed;
5. Having separate caches for different datasets;

saattrupdan:
## Summary
This PR is primarily reducing memory usage during benchmarking, as well as slightly reducing benchmarking speed and disk space use.

### Changed
- Now only stores the top-10 log probabilities of generated tokens when the generation length is less than 8 tokens. Also now keeps separate caches for each (model, dataset) combination, where it previously had a single cache for each model. Both of these help reduce the memory usage of the model output cache.
- Optimised cache saving/loading a bit, making the waiting time in between iterations slightly shorter.
- Removes the model output cache for a (model, dataset) combination when the benchmarking of the model on the dataset finishes successfully. Also removed indents in model output cache JSON files. Both of these help reducing the disk space used on caching.

### Fixed
- Only require generative models to output logprobs if the dataset is of a task that requires it. This caused the benchmarking to use excessive memory when benchmarking datasets that require long generative outputs, such as NER.

### Removed
- Removed some vLLM logging.

This closes #143

### DIFF
--- a/src/scandeval/generation.py
+++ b/src/scandeval/generation.py
@@ -80,6 +80,21 @@ def generate(
     """
     scores: dict[str, list[dict[str, float]]] = defaultdict(list)
 
+    # Initialise the model output cache. If we are testing then we save the model
+    # outputs to a different cache and ensure that that cache is deleted before the
+    # next test, to ensure that the tests are independent of each other
+    model_cache_dir = Path(model_config.model_cache_dir)
+    if not hasattr(sys, "_called_from_test"):
+        cache_name = f"{dataset_config.name}-model-outputs.json"
+    else:
+        cache_name = f"{dataset_config.name}-model-outputs-test.json"
+        (model_cache_dir / cache_name).unlink(missing_ok=True)
+    cache = ModelCache(
+        model_cache_dir=model_cache_dir,
+        cache_name=cache_name,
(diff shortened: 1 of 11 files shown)


src/scandeval/generation.py
```diff
@@ -155,7 +163,7 @@ def generate_single_iteration(
     extract_labels_fn: Callable[..., list[Any]],
     dataset_config: DatasetConfig,
     benchmark_config: BenchmarkConfig,
-    model_cache_dir: Path,
+    cache: ModelCache,
```
KennethEnevoldsen, line 166: <3

KennethEnevoldsen (approved):
Lgtm

PR #146, opened 2024-01-15
fix: Add `generative` metadata

saattrupdan:
This adds a `generative` key to the records stored to disk, which makes parsing the results for the leaderboards easier.

This is currently solely used to add a "(few-shot)" suffix to model IDs in the leaderboard, as we're planning to add finetuning support for decoders in the future.


KennethEnevoldsen (approved):
Looks good

PR #147, opened 2024-01-15
Fix/few shot checking duplicates

saattrupdan:
Previously we checked for the value of `few_shot` when checking if a model has already been benchmarked, no matter if the model is generative or not. Now we only check this if the model is generative.

### DIFF
--- a/src/scandeval/benchmarker.py
+++ b/src/scandeval/benchmarker.py
@@ -548,8 +548,8 @@ def model_has_been_benchmarked(
         if (
             record.model == model_id
             and record.dataset == dataset
-            and record.few_shot == few_shot
             and record.validation_split == validation_split
+            and (not record.generative or record.few_shot == few_shot)
         ):
             return True
     return False
--- a/tests/test_benchmarker.py
+++ b/tests/test_benchmarker.py
@@ -235,7 +235,7 @@ class TestBenchmarkResult:
                 BenchmarkResult(
                     model="model",
                     dataset="dataset",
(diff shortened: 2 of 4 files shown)


src/scandeval/benchmarker.py
```diff
@@ -548,8 +548,8 @@ def model_has_been_benchmarked(
         if (
             record.model == model_id
             and record.dataset == dataset
-            and record.few_shot == few_shot
             and record.validation_split == validation_split
+            and (not record.generative or record.few_shot == few_shot)
```
KennethEnevoldsen: Quite a long conditional. Might be relevant to name element of it to make it more clear what each segment check. (it is not too bad as the function names the overall conditional)

tests/test_benchmarker.py
```diff
@@ -326,6 +343,7 @@ def test_append_to_results(self, benchmark_result, results_path):
         "model has not been benchmarked",
         "model few-shot has not been benchmarked",
         "model few-shot has been benchmarked",
+        "model few-shot has been benchmarked, but not generative",
```
KennethEnevoldsen, line 346: Martin suggested (and I have had a few successes with it, but haven't tried it more than a few times) to create a test object when test-cases becomes quite large. E.g. `ModelHasBeenBenchmarkedCase`.

KennethEnevoldsen (approved):
A few very minor suggestion (feel free to ignore). Otherwise looks good.

PR #157, opened 2024-01-17
Feat/add no is en multiple choice

saattrupdan:
### Added
- Added (the English) datasets MMLU, ARC and HellaSwag, as well as Norwegian and Icelandic translations of it. Now the `knowledge` and `common-sense-reasoning` tasks are covered in all supported languages except Faroese (i.e., da, sv, no, is, de, nl & en).


KennethEnevoldsen (commented):
Looks good. Not much to add.

peterbjorgensen:
Some the ARC questions in the full dataset have 5 choices, i.e. some also have option "e", and in one case it is the correct answer (for the example with id `ARC-Challenge/validation/TIMSS_2003_8_pg29`). It looks like the code just ignore this option? It might also be that for the truncated mini version of the datasets, none of these few examples show up.

PR #160, opened 2024-01-18
mistral can be run without flash attention in upcoming transformers

peter-sk:
Should not be merged before 4.37.0 is out


saattrupdan:
@peter-sk It seems as if you haven't set up the pre-commit hooks, which causes linting errors. Can you run `make install` on your side, to make sure that you've installed the repo correctly?

peter-sk:
Sure. make install does, however, throw some error when running it:
(base) petersk@Mac ScandEval % make install
Installing the 'ScandEval' project...
/bin/sh: brew install gnupg pinentry-mac: command not found

peter-sk:
manually run the exact command 'brew install gnupg pinentry-mac' and then it seems to work. weird!

saattrupdan:
> pre-commit hooks installed - should I do something now?

Make a small change in the file (adjust a comment slightly or something) and try committing. Then the pre-commit hooks should kick in.

peter-sk:
mhh ... I get:

diff --git a/src/scandeval/model_setups/hf.py b/src/scandeval/model_setups/hf.py
index 8760680..be43b76 100644
--- a/src/scandeval/model_setups/hf.py
+++ b/src/scandeval/model_setups/hf.py
@@ -1,4 +1,4 @@
-"""Model setup for Hugging Face Hub models."""
+"""Model setup for Hugging Face Hub models. Awesome stuff!"""

 import logging
 import warnings
(1/1) Stage this hunk [y,n,q,a,d,e,?]? y

black....................................................................Passed
ruff.....................................................................Passed
mypy.....................................................................Passed
error: gpg failed to sign the data
fatal: failed to write commit object

peter-sk:
I registered the GPG key with github. What command is run by the hook? Then I might be able to debug it.

saattrupdan:
> mhh ... I get:
> 
> diff --git a/src/scandeval/model_setups/hf.py b/src/scandeval/model_setups/hf.py index 8760680..be43b76 100644 --- a/src/scandeval/model_setups/hf.py +++ b/src/scandeval/model_setups/hf.py @@ -1,4 +1,4 @@ -"""Model setup for Hugging Face Hub models.""" +"""Model setup for Hugging Face Hub models. Awesome stuff!"""
> 
> import logging import warnings (1/1) Stage this hunk [y,n,q,a,d,e,?]? y
> 
> black....................................................................Passed ruff.....................................................................Passed nbstripout...........................................(no files to check)Skipped mypy.....................................................................Passed error: gpg failed to sign the data fatal: failed to write commit object

That's just GPG which isn't set up properly. I should probably remove GPG support for this repo, since so many are using it. For now, locate the line in your `.env` file with `GPG_KEY_ID` in it, and replace that line with `GPG_KEY_ID=""`. That should hopefully deal with it 🙂 

Alternatively, if you want to use GPG then you should be able to commit when you've added it to your Github account.

peter-sk:
.git/hooks/pre-commit, I guess. I also found .pre-commit-config.yaml. But nothing about gpg signing.

peter-sk:
not working, when I set GPG_KEY_ID to ""

peter-sk:
my key is added to github and it doesn't work with or without GPG_KEY_ID :-(

peter-sk:
git config --local commit.gpgsign false

SAVED THE DAY :-)

saattrupdan:
@peter-sk Since the `use_flash_attention` variable is now simply a synonym for `self.benchmark_config.use_flash_attention`, we should probably remove it and use `self.benchmark_config.use_flash_attention` directly instead, to avoid clutter.

Also, I get that the models can be run without flash attention, but we probably want to inform the user if model *could* be run with flash attention (resulting in less memory usage). Also, in case `flash-attn` is installed, we should probably always use flash attention.

saattrupdan (approved):
LGTM!

PR #161, opened 2024-01-18
chore: Remove GPG

saattrupdan:
This removes GPG signing from the installation of the repo, as is almost always a pain to set up for newcomers, if they haven't set it up on their machine already.


KennethEnevoldsen (approved):
Seems like a reasonable thing to do. Makes it easier to contribute

PR #162, opened 2024-01-18
Add multiple choice datasets and speculative ngram sampling, fix errors

### LINKED ISSUE #159: [FEATURE REQUEST] Use ngram speculation
**Describe the desired feature**
Ngram speculation is a generation technique which increases throughput up to 3x with 100% same outputs. It's implemented in transformers - see https://www.linkedin.com/posts/gante_up-to-3x-faster-llm-generation-with-no-extra-ugcPost-7153089812474798081-ZBZB. 

**Rationale**
Makes few-shot evaluation faster when vLLM isn't available.

saattrupdan:
## Summary
The main bits of this PR making generation faster and more robust. We were unsure about whether the vLLM framework was good enough in terms of generation quality, but after the below fixes I conducted experiments on the NER datasets `suc3` and `dansk`, the summarization dataset `nordjylland-news` as well as the classification dataset `mmlu-da`, and the only place where there was a significant difference (two-tailed t-test, p < 0.05) was the micro-average F1-score when `MISC` tags was included in the `suc3` dataset. This result, along with the fact that the nature of the `MISC` tags varies heavily for each NER dataset resulted in swapping the primary and secondary metric for the NER task, so we now use the micro-F1 without `MISC` tags as the primary metric. With this change, all primary metrics are now not different with vLLM, justifying keeping it. Note that vLLM is ~3x faster for summarization, ~5x faster for NER, and ~0.33x faster for text classification.

### Added
- Added (the English) datasets MMLU, ARC and HellaSwag, as well as Norwegian and Icelandic translations of it. Now the `knowledge` and `common-sense-reasoning` tasks are covered in all supported languages except Faroese (i.e., da, sv, no, is, de, nl & en).
- Now uses speculative ngram sampling for text generation when vLLM is not available. This has no effect on performance and increases evaluation speed by 3x on generation heavy tasks like NER and summarization.

### Changed
- Swapped primary/secondary metrics for the NER task, as the `MISC` tag varies too much from dataset to dataset to be meaningful as a primary metric. Now uses micro-average F1-score across all tags except the `MISC` tag as a primary metric.

### Fixed
- Previously `temperature` was wrongly not set for vLLM and OpenAI models, instead defaulting to their 1.0 values. This was due to the fact that this is set in `transformers` using the `do_sample=False` argument, which doesn't transfer to the other libraries. This has now been set to 0.0.
- Now catches OpenAI `InvalidRequestError`s.
- Removed overly long or repetitive samples in the multiple choice datasets, which caused errors when evaluating OpenAI models on them.
- Now sets the `top_k` parameter in the vLLM `SamplingParams` based on the value it has in the `GenerationConfig`. This caused a discrepancy, as vLLM defaulted to -1 and `transformers` to 50.
- When loading a model using `transformers` then the quantized compute dtype is now correctly set to either `bfloat16` or `float16`, depending on the GPU available, rather than the previous `float32`. This does not affect generation performance.
- Fixed formatting of summarization metrics.

This closes #159


src/scripts/create_arc.py
```diff
@@ -72,6 +80,72 @@ def main() -> None:
...
+            & (test_df.option_c.str.len() >= MIN_NUM_CHARS_IN_OPTION)
+            & (test_df.option_c.str.len() <= MAX_NUM_CHARS_IN_OPTION)
+            & (test_df.option_d.str.len() >= MIN_NUM_CHARS_IN_OPTION)
+            & (test_df.option_d.str.len() <= MAX_NUM_CHARS_IN_OPTION)
+        ]
+
+        def is_repetitive(text: str) -> bool:
+            """Return True if the text is repetitive."""
+            max_repetitions = max(Counter(text.split()).values())
+            return max_repetitions > 50
```
peter-sk: Why 50? Shouldn't this at least be a constant MAX_NUM_REPETITIONS (or an argument, of course)?

src/scripts/create_arc.py
```diff
@@ -72,6 +80,72 @@ def main() -> None:
...
+            & ~train_df.option_d.apply(is_repetitive)
+        ]
+        val_df = val_df[
+            ~val_df.instruction.apply(is_repetitive)
+            & ~val_df.option_a.apply(is_repetitive)
+            & ~val_df.option_b.apply(is_repetitive)
+            & ~val_df.option_c.apply(is_repetitive)
+            & ~val_df.option_d.apply(is_repetitive)
+        ]
+        test_df = test_df[
```
peter-sk: Maybe a small helper function that can be called for train_df, val_df, and test_df?

src/scripts/create_hellaswag.py
```diff
@@ -38,6 +46,30 @@ def main() -> None:
...
+                lambda endings: min(len(ending) for ending in endings)
+                >= MIN_NUM_CHARS_IN_OPTION
+                and max(len(ending) for ending in endings) <= MAX_NUM_CHARS_IN_OPTION
+            )
+        ]
+
+        def is_repetitive(text: str) -> bool:
+            """Return True if the text is repetitive."""
+            max_repetitions = max(Counter(text.split()).values())
+            return max_repetitions > 50
```
peter-sk: Same 50 ...

src/scripts/create_mmlu.py
```diff
@@ -48,6 +56,34 @@ def main() -> None:
         # Rename the columns
         df.rename(columns=dict(answer="label"), inplace=True)
 
+        # Remove the samples with overly short or long texts
+        df = df[
+            (df.instruction.str.len() >= MIN_NUM_CHARS_IN_INSTRUCTION)
```
peter-sk, line 62: Maybe also a function for this? It looks similar/identical to the one for ARC.

src/scripts/create_mmlu.py
```diff
@@ -48,6 +56,34 @@ def main() -> None:
...
+            & (df.option_c.str.len() >= MIN_NUM_CHARS_IN_OPTION)
+            & (df.option_c.str.len() <= MAX_NUM_CHARS_IN_OPTION)
+            & (df.option_d.str.len() >= MIN_NUM_CHARS_IN_OPTION)
+            & (df.option_d.str.len() <= MAX_NUM_CHARS_IN_OPTION)
+        ]
+
+        def is_repetitive(text: str) -> bool:
+            """Return True if the text is repetitive."""
+            max_repetitions = max(Counter(text.split()).values())
+            return max_repetitions > 50
```
peter-sk: 50 ...

peter-sk (changes requested):
Looks great except a lot of copy-paste polymorphism. See the comments in the text.

saattrupdan:
@peter-sk Thanks for your comments! Created a `MAX_REPETITIONS` variable in `scripts/constants.py` now, and created a helper function in the `create_arc` script, to avoid a lot of the code redundancy. I'm hesitant to make utility functions across dataset scripts, as I'd ideally like them to be as stand-alone as possible, for ease of readability and portability. Currently the only thing the scripts are sharing are the constants, which is quite minimal.

peter-sk (approved):
Much better.

PR #169, opened 2024-01-21
Feat/add structured generation to ner

### LINKED ISSUE #128: Change NER output to structured generation
Currently, for generative models, we ask the models to produce JSON dictionaries that contain all of the named entities within the text. The keys in the dictionaries are things like "person" and "location" (translated to the given language), with values being lists of strings, being the entities in the text of that type. An example:

```json
{"person": ["Peter"], "location": [], "organization": [], "miscellaneous": ["Adidas"]}
```

Rather than asking the model to produce this JSON dict as-is, we can instead use _structured generation_, where the JSON structure and keys are fixed, and the model only needs to output the contents of the lists. This means that we're not penalising the models for not knowing the proper JSON format, which we can test separately, if needed. Aside from that, the models also have to generate fewer tokens, speeding up generation.

saattrupdan:
## Summary
Many generative models, especially the smaller ones, performed really poorly at the NER datasets due to their inability to generate JSON dictionaries. Since the NER task shouldn't be measuring this, this PR uses _structured generation_, which tweaks the token probabilities during generation to ensure that the resulting output is a JSON dictionary. It's not 100% perfect, since it runs out of tokens sometimes (upper limit for NER is 128), but it's way better.

For the structured generation I'm using a package called [lm-format-enforcer](https://github.com/noamgat/lm-format-enforcer), which is quite slim, is actively maintained, and integrates well with both `transformers` and `vllm`. For `transformers` it produces a function that can be supplied as the `prefix_allowed_tokens_fn` argument in the `generate` method, and for `vllm` it produces a function which is used in the `logits_processors` argument in the `SamplingParams` object. This package allows the models to produce tabs and newlines in the JSON dictionary, but I experienced a lot of the smaller models would just spam this, making them run out of tokens, so I manually tweaked the two functions to disallow these tokens to ever be generated.

### Added
- Added structured generation for the NER task, which enables the models to (almost) always output correct JSON, separating the NER capabilities from the JSON capabilities. JSON can be tested separately in a (future) coding benchmark.

### Fixed
- Removed print output from `bert_score` during summarization metric computation.

Closes #128.


src/scandeval/generation.py
```diff
@@ -321,6 +333,86 @@ def generate_single_iteration(
...
+    if dataset_config.task == NER and isinstance(tokenizer, PreTrainedTokenizer):
+        parser = get_ner_parser(dataset_config=dataset_config)
+        json_prefix_allowed_tokens_fn = build_transformers_prefix_allowed_tokens_fn(
+            tokenizer_data=tokenizer, character_level_parser=parser
+        )
+
+        # Do not allow newlines or tabs in the generated text
+        forbidden_token_ids = list(
+            set(list(tokenizer("\n\n\n\t\t\t", add_special_tokens=False).input_ids))
+        )
```
KennethEnevoldsen: it is because double newline can have its own token? Might this not break if e.g. a tokenizer has a custom token for triple newline? Might be worth spelling making clear which tokens are being forbidden.


E.g.
```
       forbidden_tokens = ["\n", "\n\n", "/t", "\t\t"]
        forbidden_token_ids = ...
```

KennethEnevoldsen (approved):
Glad to see the package. Funny to see the use of pydantic models for schemas in LLMs (an approach I also used a while back in a research project). Def. seems like a reasonable package to rely on here.

Do we only give a schema to open source LLMs or also to APIs? (we should probably be consistent as I imagine it might also improve APIs)

> This package allows the models to produce tabs and newlines in the JSON dictionary, but I experienced a lot of the smaller models would just spam this, making them run out of tokens, so I manually tweaked the two functions to disallow these tokens to ever be generated.

Hmm right. That might be worth a PR or issue to the package.

src/scandeval/generation.py
saattrupdan: That's exactly the reason, yeah. And you're right, it _would_ break if it has a special triple newline token, so I'll fix that with your suggestion 🙂

saattrupdan:
> Do we only give a schema to open source LLMs or also to APIs? (we should probably be consistent as I imagine it might also improve APIs)

We can't really do it for the OpenAI API at least, since they don't give access to token probabilities (anymore).

saattrupdan:
> > This package allows the models to produce tabs and newlines in the JSON dictionary, but I experienced a lot of the smaller models would just spam this, making them run out of tokens, so I manually tweaked the two functions to disallow these tokens to ever be generated.
> 
> Hmm right. That might be worth a PR or issue to the package.

It doesn't seem like the maintainers want it to be a part of the package, someone already tried:
https://github.com/noamgat/lm-format-enforcer/pull/60

So I think we just have to deal with it here, separately.

src/scandeval/generation.py
saattrupdan: Changed it now.

KennethEnevoldsen:
> > Do we only give a schema to open source LLMs or also to APIs? (we should probably be consistent as I imagine it might also improve APIs)
> 
> We can't really do it for the OpenAI API at least, since they don't give access to token probabilities (anymore).

Oh well, I think it is fair that they don't get a bonus.

src/scandeval/generation.py
```diff
@@ -356,9 +356,14 @@ def get_prefix_allowed_fn(
         )
 
         # Do not allow newlines or tabs in the generated text
-        forbidden_token_ids = list(
-            set(list(tokenizer("\n\n\n\t\t\t", add_special_tokens=False).input_ids))
-        )
+        forbidden_token_ids = list()
+        for raw_token in ["\n", "\t"]:
```
KennethEnevoldsen:
```suggestion
        forbidden_tokens =  ["\n", "\t"]
        for raw_token in forbidden_tokens:
```
Works as a comment. I would then remove the comment above (I would also recommend this [video](https://www.youtube.com/watch?v=Bf7vDBBOBUA))

src/scandeval/generation.py
```diff
@@ -356,9 +356,14 @@ def get_prefix_allowed_fn(
...
+            for num_tokens in range(1, 4):
```
KennethEnevoldsen: I would probably just write them out

KennethEnevoldsen (approved):
A few sugesstion otherwise looks good

src/scandeval/generation.py
saattrupdan: Thanks for the video link! 

I agree that the comment here is not needed, and I've implemented your suggestion, but I don't agree with the overall sentiment of the video. In the end it's very subjective, and he makes it very clear when he mentions that *he* doesn't usually read comments to understand what the code is doing. Others (myself included) _first_ read the comments in a code base to get an idea of what is going on, and _then_ read the code as the implementation of the thing. For me it's simply less mental effort to read English than reading Python - even if the variables are properly named 🙂

src/scandeval/generation.py
saattrupdan: Done!

src/scandeval/generation.py
KennethEnevoldsen: Hmm actually think I do docstring first, then code, then comments - but it depends, the few times I did some PRs for the explosion library I loved the few comments that helped me understand - so get where you are coming from (the video is a controversial piece)

saattrupdan: Agree with the docstring bit!

src/scandeval/generation.py
```diff
@@ -321,6 +333,89 @@ def generate_single_iteration(
...
+    prefix_allowed_tokens_fns = list()
+
+    if dataset_config.task == NER and isinstance(tokenizer, PreTrainedTokenizer):
+        parser = get_ner_parser(dataset_config=dataset_config)
+        json_prefix_allowed_tokens_fn = build_transformers_prefix_allowed_tokens_fn(
+            tokenizer_data=tokenizer, character_level_parser=parser
+        )
+
+        forbidden_token_ids = list()
+        forbidden_tokens = ["\n", "\n\n", "\n\n\n", "\t", "\t\t", "\t\t\t"]
```
peter-sk, line 359: I am worried that for some tokenizers, this might not cover all the tokens that express new lines etc. See the discussion here on the llama tokenizers and their context dependence. I encountered this when using the completiononly collator.

https://huggingface.co/docs/trl/sft_trainer#using-tokenids-directly-for-responsetemplate

PR #195, opened 2024-01-25
chore: Swap multiple choice metrics

saattrupdan:
### Changed
- Swapped primary/secondary metrics for the multiple choice tasks, where we now set MCC as the primary metric and accuracy and secondary. This is due to the fact that MCC handles class imbalance better.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,13 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+## [Unreleased]
+### Changed
+- Swapped primary/secondary metrics for the multiple choice tasks, where we now set MCC
+  as the primary metric and accuracy and secondary. This is due to the fact that MCC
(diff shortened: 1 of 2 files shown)


KennethEnevoldsen (approved):
Sounds reasonable. I assume it is similar in other multiple choice benchmarks?

saattrupdan:
> Sounds reasonable. I assume it is similar in other multiple choice benchmarks?

Actually not. Accuracy is the main metric for multiple choice, but that again assumes that the classes are equally distributed, so it's not ideal. But I keep accuracy as secondary metric, precisely to still allow a comparison.

KennethEnevoldsen:
Sounds fine to merge - do you have a reference on the comparison (then I will just take a look at it)

saattrupdan:
> Sounds fine to merge - do you have a reference on the comparison (then I will just take a look at it)

Here's a few:

1. https://doi.org/10.1186/s13040-021-00244-z
2. https://doi.org/10.1186/s12864-019-6413-7
3. https://doi.org/10.1109/ACCESS.2021.3084050
4. https://doi.org/10.1371/journal.pone.0177678

PR #199, opened 2024-01-27
Many smaller changes and bug fixes

saattrupdan:
## Summary
A bunch of smaller changes:

1. New minor version of `transformers` is out, 4.37, which as always introduces breaking changes 🙄  For instance, speculative n-gram sampling now only supports batch size 1, which now doesn't add any speed increase, so we remove it. They also changed their Flash Attention API, so instead of setting `use_flash_attention_2=True` in the `generate` call, we have to set `attn_implementation="flash_attention_2"`.
2. There was a bug when changing from one vLLM model to another, where the memory wasn't released properly, causing OOM errors. This was due to the fact that in the `VLLMModel.__del__` method there was a reference to `self._model`, but this is removed after calling `destroy_model_parallel`, so the `clear_memory` line afterwards was never called.
3. For vLLM models, we now include a `set_tokenizer` method instead of setting the property, since this allows us to change the underlying tokenizer in the vLLM model.
4. We use a dummy input value in generative models two places, the `model_is_generative` utility method and when we're checking what the maximum sequence length of the model is. The token ID that we use as the dummy input is fixed, and this is now recorded in `DUMMY_INPUT_VALUE`, and is set to 100, as it cannot be a special token (like `<s>`) and these are usually either in the beginning or end of the vocabulary.
5. Style changes:
    1. Updated pre-commit hooks, which includes a lot of `black` changes throughout the code base.
    2. Fixed a bunch of docstrings that hadn't been changed in "the great docstring change".

### Changed
- Now requires `transformers` versions `4.37.x`. As they often introduce breaking changes in minor versions, we now only allow a patch version difference and manually update to `4.38.x` when it comes out.
- Removed speculative ngram sampling again, as `transformers` now requires the batch size to be 1, which doesn't make it any faster than normal.

### Fixed
- Fixed an issue with OOM errors when changing from benchmarking one generative model to another.
- When testing a model's maximum sequence length, we put dummy inputs into them. This causes errors if the dummy inputs are one of the special tokens. Since the special tokens have not always been set up in the tokenizer, we instead rely on a heuristic that the 100th token ID is not a special token.
- An import depended on `vllm`, which is not installed on non-Linux devices, causing an `ImportError`. This has now been removed.
- Fixed an issue where structured generation wasn't triggered when vLLM wasn't available.


KennethEnevoldsen (approved):
Looks good!

not sure you are aware, but ruff implements black in their formatter (almost exactly the same, so removes a dependency)

saattrupdan:
> Looks good!
> 
> 
> 
> not sure you are aware, but ruff implements black in their formatter (almost exactly the same, so removes a dependency) 

Ooh good to know always happy to remove a dependency 😊

PR #203, opened 2024-02-01
Update deps, use ruff format, small fixes

saattrupdan:
This PR updates the following dependencies:

1. `pandas` is now ^2.2.0 from ^1.4.0
2. `bitsandbytes` is now >=0.42.0,<1.0.0 from >=0.40.0.post4,<1.0.0
3. `vllm` is now >=0.3.0,<1.0.0 from >=0.2.7,<1.0.0
    - This now allows `pydantic>=2` and also finally doesn't download _both_ PyTorch and Safetensors model files if they're both available)
5. `pydantic` is now ^0.2.6 from >=1.10.0,<3.0.0
6. The dev dependency `lxml` is now >=4.9.3 from ^4.9.3

Aside from dependencies, this PR also adds the following:

1. We also get rid of `black` and instead use `ruff-format`, which results in a lot of formatted files.
2. Fixes the bug that causes OOM when transitioning from evaluating one vLLM model to another vLLM model.
3. Fixes the bug that model revision didn't previously work with vLLM models, as the `revision` argument to `LLM` wasn't used.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: jpetrucciani/black-check@master
+      - uses: jpetrucciani/ruff-check@main
 
   pytest:
     strategy:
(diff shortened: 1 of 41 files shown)


KennethEnevoldsen (commented):
Looks good!

PR #281, opened 2024-03-11
Fix/include outlines transformers integration

saattrupdan:
### Fixed
- Now includes the `transformers` integration with `outlines` directly in the code,
  which caused issues as they weren't part of the newest `outlines` release. When it
  does get included then we will import these as before.
- When evaluating OpenAI models we now do not perform any structured generation, as we
  do not have access to the logits.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Now includes the `transformers` integration with `outlines` directly in the code,
   which caused issues as they weren't part of the newest `outlines` release. When it
   does get included then we will import these as before.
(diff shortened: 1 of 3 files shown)


larsbun:
My installation from pip (12.2.0) breaks because of references to outlines.integrations. An indication of a package update would be great. Ty.

saattrupdan:
> My installation from pip (12.2.0) breaks because of references to outlines.integrations. An indication of a package update would be great. Ty.

@larsbun  This should be fixed in version 12.2.1. Can you check if that fixes your issue?

larsbun:
> > My installation from pip (12.2.0) breaks because of references to outlines.integrations. An indication of a package update would be great. Ty.
> 
> @larsbun This should be fixed in version 12.2.1. Can you check if that fixes your issue?

Works like a charm. TY.

PR #301, opened 2024-03-21
Bandaid fixes to support azure

BramVanroy:
This PR adds a number of changes to support Azure's OpenAI service. Some noteworthy changes:

- add `azure_openai_api_version` argument, which is a required for API calls
- add corresponding environment variable
- naively assumes that Azure models only use cl100k_base tokenizer (they only support a handful of models like 3.5 and 4, not the older ones). The reason for this is that `model_config.model_id` must be the `deployment name` that was chosen when creating the endpoint. This can be arbitrary and will likely not be a valid model identifier. I do not know if/how you can get the currently deployed model from the client but perhaps someone else knows a way. That would make life easier.

Some stylistic issues may need to be changed and perhaps I missed some places where to add `azure_openai_api_version` (there's a lot of repetition of arguments so I hope I got them all).

With these changes I am currently running the benchmarks. Example command:

```shell
scandeval -l nl -m <deployment-name> --azure-openai-api-key <apik-key> --azure-openai-endpoint <https-endpoint> --azure-openai-api-version <api-version> --prefer-azure
```

### DIFF
--- a/src/scandeval/benchmark_config_factory.py
+++ b/src/scandeval/benchmark_config_factory.py
@@ -41,6 +41,7 @@ def build_benchmark_config(
     prefer_azure: bool,
     azure_openai_api_key: str | None,
     azure_openai_endpoint: str | None,
+    azure_openai_api_version: str | None,
     force: bool,
     verbose: bool,
     trust_remote_code: bool,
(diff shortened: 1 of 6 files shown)


saattrupdan (approved):
LGTM, thanks!

PR #363, opened 2024-03-29
Move Tensor to same device as dummy inputs

ThomasKluiters:
Hey guys, I was benchmarking some models - and one of these models happened to be a T5-based model. This would fail as the Torch.zeros() was not on the same device as the input_ids.


saattrupdan (approved):
Nice catch, thanks for the fix! 😊

PR #396, opened 2024-04-16
Feat/add evaluation app

saattrupdan:
### Added
- Added a script to evaluate human performance on datasets. This is a Gradio app which
  can be run using the command `human_evaluate --annotator-id <id>`, where
  `annotator-id` is the ID of the human annotator (from 0 to 10, inclusive). They will
  then annotate their answers for validation splits from the iteration corresponding to
  their annotator ID. All of the annotated examples of dataset `<dataset>` are saved to
  the CSV file `.scandeval_cache/human-evaluation/<dataset>/human-<id>.csv`.


KennethEnevoldsen (approved):
The code looks clean, haven't tested out the UI.

PR #458, opened 2024-06-08
Bump outlines library

s-smits:
Quite new to updating poetry libraries, does it automatically update poetry.lock or do we have to **poetry add** it through CLI?


saattrupdan:
I appreciate what you're trying to do, but please open an issue before opening a PR.

The issue for the vLLM-Outlines dependency clash can be found [here](https://github.com/ScandEval/ScandEval/issues/414) and will be fixed when when vLLM has a new release.

If you have trouble installing ScandEval or upgrading outlines separately, then please open a separate issue. Thanks!

PR #564, opened 2024-10-28
Dataset Icelandic qa

oliverkinch:
Added `icelandic-qa`, an Icelandic question answering dataset about Icelandic culture and history ([source](https://huggingface.co/datasets/mideind/icelandic_qa_scandeval)).

The dataset is very small. There are a total of 2000 samples, but after asserting that the answer must be in the context, there are only 375 samples left.

Fix https://github.com/ScandEval/ScandEval/issues/515

Wops, made this branch from https://github.com/ScandEval/ScandEval/pull/561 so that is why these changes are also included here.

### DIFF
--- a/src/scripts/create_icelandic_qa.py
+++ b/src/scripts/create_icelandic_qa.py
@@ -120,12 +120,18 @@ def main() -> None:
     ]
     test_df = df_with_answer_filtered.sample(n=test_size, random_state=4242)
 
-    # Create train split
-    train_size = 531
+    # Create train split, which is the rest of the data
+    train_size = len(df_with_answer) - val_size - test_size
     full_train_df_with_answer = df_with_answer_filtered.loc[
         ~df_with_answer_filtered.index.isin(test_df.index)
     ]
     train_df = full_train_df_with_answer.sample(n=train_size, random_state=4242)
+
+    train_num_samples_lower_bound = 500
+    assert len(train_df) > train_num_samples_lower_bound, (
+        f"There are only {len(train_df):,} samples in the training set - there should be at "
+        f"least {train_num_samples_lower_bound:,}."
+    )
 
     val_df = val_df.reset_index(drop=True)
     test_df = test_df.reset_index(drop=True)
--- a/src/scripts/create_icelandic_qa.py
+++ b/src/scripts/create_icelandic_qa.py
@@ -24,7 +24,7 @@ load_dotenv()
 
 def main() -> None:
     """Create the Icelandic QA dataset about Icelandic history and culture and upload it to the HF Hub."""
-    client = OpenAI(api_key=os.environ.get("OPENAI_API_KEY"))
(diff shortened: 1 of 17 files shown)


saattrupdan:
Yeah I think for this one it's important that we keep as many samples as possible, since they've all been manually reviewed. The answers of course have to appear in the context, so we would have to change the answers a bit. We could potentially get an LLM to rephrase the answer to make it correspond to a phrase that appears in the context, making the dataset creation script a bit more involved.

oliverkinch:
> Yeah I think for this one it's important that we keep as many samples as possible, since they've all been manually reviewed. The answers of course have to appear in the context, so we would have to change the answers a bit. We could potentially get an LLM to rephrase the answer to make it correspond to a phrase that appears in the context, making the dataset creation script a bit more involved.

I have tried to implement this. We now keep 1683 samples instead of only 375.

src/scripts/create_icelandic_qa.py
```diff
@@ -0,0 +1,156 @@
...
+        ~df_with_answer.index.isin(val_df.index)
+    ]
+    test_df = df_with_answer_filtered.sample(n=test_size, random_state=4242)
+
+    # Create train split
+    train_size = 531
+    full_train_df_with_answer = df_with_answer_filtered.loc[
+        ~df_with_answer_filtered.index.isin(test_df.index)
+    ]
+    train_df = full_train_df_with_answer.sample(n=train_size, random_state=4242)
```
saattrupdan: It would be a bit more robust if we simply take the remaining samples here, rather than specifically 531:

```suggestion
    # Create train split
    train_df = df_with_answer_filtered.loc[
        ~df_with_answer_filtered.index.isin(test_df.index)
    ]
    
    # Sanity check that we include enough training samples
    train_num_samples_lower_bound = 500
    assert len(train_df) > train_num_samples_lower_bound, (
        f"There are only {len(train_df):,} samples in the training set - there should be at "
        f"least {train_num_samples_lower_bound:,}."
    )
```

src/scripts/create_icelandic_qa.py
```diff
@@ -0,0 +1,156 @@
...
+from huggingface_hub.hf_api import HfApi
+from openai import OpenAI
+from requests.exceptions import HTTPError
+
+load_dotenv()
+
+
+def main() -> None:
+    """Create the Icelandic QA dataset about Icelandic history and culture and upload it to the HF Hub."""
+    client = OpenAI(api_key=os.environ.get("OPENAI_API_KEY"))
```
saattrupdan: If we don't have an API key then we should fail fast, as it's required:

```suggestion
    client = OpenAI(api_key=os.environ["OPENAI_API_KEY"])
```

saattrupdan (changes requested):
A couple of points, otherwise looks great. Fantastic that we can keep so many samples now!

saattrupdan (approved):
Looks good! Just need the tests to pass now.

PR #608, opened 2024-11-27
 Add question answering part of the NorGLM NO-Multi-QA-Sum dataset

viggo-gascou:
Added the [NorGLM NO-Multi-QA-Sum dataset](https://huggingface.co/datasets/NorGLM/NO-Multi-QA-Sum)

Fixes more of https://github.com/ScandEval/ScandEval/issues/536

A lot of the answers in the original dataset did not contain the answer from the context.  These have been rephrased using gpt-4o to do so.

The dataset consists of 2406 samples with the following splits:

- 1024 training samples
- 256 validation samples
- 1126 test samples


src/scandeval/dataset_configs.py
```diff
@@ -771,6 +771,21 @@ def get_dataset_config(dataset_name: str) -> DatasetConfig:
...
+    "NO-Multi-QA-Sum",
+    huggingface_id="ScandEval/norglm-multi-qa",
+    task=RC,
+    languages=[NB, NN, NO],
+    prompt_prefix="Her følger tekster med tilhørende spørsmål og svar.",
+    prompt_template="Tekst: {text}\nSpørsmål: {question}\nSvar på maks 3 ord: {label}",
+    instruction_prompt="Tekst: {text}\n\nBesvar følgende spørsmål om teksten ovenfor "
+    "med maks 3 ord.\n\nSpørsmål: {question}",
+    num_few_shot_examples=2,
+    max_generated_tokens=32,
```
saattrupdan, line 788:
```suggestion
    max_generated_tokens=32,
    unofficial=True,
```

src/scripts/create_norglm_multiqa.py
```diff
@@ -0,0 +1,203 @@
...
+    val_size = 256
+    val_df = df.sample(n=val_size, random_state=4242)
+
+    # Create test split
+    test_size = 1024
+    filtered_df = df[~df.index.isin(val_df.index)]
+    test_df = filtered_df.sample(n=test_size, random_state=4242)
+
+    # Create train split
+    train_df = filtered_df[~filtered_df.index.isin(test_df.index)]
```
saattrupdan: How many samples are there? Maybe you can add a `assert len(cleaned_df) == XXX` before the splits, just to make it clear to the reader what's going on?

saattrupdan (changes requested):

src/scripts/create_norglm_multiqa.py
viggo-gascou: Yeah makes sense! I also just noticed that I used the wrong df for splitting.. gonna run the rephrasing again and get the total 😄

viggo-gascou: Should be fixed now! Let me know if there are other fixes I may have missed 😄

src/scripts/create_norglm_multiqa.py
```diff
@@ -0,0 +1,209 @@
...
```
saattrupdan: I'd suggest we change the split to 1024/256/X train/val/test, where X is "the rest". Just to make my reasoning a bit more systematic, these are the rules of thumb I'm following when coming up with the split sizes:

1. There should be at most 1,024 train samples, 256 val samples and 2,048 test samples.
2. There should always be more test samples than train samples. The test split is the most crucial split.
3. The number of validation samples should be way less than the other splits.

The reason for the rules is to make the datasets as consistent as possible with each other.

```suggestion
    # Create validation split
    val_size = 256
    val_df = df.sample(n=val_size, random_state=4242)

    # Create train split
    train_size = 1024
    filtered_df = df[~df.index.isin(val_df.index)]
    train_df = filtered_df.sample(n=train_size, random_state=4242)

    # Create test split, using the remaining samples
    test_df = filtered_df[~filtered_df.index.isin(train_df.index)]
```

saattrupdan (changes requested):

src/scripts/create_norglm_multiqa.py
viggo-gascou: Alright, sounds good! Makes sense will keep that in mind, thanks 😄

saattrupdan (approved):
Looks good now! 😊

PR #703, opened 2025-01-12
feat: Update squad-nl

### LINKED ISSUE #670: [BENCHMARK DATASET REQUEST] NetherlandsForensicInstitute/squad-nl-v2.0
### Dataset name

NetherlandsForensicInstitute/squad-nl-v2.0

### Dataset link

https://huggingface.co/datasets/NetherlandsForensicInstitute/squad-nl-v2.0

### Dataset languages

- [ ] Danish
- [ ] Swedish
- [ ] Norwegian (Bokmål or Nynorsk)
- [ ] Icelandic
- [ ] Faroese
- [ ] German
- [X] Dutch
- [ ] English

### Describe the dataset

This is also a machine translated version of SQuAD to Dutch, but the translation is better than the current one we use, so this should simply replace the old dataset. Note that we also need to re-benchmark all the models on this new one.

See a comparison [here](https://huggingface.co/datasets/yhavinga/squad_v2_dutch/discussions/2#6763ed4c42436c7f7005f4b4).

saattrupdan:
### Changed
- Updated the Dutch reading comprehension dataset SQuAD-nl, being a machine translated
  version of the English SQuAD dataset. Previously we used the `yhavinga/squad_v2_dutch`
  version, but this has been changed to `NetherlandsForensicInstitute/squad-nl-v2.0`,
  following [this evaluation showing that the latter is of higher
  quality](https://huggingface.co/datasets/yhavinga/squad_v2_dutch/discussions/2#6763ed4c42436c7f7005f4b4).

Fixes #670

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 
 ## [Unreleased]
+### Changed
+- Updated the Dutch reading comprehension dataset SQuAD-nl, being a machine translated
+  version of the English SQuAD dataset. Previously we used the `yhavinga/squad_v2_dutch`
+  version, but this has been changed to `NetherlandsForensicInstitute/squad-nl-v2.0`,
+  following [this evaluation showing that the latter is of higher
+  quality](https://huggingface.co/datasets/yhavinga/squad_v2_dutch/discussions/2#6763ed4c42436c7f7005f4b4).
+
 ### Fixed
 - The `trust_remote_code` argument was not supplied when loading the Hugging Face
   configuration in some places, which caused an unnecessary dialogue with the user when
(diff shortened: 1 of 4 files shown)


Rijgersberg:
A bit late but I think you should use the canonical version of SQuAD-NL v2.0 as uploaded by @wietsedv, as discussed here: https://github.com/ScandEval/ScandEval/issues/670

That dataset is https://huggingface.co/datasets/GroNLP/squad-nl-v2.0

saattrupdan:
@Rijgersberg Ah my bad. Is the data the same? Because in that case I can just change the script and leave the dataset as-is.

Rijgersberg:
No problem :)

Yes, the data is the same, with one exception: `NetherlandsForensicInstitute/squad-nl-v2.0` adds a column `score` that is derived from the `answers` column. This column score is superfluous for ScandEval. So it should be just a matter of swapping out the dataset identifier.

saattrupdan:
> No problem :)
> 
> Yes, the data is the same, with one exception: `NetherlandsForensicInstitute/squad-nl-v2.0` adds a column `score` that is derived from the `answers` column. This column score is superfluous for ScandEval. So it should be just a matter of swapping out the dataset identifier.

Perfect!

PR #723, opened 2025-01-20
remove framework as a CLI argument

peter-sk:
this would remove framework from the command line parameters. we should DEFINITELY wait to see if this still might be needed to be re-activated once the local models PR has been finalized.


peter-sk:
I updated it such that it could be merged right after the local-models-reloaded branch.

saattrupdan (approved):
LGTM!

PR #722, opened 2025-01-20
reloaded attempt at enabling local models

peter-sk:
this works for me for local models such as a downloaded TinyLlama/TinyLlama_v1.1 or a self-trained OLMo2 model.

PR #724, opened 2025-01-20
Load local HF model w/ vLLM

### LINKED ISSUE #713: [FEATURE REQUEST] Support local models

Enable evaluations of local models. This should make it possible to just point at a local path, rather than a Hugging Face Hub model ID.

Ideally this should use the existing evaluation modules, and just tweak the model existence and model loading functions therein.

### Alternatives

_No response_

### Additional context

_No response_

rlrs:
Closes #713. 

Some issues:
- Several things are hardcoded that likely shouldn't be, such as the `task` and the `framework`?
- Safetensors are apparently not detected in the local directory.

I also added another possible padding token, `"<eos>"`, since that's what I needed. However, this might be possible to read from the config instead?

### DIFF
--- a/src/scandeval/benchmark_modules/vllm.py
+++ b/src/scandeval/benchmark_modules/vllm.py
@@ -254,7 +254,7 @@ class VLLMModel(HuggingFaceEncoderModel):
             self._tokenizer.pad_token_id = self._tokenizer.eos_token_id
             self._tokenizer.pad_token = self._tokenizer.eos_token
         elif self._tokenizer.pad_token_id is None:
-            pad_token_candidates = ["<pad>", "[pad]", "<|endoftext|>", "<|im_end|>"]
+            pad_token_candidates = ["<pad>", "[pad]", "<|endoftext|>", "<|im_end|>", "<eos>"]
             pad_token_candidates.extend([c.upper() for c in pad_token_candidates])
             for candidate in pad_token_candidates:
                 if candidate in self._tokenizer.get_vocab():
@@ -380,6 +380,11 @@ class VLLMModel(HuggingFaceEncoderModel):
             Whether the model exists, or an error describing why we cannot check
             whether the model exists.
         """
+        if os.path.isdir(model_id):  # Check if it's a local path
+            # Verify essential files exist
+            required_files = ["config.json"]  # Add/modify as needed
+            return all(os.path.exists(os.path.join(model_id, f)) for f in required_files)
+
         using_api = (
             benchmark_config.api_base is not None
             or benchmark_config.api_version is not None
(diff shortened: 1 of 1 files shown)


src/scandeval/benchmark_modules/vllm.py
```diff
@@ -380,6 +380,11 @@ def model_exists(
             Whether the model exists, or an error describing why we cannot check
             whether the model exists.
         """
+        if os.path.isdir(model_id):  # Check if it's a local path
+            # Verify essential files exist
+            required_files = ["config.json"]  # Add/modify as needed
+            return all(os.path.exists(os.path.join(model_id, f)) for f in required_files)
```
saattrupdan, line 386: This only checks if the model exists locally, but doesn't check if it is a generative model. We could grab this information from the config like this (import `load_hf_model_config` from `.hf`):

```suggestion
        if os.path.isdir(model_id):  # Check if it's a local path
            # Verify essential files exist
            required_files = ["config.json"]  # Add/modify as needed
            required_files_exist = all(os.path.exists(os.path.join(model_id, f)) for f in required_files)
            if not required_files_exist:
                return False
                
            # Check if model is generatove
            hf_config = load_hf_model_config(
                model_id=model_id,
                num_labels=0,
                id2label=dict(),
                label2id=dict(),
                revision=revision,
                model_cache_dir=create_model_cache_dir(
                  cache_dir=benchmark_config.cache_dir, model_id=model_id
              ),
              api_key=benchmark_config.api_key,
              trust_remote_code=benchmark_config.trust_remote_code,
              run_with_cli=benchmark_config.run_with_cli,
          )
          class_names = hf_config.architectures
          generative_class_names = [
              class_name
              for model_task in GENERATIVE_MODEL_TASKS
              for class_name in TASK_MAPPING[model_task].values()
          ]
          model_is_generative = class_names is not None and any(
              class_name in generative_class_names for class_name in class_names
          )
          return model_is_generative
```

Also, this should be copied to `hf.py` but with `return not model_is_generative` at the end.

src/scandeval/benchmark_modules/vllm.py
```diff
@@ -412,6 +417,22 @@ def get_model_config(
         Returns:
             The model configuration.
         """
+        if os.path.isdir(model_id):
+            framework = Framework.PYTORCH  # Assuming PyTorch model
+            
+            return ModelConfig(
+                model_id=model_id,
+                revision="local",  # or some other appropriate indicator
```
saattrupdan, line 425: Maybe we could create a new m`ModelType` called `LOCAL_GENERATIVE`?

saattrupdan (changes requested):
Looks good overall! The main thing missing is that this doesn't work with encoder models - I've added a suggestion to fix this.

peter-sk:
This version has two problems:
1) It does not (yet) work for encoder-only models. Making it work would some require code duplication.
2) It already duplicates code that is in get_model_repo_info. This is introduces unnecessary technical debt.
I removed the --pipeline-tag CLI argument reusing the code in get_model_repo_info smartly (thanks to the inspiration from your discussion) and made a clean and minimal implementation.
I am closing this PR and reopened PR #722.

PR #750, opened 2025-01-23
Create Icelandic knowledge dataset using LLM

oliverkinch:
This is the same dataset as the IcelandicQA dataset, but now phrased as a knowledge dataset. An LLM has ben used to generate candidate answers. The split is given as 845 / 128 / 1024 for train, val, and test respectively.

Fix https://github.com/ScandEval/ScandEval/issues/593


oliverkinch:
@saattrupdan do you have some feedback to my prompt before I start the script to build the dataset (1997 samples)?

```python
messages = [
    {
        "role": "user",
        "content": f"For the question: {row.question} where the correct answer is: {row.answer}, please provide 3 plausible alternatives in Icelandic. Format your response using XML tags: <option1>first</option1> <option2>second</option2> <option3>third</option3>",
    }
]
```

saattrupdan:
@oliverkinch I think that could work, but you probably would want to use structured generation instead of XML. I can show you tomorrow 🙂

docs/datasets/icelandic.md
```diff
@@ -629,6 +629,60 @@ You can evaluate this dataset directly as follows:
...
+  Svaraðu með að hámarki 3 orðum: {label}
+  ```
+- Instruction-tuned prompt template:
+  ```
+  Texti: {text}
+
+  Svaraðu eftirfarandi spurningu um textann að hámarki í 3 orðum.
+
+  Spurning: {question}
+  ```
```
saattrupdan: These prompts don't match the multiple choice format, but are instead the reading comprehension prompts.

docs/datasets/icelandic.md
```diff
@@ -629,6 +629,60 @@ You can evaluate this dataset directly as follows:
 $ scandeval --model <model-id> --dataset mmlu-is
 ```
 
+### Unofficial: IcelandicKnowledge
+This dataset is based on the IcelandicQA dataset, which was published [here](https://huggingface.co/,datasets/mideind/icelandic_qa_scandeval), but is here phrased as a knowledge dataset. The candidate answers has been generated by GPT-4o.
+
+The original dataset has 2,000 samples, but only 1,997 unique questions, and the total length of this dataset is therefore 1,997. The split is given by 845 / 128 / 1024 for train, val, and test, respectively.
+
```
saattrupdan, line 650: I would include here a description of the OpenAI pipeline of adding negative options to each question

CHANGELOG.md
```diff
@@ -7,11 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 
 ## [Unreleased]
-
-
+### Added
+- Added the dataset icelandic-knowledge, which is derived from the IcelandicQA dataset, reformatted as a knowledge dataset with GPT-4 generated candidate answers. The split is given by 845 / 128 / 1024 for train, val, and test, respectively. It is marked as `unofficial` for now.
```
saattrupdan: Should be 4o instead of 4

saattrupdan (changes requested):
Just a few docs related things 🙂

PR #771, opened 2025-02-04
docs: Add model comparison section

KennethEnevoldsen:
These changes result in: 

Screenshot 2025-02-04 at 10 11 51


mkdocs.yaml
```diff
@@ -41,11 +42,23 @@ plugins:
   - mkapi
 nav:
   - About: README.md
-  - Leaderboards: leaderboards
+  - Leaderboards:
```
KennethEnevoldsen, line 45: this specification is required to get "compare models" at the end (otherwise it appears at the top)

mkdocs.yaml
```diff
@@ -14,6 +14,7 @@ theme:
     - navigation.instant.prefetch
     - navigation.instant.progress
     - navigation.tracking
+    - navigation.sections
```
KennethEnevoldsen, line 17: required for creating the bold section (to distinguish it from the benchmarks)

mkdocs.yaml
```diff
@@ -41,11 +42,23 @@ plugins:
...
+      - leaderboards/dutch.md
+      - leaderboards/english.md
+      - leaderboards/faroese.md
+      - leaderboards/french.md
+      - leaderboards/german.md
+      - leaderboards/icelandic.md
+      - leaderboards/norwegian.md
+      - leaderboards/swedish.md
+      - Compare Models:
+          - leaderboards/radial_plotter.md
```
saattrupdan: Would this work, if we create an `extras` folder? Naming all the leaderboards manually is sure going to cause bugs in the future. 

```suggestion
      - leaderboards
      - Compare Models:
          - extras/radial_plotter.md
```

KennethEnevoldsen: It works (already discovered missing mixed):
Screenshot 2025-02-04 at 10 34 29

we might consider moving monolingual models into a folder and removing flags
Screenshot 2025-02-04 at 10 38 15

KennethEnevoldsen: Would also change Mixed -> Multilingual

docs/leaderboards/Multilingual/european.md
```diff
@@ -2,7 +2,7 @@
 hide:
     - toc
 ---
-# 🇪🇺 European
+# European
```
saattrupdan: No flag? ☹️

docs/leaderboards/Multilingual/germanic.md
```diff
@@ -2,7 +2,7 @@
 hide:
     - toc
 ---
-# 🇩🇰🇳🇱🇬🇧🇫🇴🇩🇪🇮🇸🇳🇴🇸🇪 Germanic
+# Germanic
```
saattrupdan: No flags? ☹️

docs/leaderboards/Multilingual/mainland-scandinavian.md
```diff
@@ -2,7 +2,7 @@
 hide:
     - toc
 ---
-# 🇩🇰🇳🇴🇸🇪 Mainland Scandinavian
+# Mainland Scandinavian
```
saattrupdan: No flags? ☹️

saattrupdan (commented):
Some flag comments

saattrupdan:
> Some flag comments

Could maybe clarify a bit more here. The reason why I think the flags are important (aside from the visual appeal) is to make it more clear what languages are included. For instance, "Germanic" doesn't cover all Germanic languages (e.g., South African, Frisian), and "Mainland Scandinavian" can cause confusion if not specified.

KennethEnevoldsen:
>Could maybe clarify a bit more here. The reason why I think the flags are important (aside from the visual appeal) is to make it more clear what languages are included. For instance, "Germanic" doesn't cover all Germanic languages (e.g., South African, Frisian), and "Mainland Scandinavian" can cause confusion if not specified.

It was simply because it resulted in odd formatting:

Screenshot 2025-02-04 at 14 44 08

I am fine with keeping them. 

In addition, we should probably add language codes to each leaderboard (I can add those).

saattrupdan:
> It was simply because it resulted in odd formatting
> I am fine with keeping them.

Great - can you add them back in?

> In addition, we should probably add language codes to each leaderboard (I can add those).

To keep things as simple as possible, I would rather leave them out of the leaderboards. Interested people can always check this in the datasets docs. We should keep in mind that the majority of people using these leaderboards are practitioners and not researchers, so it should both be good to non-technical people as well as allowing the technical people to dive into details.

KennethEnevoldsen:
Current view with fixes:

Screenshot 2025-02-04 at 19 05 02

saattrupdan (approved):
LGTM!

PR #776, opened 2025-02-08
Safetensors CLI check

Mikeriess:
# Add safetensors availability check for CLI users
The motivation for this PR is that I personally wont be able to contribute benchmarks for models that are not released in Safetensors format, and when benchmarking multiple models it becomes cumbersome to manually check this every time. Having the CLI implementation check if available and ask for permission to proceed is therefore very useful. 

## Changes
This PR suggests a check for Safetensors availability when running a benchmark from CLI. It warns the user if safetensors version is not available and prompts them to proceed before loading the model. 

## Testing
The implementation has been successfully tested on the following model: [mikeriess/gemma-2b-it-nosafetensor-test](https://huggingface.co/mikeriess/gemma-2b-it-nosafetensor-test/tree/main) which is a clone of gemma-2-2b-it reuploaded without safetensors. 

## Related Issues
This is related to https://github.com/ScandEval/ScandEval/issues/686 and allows more flexibility than originally suggested.

### DIFF
--- a/src/scandeval/benchmarker.py
+++ b/src/scandeval/benchmarker.py
@@ -27,6 +27,8 @@ from .scores import log_scores
 from .speed_benchmark import benchmark_speed
 from .tasks import SPEED
 from .utils import enforce_reproducibility
+import requests
+from huggingface_hub import list_repo_files
 
 if t.TYPE_CHECKING:
     from .benchmark_modules import BenchmarkModule
@@ -209,6 +211,38 @@ class Benchmarker:
         else:
             return list()
 
+    def _check_safetensors_available(self, model_id: str) -> bool:
+        """Check if the model has safetensors weights available.
+        
+        Args:
+            model_id: The Hugging Face model ID to check
+            
+        Returns:
+            bool: True if safetensors available, False otherwise
+        """
+        try:
+            # List files in the repo
+            files = list_repo_files(model_id)
+            # Check if any file ends with .safetensors
+            return any(f.endswith('.safetensors') for f in files)
+        except Exception:
+            # If we can't check (e.g. API error), assume it's fine
+            return True
+
+    def _prompt_safetensors_warning(self, model_id: str) -> bool:
+        """Prompt user about missing safetensors if running from CLI.
+        
+        Args:
+            model_id: The model ID being checked
+            
+        Returns:
+            bool: True to continue, False to abort
+        """
+        print(f"\nWarning: Model {model_id} does not have safetensors weights available.")
+        print("This may pose security risks when loading the model.")
+        response = input("Do you want to continue anyway? [y/N]: ").lower()
+        return response in ['y', 'yes']
+
     def benchmark(
         self,
         model: list[str] | str,
@@ -350,6 +384,16 @@ class Benchmarker:
 
         if benchmark_config.clear_model_cache:
             clear_model_cache_fn(cache_dir=benchmark_config.cache_dir)
+
+        # Only do safetensors check if running from CLI
+        if hasattr(sys, 'ps1') is False:  # Check if running from CLI
+            for model_id in model:
+                # Skip check for non-HF models
+                if not (model_id.startswith('http') or '/' not in model_id):
+                    if not self._check_safetensors_available(model_id):
+                        if not self._prompt_safetensors_warning(model_id):
+                            print("Aborting benchmark.")
+                            sys.exit(1)
 
         model_ids = self._prepare_model_ids(model_id=model)
         dataset_configs = prepare_dataset_configs(


saattrupdan (changes requested):
Thanks for the contribution!

I get what you are trying to do, but the current way seems a bit too hacky and error prone:

1. We should never catch _all_ exceptions in a try-except block.
2. This PR change should concern models from the Hugging Face Hub, so it shouldn't be dealt with in the global `benchmarker` module, as we're then also conducting the check for all API models and freshly initialised models (and future benchmark module types).
3. All print statements should use the `scandeval` logger instead, to allow people to disable logging.
4. To check whether we're using the CLI or not, the benchmark config has a `run_with_cli` attribute for exactly this purpose.

Further, this should not be default behaviour, as most people don't mind loading PyTorch models, for instance.

I'd suggest we instead add an optional CLI flag named something like `--only-allow-safetensors`. The actual check of whether the model exists in the safetensors format can then happen in the `HuggingFaceEncoderModel`  and `VLLMModel` classes, since these are the ones loading models from the Hugging Face Hub. Also, there needs to be a separate check in case the models are loaded locally (this is detected in the `benchmark_modules.hf.get_model_repo_info` function).

So it's a bit more work to ensure that it works for everyone, but I hope you're up for the task!

Mikeriess:
Thanks for the detailed feedback - your suggestion sounds fine as it still solves the core problem. Will have a look at it asap.

Mikeriess:
If this one os too hacky I will blame it on Claude :-) I did the unit test locally this time, but feel free to reject this one if this is too far from what you had in mind. Otherwise, the feedback is very helpful from a learning perspective 👍

src/scandeval/benchmark_modules/hf.py
```diff
@@ -810,6 +810,19 @@ def get_model_repo_info(
...
+        repo_files = hf_api.list_repo_files(repo_id=model_id, revision=revision)
+        has_safetensors = any(f.endswith('.safetensors') for f in repo_files)
+        if not has_safetensors:
+            logger.error(
+                f"Model {model_id} does not have safetensors weights available. "
+                "Skipping due to --only-allow-safetensors flag."
+            )
+            raise InvalidModel(
+                f"Model {model_id} does not have safetensors weights available"
+            )
```
saattrupdan: No need to log separately, as this is done automatically when `InvalidModel` is raised. Instead, we can just add the extra information into the exception body, and also condition it on whether we're using the CLI or not:

```suggestion
            msg = f"Model {model_id} does not have safetensors weights available. "
            if benchmark_config.run_with_cli:
                msg += "Skipping since the `--only-allow-safetensors` flag is set."
            else:
                msg += "Skipping since the `only_allow_safetensors` argument is set to `True`."
            raise InvalidModel(msg)
```

tests/test_benchmark_modules/test_hf.py
```diff
@@ -33,3 +36,55 @@ def test_get_torch_dtype(test_device, torch_dtype_is_set, bf16_available, expect
...
+def test_safetensors_check():
+    """Test the safetensors availability check functionality."""
+    
+    # Mock benchmark config with safetensors check enabled
+    benchmark_config = MagicMock(spec=BenchmarkConfig)
+    benchmark_config.only_allow_safetensors = True
+    benchmark_config.cache_dir = ".cache"
+    benchmark_config.api_key = None
+    benchmark_config.trust_remote_code = False
+    benchmark_config.run_with_cli = True
```
saattrupdan: This test seems to give false confidence, since currently the `BenchmarkConfig` does not have an `only_allow_safetensors` attribute. 

There is already a `benchmark_config` fixture, so you could just use that as follows:

```suggestion
def test_safetensors_check(benchmark_config):
    """Test the safetensors availability check functionality."""
    benchmark_config.only_allow_safetensors = True
```

src/scandeval/data_models.py
```diff
@@ -206,6 +206,7 @@ class BenchmarkConfigParams(pydantic.BaseModel):
     api_version: str | None
     debug: bool
     run_with_cli: bool
+    only_allow_safetensors: bool = False
```
saattrupdan: Having a default value here is error-prone - we do that elsewhere to avoid silent errors (which is currently what is happening now).

```suggestion
    only_allow_safetensors: bool
```

saattrupdan (changes requested):
Almost there! You just need to set the additional `only_allow_safetensors` argument in the following places:

1. `data_models.BenchmarkConfig` (+docstring)
2. `benchmarker.Benchmarker.__init__` (+docstring)
3. `benchmarker.Benchmarker.benchmark_config_default_params`
4. `benchmarker.Benchmarker.benchmark` (+docstring, and in the call to `_get_updated_benchmark_config`)
5. `tests/test_data_models` tests
6. `tests/test_cli` tests
7. The `benchmark_config` fixture in `tests/conftest`

Mikeriess:
@saattrupdan thanks for taking the time to do the detailed feedback. Appreciate it! 🙌

Mikeriess:
Working out the last comments now -

Mikeriess:
@saattrupdan could you describe in more detail what is needed for these points?

> 5. tests/test_data_models tests
> 6. tests/test_cli tests
> 7. The benchmark_config fixture in tests/conftest

saattrupdan:
> @saattrupdan could you describe in more detail what is needed for these points?

Sure thing!

> > 5. tests/test_data_models tests

This is not relevant after all - I thought we had some unit tests relating to the `BenchmarkConfig` attributes.

> > 6. tests/test_cli tests

The `test_cli_param_names` and `test_cli_param_types` tests should include the new flag.

> > 7. The benchmark_config fixture in tests/conftest

The `benchmark_config` fixture should include the new parameter in `BenchmarkConfig`.

Mikeriess:
@saattrupdan ok thanks, I _think_ ive been over all the points now, so it should be ready for another review 👍

saattrupdan (approved):
Looks great! Thanks for your contribution 😊

Mikeriess:
@saattrupdan Thanks for the feedback! 🙌

PR #785, opened 2025-02-15
Update number of original samples for the scala datasets

oliverkinch:
The number of original samples has been wrong for the scala dataset. I have updated them in this PR.

See https://github.com/ScandEval/ScandEval/pull/751#discussion_r1948771232


saattrupdan (approved):
LGTM!

PR #793, opened 2025-02-27
Feat/remove speed

### LINKED ISSUE #792: Remove Speed benchmark
The speed benchmark is not that trustworthy currently, as it heavily depends on the hardware on which it has been run. Further, when checking the speed of API models, this fluctuates a lot, and one can also make that faster by running requests asynchronously.

Thus, there probably isn't much value in having the speed benchmark at all, and we should probably simply remove it. People could then still use the number of parameters as a proxy for the inference speed, for the openly available models.

saattrupdan:
### Removed
- Removed the speed benchmark, as it provided unreliable results and was not used in
  practice.

Fixes #792.


KennethEnevoldsen (commented):
Isn't it still relevant to have speed if you want to compare two models locally? (just wouldn't have it in the leaderboard or in the default tasks)

Totally up for removing it. The user can use something else to estimate speed (that is not something we claim to do well anyway).

saattrupdan:
Closing this, as I agree there might be value in locally testing the speed of different models.

PR #795, opened 2025-02-27
Feat/change name

### LINKED ISSUE #781: Backwards compatible name change
We're planning on changing the name of the benchmark to reflect that we're focusing on all the European languages, and not solely the Scandinavian ones. This change should be 100% backwards compatible, to make people access and work with the benchmark exactly as they're already used to. This means that we have to do the following:

- [x] Tweak the publish script to publish two packages, both `scandeval` and whatever the new name would be. We would thus have two identical packages with different names, so that one could both `pip install scandeval[all]` as usual, as well as installing the new package name. Simply creating an alias package wouldn't quite work, as scripts using, e.g., `from scandeval.benchmark_configs import X` would cease to work in that case. Note that the publish script should both change the _publishing name_ as well as the _package name_. The package name should be changed dynamically by renaming the `src/scandeval` directory to the new name as well as changing `pyproject.toml`, building the package with the new name, and then republishing (and change back again afterwards).
- [x] Buy a new domain
- [x] Move site to new domain
- [x] Set up redirect from scandeval.com to the new domain
- [x] Change the repository names on GitHub - GitHub automatically sets up redirects, so that the current https://github.com/ScandEval/ScandEval would simply redirect to the new repository URL.
- [x] Change the repository names on Hugging Face Hub - this automatically sets up redirects, so that the old dataset IDs will redirect to the new ones.
- [x] Add an additional terminal script in `pyproject.toml`, so that people can both call `scandeval --model <model-id>` as well as replacing `scandeval` with the new command.

saattrupdan:
This does the following:

- Updates the package name to `euroeval`
- Updates all documentation to 'EuroEval'
- Adds `euroeval` CLI command
- Updates CNAME to 'euroeval.com'
- Updates `make publish` to both publish the `euroeval` package, as well as the `scandeval` package, by changing the project name in `pyproject.toml` and renaming `src/euroeval` to `src/scandeval` temporarily.
- Updates all references to the Hugging Face organisation to `EuroEval` organisation (the dataset IDs with the old `ScandEval` organisation name will automatically redirect to the new ones).
- Updates the `Dockerfile` to use `euroeval`

Fixes #781


KennethEnevoldsen:
Image on the current leaderboard breaks

KennethEnevoldsen (approved):
Looks good. I think the image needs a rename also - otherwise I didn't see any biggies.

What do we do with the org + repo name (new org and transfer repo + rename?)

KennethEnevoldsen:
+ github links are broken as they refer to the org (website links to https://github.com/EuroEval/EuroEval)

saattrupdan:
@KennethEnevoldsen Yeah the Git repo change is "future proof", since I'll be changing the name of both the repo and org anyway, so that link will work when I've done that.

As for the logo, you're completely right, and it depends on the logo we end up going for. Out of the suggestions in [the thread](https://github.com/ScandEval/ScandEval/issues/781#issuecomment-2688225758), number 4 seems to be the winner, but if you want to suggest others, then please post more in there 🙁

PR #827, opened 2025-03-13
Simplify dataset configs with templates and defaults for each task

viggo-gascou:
This PR aims to simplify the addition of new datasets and languages by having a single spot where prompt prefixes and templates are defined. This way we ensure that the prompts are always the same regardless of the dataset.

#### Todo:
- [x] Set up prompt data class config for tasks
- [x] Set up template data class for prompts
- [x] Add templates for all tasks and languages
    - [x] `COMMON_SENSE`
    - [x] `KNOW`
    - [x] `LA`
    - [x] `MCRC`
    - [x] `NER`
    - [x] `RC`
    - [x] `SENT`
    - [x] `SPEED`
    - [x] `SUMM`
- [ ] Clean up `dataset_configs.py`
- [x] Automatically create labels as string for datasets that have labels e.g., `labels=["a","b","c"]` as a string: `"'a', 'b', or 'c',"` - In the correct language.
    - [x] Handle NER special-case with label_mapping being the labels and separator being 'and' 
- [x] Set up defaults for all tasks - so they can be overridden if necessary


viggo-gascou:
Do you want to review things so far? Before I start deleting things in the dataset configs 😅 @saattrupdan

src/euroeval/data_models.py
```diff
@@ -360,6 +365,75 @@ def __hash__(self) -> int:
...
+
+        If the task is NER, we separate using 'and' and use the mapped labels instead of
+        the BIO NER labels.
+
+        Args:
+            language: The language to be used when converting the labels.
+
+        Returns:
+            str: The natural string representation of the labels in specified language.
+            If the language is not found, it defaults to English.
```
saattrupdan: We normally avoid including types in the docstrings if we're using type hints, to avoid unnecessary duplication.

```suggestion
            The natural string representation of the labels in specified language.
            If the language is not found, it defaults to English.
```

src/euroeval/data_models.py
```diff
@@ -360,6 +365,75 @@ def __hash__(self) -> int:
...
+
+        Example:
+            >>> print(get_labels_str(["a", "b", "c"]), "da")
+            "'a', 'b', 'c' eller 'd'"
+        """
+        if self.task.name == "named-entity-recognition":
```
saattrupdan: Using strings here is error prone (I've made that mistake with NER before in this repo, even!).

It seems like this should be the case for all token classification tasks (which currently only includes NER). We could thus change it to:

```suggestion
        if self.task.task_group == TaskGroup.TOKEN_CLASSIFICATION:
```

Would of course need to import `TaskGroup` from the `enums` module.

src/euroeval/data_models.py
```diff
@@ -360,6 +365,75 @@ def __hash__(self) -> int:
...
+            sep_word = LANG_TO_AND.get(language, "and")
+            # Create a list of the unique mapped labels with single-quotes.
+            quoted_labels = [
+                f"'{label}'" for label in set(self.prompt_label_mapping.values())
+            ]
+        else:
+            sep_word = LANG_TO_OR.get(language, "or")
+            # Convert labels to single-quoted labels.
+            quoted_labels = [f"'{label}'" for label in self.labels]
```
saattrupdan: Note that `self.labels` is always in English. I assume you want the localised labels here, which you can get by applying the `self.prompt_label_mapping` to each label.

src/euroeval/data_models.py
```diff
@@ -360,6 +365,75 @@ def __hash__(self) -> int:
...
+            return ""
+        elif len(quoted_labels) == 1:
+            return quoted_labels[0]
+        elif len(quoted_labels) == 2:
+            return f"{quoted_labels[0]} {sep_word} {quoted_labels[1]}"
+        else:
+            return f"{', '.join(quoted_labels[:-1])} {sep_word} {quoted_labels[-1]}"
+
+    def __post_init__(self) -> None:
+        """Post Initialisation setup of defaults."""
```
saattrupdan:
```suggestion
        """Post initialisation setup of defaults."""
```

src/euroeval/data_models.py
```diff
@@ -360,6 +365,75 @@ def __hash__(self) -> int:
...
+        # Skip setting defaults if the task is "speed" as it doesn't have any
```
saattrupdan: Any what? 🤔

src/euroeval/data_models.py
```diff
@@ -360,6 +365,75 @@ def __hash__(self) -> int:
...
+        if self.task.name != "speed":
```
saattrupdan: Again, wouldn't use strings here. You could again use task groups for this:

```suggestion
        if self.task.task_group != TaskGroup.SPEED:
```

src/euroeval/data_models.py
```diff
@@ -360,6 +365,75 @@ def __hash__(self) -> int:
...
+            # TODO: should we just use the first language in the list?
+            main_language = self.languages[0]
+            template = TEMPLATES_DICT[self.task.name][main_language.code]
```
saattrupdan: Multiple languages are a bit tricky, but currently this only includes Bokmål/Nynorsk IIRC. Is that being dealt with in the `TEMPLATES_DICT`?

src/euroeval/data_models.py
```diff
@@ -360,6 +365,75 @@ def __hash__(self) -> int:
...
+            if not self.num_few_shot_examples:
+                self.num_few_shot_examples = template.num_few_shot_examples
+            if not self.prompt_template:
+                self.prompt_template = template.prompt_template
+            if not self.prompt_prefix:
+                self.prompt_prefix = template.prompt_prefix
+            if not self.instruction_prompt:
+                self.instruction_prompt = template.instruction_prompt
+
+            labels_str = self._get_labels_str(main_language.code)
```
saattrupdan:
```suggestion
            labels_str = self._get_labels_str(language=main_language.code)
```

src/euroeval/data_models.py
```diff
@@ -360,6 +365,75 @@ def __hash__(self) -> int:
...
         return hash(self.name)
 
+    def _get_labels_str(self, language: str) -> str:
+        """Converts a set of labels to a natural string, in the specified language.
+
+        If the task is NER, we separate using 'and' and use the mapped labels instead of
+        the BIO NER labels.
+
+        Args:
+            language: The language to be used when converting the labels.
```
saattrupdan: Maybe use the `Language` class instead of a string for the language code?

src/euroeval/data_models.py
```diff
@@ -360,6 +365,75 @@ def __hash__(self) -> int:
...
+            if not self.prompt_template:
+                self.prompt_template = template.prompt_template
+            if not self.prompt_prefix:
+                self.prompt_prefix = template.prompt_prefix
+            if not self.instruction_prompt:
+                self.instruction_prompt = template.instruction_prompt
+
+            labels_str = self._get_labels_str(main_language.code)
+
+            def replace_labels(text: str) -> str:
```
saattrupdan, line 480: Missing docstring

src/euroeval/data_models.py
```diff
@@ -360,6 +365,75 @@ def __hash__(self) -> int:
...
+                return text.replace("{labels_str}", labels_str)
+
+            self.prompt_template = replace_labels(self.prompt_template)
+            self.prompt_prefix = replace_labels(self.prompt_template)
+            self.instruction_prompt = replace_labels(self.prompt_template)
```
saattrupdan:
```suggestion
            self.prompt_template = replace_labels(text=self.prompt_template)
            self.prompt_prefix = replace_labels(text=self.prompt_template)
            self.instruction_prompt = replace_labels(text=self.prompt_template)
```

src/euroeval/templates/__init__.py
```diff
@@ -0,0 +1,25 @@
...
+TEMPLATES_DICT = {
+    "common-sense-reasoning": COMMON_SENSE_TEMPLATES,
+    "knowledge": KNOW_TEMPLATES,
+    "linguistic-acceptability": LA_TEMPLATES,
+    "multiple-choice-reading-comprehension": MCRC_TEMPLATES,
+    "named-entity-recognition": NER_TEMPLATES,
+    "reading-comprehension": RC_TEMPLATES,
+    "sentiment-classification": SENT_TEMPLATES,
+    "summarization": SUMM_TEMPLATES,
+}
```
saattrupdan: Again, I'm wary of using strings here. Can we either change them to `Task` classes, or alternatively include a test that checks if all tasks are in there, and that all the keys are indeed tasks?

src/euroeval/templates/base.py
```diff
@@ -0,0 +1,86 @@
+"""Base classes for task-specific prompting across languages."""
```
saattrupdan: Can we rename the `templates` directory to `prompt_templates`?

src/euroeval/templates/base.py
```diff
@@ -0,0 +1,86 @@
...
+    num_few_shot_examples: int
+    max_generated_tokens: int
+    labels: list[str] = field(default_factory=list)
+    prompt_label_mapping: dict[str, str] | str = field(default_factory=dict)
+
+    def __post_init__(self) -> None:
+        """Post Initialisation setup of defaults."""
+        if isinstance(self.prompt_label_mapping, str):
+            if self.prompt_label_mapping == "auto":
+                self.prompt_label_mapping = {label: label for label in self.labels}
```
saattrupdan: Can we add a debug log after this line? You can use the `log_once` function from the `utils` module. Logging something like "Using the prompt label mapping {self.prompt_label_mapping} since `prompt_label_mapping` was set to "auto".

src/euroeval/templates/base.py
```diff
@@ -0,0 +1,86 @@
...
+
+    Defines the prompt templates needed for evaluating a specific task in a
+    given language.
+
+    Attributes:
+        labels:
+            The labels in the dataset. Defaults to an empty list.
+        prompt_label_mapping:
+            A mapping from the labels to another phrase which is used as a substitute
+            for the label in few-shot evaluation. Defaults to an empty dictionary.
```
saattrupdan: There are no defaults.

```suggestion
        labels:
            The labels in the dataset.
        prompt_label_mapping:
            A mapping from the labels to another phrase which is used as a substitute
            for the label in few-shot evaluation.
```

src/euroeval/templates/base.py
```diff
@@ -0,0 +1,86 @@
...
+            The template for the prompt to use when benchmarking the dataset using
+            few-shot evaluation.
+        instruction_prompt:
+            The prompt to use when benchmarking the dataset using instruction-based
+            evaluation.
+        prompt_label_mapping (optional):
+            A mapping from the labels to another phrase which is used as a substitute
+            for the label in few-shot evaluation. Defaults to an empty dictionary. If
+            set to "auto", the mapping will be set to a 1:1 mapping between the labels
+            and themselves.
```
saattrupdan: No default

```suggestion
        prompt_label_mapping:
            A mapping from the labels to another phrase which is used as a substitute
            for the label in few-shot evaluation. If set to "auto", the mapping will be 
            set to a 1:1 mapping between the labels and themselves.
```

src/euroeval/templates/constants.py
```diff
@@ -0,0 +1,177 @@
+"""Constants used for the template configurations."""
+
+NER_TAG_MAPPING = {
```
saattrupdan: I would probably move these label mappings into their respective template modules, to avoid spreading all the localisations out too much.

src/euroeval/templates/constants.py
```diff
@@ -0,0 +1,177 @@
...
+    "per": {
+        "da": "person",
```
saattrupdan: Could we use the `Language` objects as keys here?

src/euroeval/templates/constants.py
```diff
@@ -0,0 +1,177 @@
...
+    "fo": "og",
+    "fr": "et",
+    "is": "og",
+    "it": "e",
+    "nb": "og",
+    "nl": "en",
+    "nn": "og",
+    "no": "og",
+    "sv": "och",
+}
```
saattrupdan: Consider moving these directly into the `Language` objects, as they're general and not task-specific. We could make it into a property which defaults to raising `NotImplementedError`, and just add it to these languages.

src/euroeval/templates/linguistic_acceptability.py
```diff
@@ -0,0 +1,155 @@
...
+    Args:
+        language_code: The language code for the target language.
+
+    Returns:
+        The translated labels for the specified language.
+    """
+    try:
+        return {
+            f"{label}": LA_LABEL_MAPPING[label][language_code]
+            for label in LA_LABEL_MAPPING.keys()
```
saattrupdan:
```suggestion
            label: language_to_local_label[language_code]
            for label, language_to_local_label in LA_LABEL_MAPPING.items()
```

src/euroeval/templates/constants.py
viggo-gascou: Just to be sure, by "these", you mean the OR and AND mappings right? 😄

src/euroeval/templates/base.py
viggo-gascou: Yeah that makes sense, done!

src/euroeval/data_models.py
viggo-gascou: All of the Norwegian prompt templates have an entry, they are identical for now.

src/euroeval/data_models.py
viggo-gascou: any defaults 😅

viggo-gascou:
Oh, and whether we still want to keep all of the configs in one file or split them by task?

src/euroeval/data_models.py
```diff
@@ -67,11 +67,35 @@ class Language:
...
     name: str
+    and_separator: str | None = field(default=None)
+    or_separator: str | None = field(default=None)
 
     def __hash__(self) -> int:
         """Return a hash of the language."""
         return hash(self.code)
 
+    def get_and_separator(self) -> str:
+        """Get the word 'and' in the language.
```
saattrupdan, line 83: Docstring missing "Returns" block.

src/euroeval/data_models.py
```diff
@@ -67,11 +67,35 @@ class Language:
...
+        if `and_separator` is `None` it raises NotImplementedError.
+        """
+        if not self.and_separator:
+            raise NotImplementedError(
+                f"Separator for the word 'and' has not been defined for {self.name}."
+            )
+        return self.and_separator
+
+    def get_or_separator(self) -> str:
+        """Get the word 'or' in the language.
```
saattrupdan, line 104: Docstring missing "Returns" block.

src/euroeval/data_models.py
```diff
@@ -67,11 +67,35 @@ class Language:
...
+    or_separator: str | None = field(default=None)
 
     def __hash__(self) -> int:
         """Return a hash of the language."""
         return hash(self.code)
 
+    def get_and_separator(self) -> str:
+        """Get the word 'and' in the language.
+
+        if `and_separator` is `None` it raises NotImplementedError.
```
saattrupdan: Raising errors uses the reserved keyword "Raises:" - see [this section of the Google style](https://google.github.io/styleguide/pyguide.html#doc-function-raises):

```suggestion
        Raises:
            NotImplementedError:
                If `and_separator` is `None`.
```

src/euroeval/data_models.py
```diff
@@ -67,11 +67,35 @@ class Language:
...
+        if not self.and_separator:
+            raise NotImplementedError(
+                f"Separator for the word 'and' has not been defined for {self.name}."
+            )
+        return self.and_separator
+
+    def get_or_separator(self) -> str:
+        """Get the word 'or' in the language.
+
+        if `or_separator` is `None` it raises NotImplementedError.
```
saattrupdan: Move to "Raises" block, as described above.

src/euroeval/data_models.py
```diff
@@ -375,23 +399,22 @@ def _get_labels_str(self, language: str) -> str:
             language: The language to be used when converting the labels.
 
         Returns:
-            str: The natural string representation of the labels in specified language.
-            If the language is not found, it defaults to English.
+            The natural string representation of the labels in specified language.
+            If the language is not set, it raises a NotImplementedError, see `Language`.
```
saattrupdan: Move to "Raises" block.

src/euroeval/data_models.py
```diff
@@ -375,23 +399,22 @@ def _get_labels_str(self, language: str) -> str:
...
 
         Example:
-            >>> print(get_labels_str(["a", "b", "c"]), "da")
+            >>> get_labels_str(["a", "b", "c"], DA)
```
saattrupdan:
```suggestion
            >>> get_labels_str(language=DA)
```

src/euroeval/languages.py
```diff
@@ -52,25 +52,25 @@ def get_all_languages() -> dict[str, Language]:
 CR = Language(code="cr", name="Cree")
 HR = Language(code="hr", name="Croatian")
 CS = Language(code="cs", name="Czech")
-DA = Language(code="da", name="Danish")
+DA = Language(code="da", name="Danish", and_separator="og", or_separator="eller")
```
saattrupdan: Can we move the languages with specified operators to the top?

src/euroeval/prompt_templates/__init__.py
```diff
@@ -0,0 +1,39 @@
...
+from .linguistic_acceptability import get_linguistic_acceptability_templates
+from .multiple_choice import get_multiple_choice_templates
+from .named_entity import get_ner_templates
+from .reading_comprehension import get_reading_comprehension_templates
+from .sentiment import get_sentiment_templates
+from .summarization import get_summarization_templates
+
+if t.TYPE_CHECKING:
+    from ..data_models import Language, Task
+    from ..tasks import COMMON_SENSE, KNOW, LA, MCRC, NER, RC, SENT, SUMM
```
saattrupdan: Is this line needed?

```suggestion
```

src/euroeval/prompt_templates/common_sense.py
```diff
@@ -0,0 +1,116 @@
+"""Templates for the Common Sense Reasoning task."""
+
+from dataclasses import asdict
+
+from ..types import TemplateDict
+from .base import BasePromptConfig, PromptConfig
+
+
+def get_common_sense_templates() -> TemplateDict:
+    """Get the templates for common sense reasoning task."""
```
saattrupdan: Missing Returns block

saattrupdan: (same thing for several other functions)

src/euroeval/languages.py
```diff
@@ -17,6 +17,26 @@ def get_all_languages() -> dict[str, Language]:
...
+NO = Language(code="no", name="Norwegian", _and_separator="og", _or_separator="eller")
+NB = Language(
+    code="nb", name="Norwegian Bokmål", _and_separator="og", _or_separator="eller"
+)
+NN = Language(
+    code="nn", name="Norwegian Nynorsk", _and_separator="og", _or_separator="eller"
+)
+ES = Language(code="es", name="Spanish", _and_separator="y", _or_separator="o")
+SV = Language(code="sv", name="Swedish", _and_separator="och", _or_separator="eller")
+### Currently Supported Languages ###
```
saattrupdan:
```suggestion
```

src/euroeval/prompt_templates/common_sense.py
```diff
@@ -7,7 +7,11 @@
 
 
 def get_common_sense_templates() -> TemplateDict:
-    """Get the templates for common sense reasoning task."""
+    """Get the templates for common sense reasoning task.
+
+    Returns:
+        A `TemplateDict` containing the templates for common sense reasoning task.
```
saattrupdan: Missing fix!

saattrupdan (changes requested):

saattrupdan (approved):
LGTM!

PR #865, opened 2025-03-28
Fix|Feat/ Load model in float16 for bfloat16 models on GPUs with CUDA capability < 8.0 

### LINKED ISSUE #864: [FEATURE REQUEST] Feature / bug? VLLM support for CUDA GPUs with compute capability below 8

Hi, not sure whether to post this as a bug or feature request, but as of now,
generative evaluation requires vLLM to run/serve the model. 
The issue I am facing is that vLLM will not load checkpoints with bfloat16 for these GPUs, but 
this can be circumvented by telling vLLM to use float16 instead of bfloat16. For vLLM CLI, that's `--dtype=half`
but I am pretty sure this can be set in the constructor of the equivalent python class.

I saw that this issue was brought up earlier but marked as resolved because EuroEval would fall back on using 🤗 transformers backend, but that no longer seems to be the case.

https://github.com/EuroEval/EuroEval/issues/422
 

I have added a pull request for the suggested change here:

#865

marksverdhei:
Resolves #864 

This PR adds a util function for checking device capability and a
check in the vLLM wrapper code for loading models in `torch.bfloat16` with CUDA capability < 8.0 GPUs.

### DIFF
--- a/src/euroeval/benchmark_modules/vllm.py
+++ b/src/euroeval/benchmark_modules/vllm.py
@@ -30,6 +30,7 @@ from ..constants import (
     REASONING_MAX_TOKENS,
     TASK_GROUPS_USING_LOGPROBS,
     TASKS_USING_JSON,
+    VLLM_BF16_MIN_CUDA_COMPUTE_CAPABILITY,
 )
 from ..data_models import (
     BenchmarkConfig,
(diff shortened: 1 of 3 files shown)


saattrupdan (approved):
LGTM, thanks for your contribution!

PR #868, opened 2025-03-31
French: improve AlloCiné spelling

Alkarex:
Correct casing and e-acute

### DIFF
--- a/src/euroeval/dataset_configs.py
+++ b/src/euroeval/dataset_configs.py
@@ -244,7 +244,7 @@ FOSENT_CONFIG = DatasetConfig(
 ALLOCINE_CONFIG = DatasetConfig(
     name="allocine",
     pretty_name="the truncated version of the French sentiment classification "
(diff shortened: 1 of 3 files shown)


src/euroeval/dataset_configs.py
```diff
@@ -244,7 +244,7 @@ def get_dataset_config(dataset_name: str) -> DatasetConfig:
 ALLOCINE_CONFIG = DatasetConfig(
     name="allocine",
```
Alkarex, line 245:
```suggestion
    name="AlloCiné",
```

I am not sure whether it can also be changed here. Would make the header of the Web table nicer

saattrupdan (approved):
Thanks for the correction! The `DatasetConfig.name` attribute doesn't need to be changed, as that's the name used in the CLI, i.e., `euroeval --dataset allocine --model <model-id>`. But the rest are correct 👍

PR #870, opened 2025-03-31
Fix docs for the ScandicQA

oliverkinch:
Fixed docs for ScandiQA-da and ScandiQA-sv, where it was incorrectly stated that the splits were made by considering the original train/validation/test splits.


saattrupdan (approved):
LGTM!

PR #943, opened 2025-04-21
feat: Use async requests for litellm inference

### LINKED ISSUE #942: [FEATURE REQUEST] Speed up ollama inference

**Feature proposal:**
Enable high‑throughput, low‑latency Ollama model inference in EuroEval by leveraging asynchronous requests, server parallelism tuning, and optimized batching.

**Motivation:**
EuroEval’s current Ollama-based benchmark path issues one synchronous HTTP call per sample (SINGLE_SAMPLE batching), serializing requests and inflating round‑trip latency. Users running large-scale evaluations experience significantly slowed end‑to‑end throughput, limiting both development velocity and reproducibility of timely benchmarks.

**Pitch:**
- Asynchronous Inference: Switch from the blocking `Client.generate`/`Client.chat` calls to the Ollama SDK’s AsyncClient and use `asyncio.gather` to fire multiple inferences in parallel. This overlaps network I/O and model compute.
- Server Parallelism Tuning: Expose and document environment variables such as `OLLAMA_NUM_PARALLEL` and `OLLAMA_MAX_LOADED_MODELS` so users can configure how many concurrent requests and model instances Ollama’s backend will handle.
- Batching Preference Adjustment: Update the BatchingPreference for Ollama in `litellm.py` (and any custom modules) from `SINGLE_SAMPLE` to `ALL_AT_ONCE`, allowing EuroEval’s driver to package entire slices of prompts into fewer HTTP requests when safe.
- Optional Native Module: Provide an out‑of‑the‑box `OllamaBenchmarkModule` in `src/euroeval/benchmark_modules` that directly uses [ollama-python](https://github.com/ollama/ollama-python) (sync or async) without the LiteLLM adapter abstraction, minimizing overhead and dependencies.

**Why it matters:**
- Speed: Reduces per‑run benchmark times.

mathiasesn:
Fixes #942.

This pull request refactors the `LiteLLMModel` class in `src/euroeval/benchmark_modules/litellm.py` to improve its handling of batched inputs, introduce asynchronous generation, and enhance the structure of the generated outputs. The key changes include switching to a new batching preference, implementing asynchronous message handling, and restructuring the output generation process to support multiple sequences and scores.

### Batch processing updates:
* Changed `batching_preference` from `BatchingPreference.SINGLE_SAMPLE` to `BatchingPreference.ALL_AT_ONCE`, enabling support for batched inputs.
* The restriction on single-sample batching in the `generate` method was removed, allowing the model to process multiple messages at once.

### Asynchronous generation:
* Replaced the synchronous `litellm.completion_with_retries` call with an asynchronous `_generate_async` method to handle batched message generation. This method includes retry logic and processes responses concurrently. [[1]](diffhunk://#diff-2ce92d47f722b85bf65bfd99b88221ce34380808b0c12da4dbf29ca6efb22f5dL351-R356) [[2]](diffhunk://#diff-2ce92d47f722b85bf65bfd99b88221ce34380808b0c12da4dbf29ca6efb22f5dL455-R533)

### Output restructuring:
* Added the `_create_model_output` helper method to construct `GenerativeModelOutput` objects from multiple responses, supporting multiple sequences and optional scores. [[1]](diffhunk://#diff-2ce92d47f722b85bf65bfd99b88221ce34380808b0c12da4dbf29ca6efb22f5dL455-R533) [[2]](diffhunk://#diff-2ce92d47f722b85bf65bfd99b88221ce34380808b0c12da4dbf29ca6efb22f5dL473-R542) [[3]](diffhunk://#diff-2ce92d47f722b85bf65bfd99b88221ce34380808b0c12da4dbf29ca6efb22f5dL484-R569)
* Ensured that sequences and scores are validated for consistency and handled gracefully when no valid outputs are generated.

### Miscellaneous:
* Added the `asyncio` import to support asynchronous functionality.


src/euroeval/benchmark_modules/litellm.py
```diff
@@ -452,44 +465,135 @@ def generate(self, inputs: dict) -> GenerativeModelOutput:
...
+        Returns:
+            A GenerativeModelOutput object.
+        """
+        sequences = []
+        scores = []
+        for model_response in model_responses:
+            if not isinstance(model_response, ModelResponse):
+                logger.warning(f"Expected a ModelResponse, got {type(model_response)}")
+                continue
+
```
saattrupdan: This check seems redundant, since it cannot be anything else according to the type hints. So I think either this block should be removed, or the type hint should be updated.

```suggestion
```

src/euroeval/benchmark_modules/litellm.py
```diff
@@ -452,44 +465,135 @@ def generate(self, inputs: dict) -> GenerativeModelOutput:
...
-                "any text. This is likely because the model ran out of tokens while "
-                "reasoning. Returning an empty string."
+                "No sequences were generated by the model "
+                f"{self.model_config.model_id!r}. This may be due to the "
+                "model running out of tokens or an issue with the input data. "
+                "Returning an empty GenerativeModelOutput."
+            )
+            return GenerativeModelOutput(sequences=[], scores=None)
+        if scores and len(sequences) != len(scores):
+            raise ValueError(
```
saattrupdan: I'd raise `InvalidBenchmark` here, as that's being handled gracefully "higher up":

```suggestion
            raise InvalidBenchmark(
```

src/euroeval/benchmark_modules/litellm.py
```diff
@@ -348,9 +353,17 @@ def generate(self, inputs: dict) -> GenerativeModelOutput:
...
+                    self._generate_async(
+                        messages=messages,
+                        generation_kwargs=generation_kwargs,
+                        max_retries=3,
+                    )
                 )
+                if failures:
+                    logger.warning(
+                        f"Failures detected during _generate_async: {failures}"
+                    )
```
saattrupdan: Surely we need the generations for these failures to be able to compute scores properly? Since you're already trying to fix the errors in the `_generate_async` function, maybe we should either raise an `InvalidBenchmark` exception here, or replace the failures with empty outputs, no?

src/euroeval/benchmark_modules/litellm.py
```diff
@@ -452,44 +465,135 @@ def generate(self, inputs: dict) -> GenerativeModelOutput:
...
+        model_output = self._create_model_output(model_responses=model_response)
+
+        return model_output
+
+    async def _generate_async(
+        self,
+        messages: list[dict[str, t.Any]],
+        generation_kwargs: dict[str, t.Any],
+        max_retries: int = 3,
+        max_reruns: int = 30,
```
saattrupdan: I'd rather be explicit here, since this is a private function and thus can just be called with all arguments:

```suggestion
        max_retries: int,
        max_reruns: int,
```

src/euroeval/benchmark_modules/litellm.py
```diff
@@ -452,44 +465,135 @@ def generate(self, inputs: dict) -> GenerativeModelOutput:
...
+
+        Args:
+            messages (list[dict[str, t.Any]]):
+                The messages to pass to the model.
+            generation_kwargs (dict[str, t.Any]):
+                The generation kwargs to pass to the model.
+            max_retries (int):
+                The maximum number of retries to make.
+            max_reruns (int):
+                The maximum number of reruns to make.
```
saattrupdan: We leave out types from the docstrings, as they're already mentioned in the type hints. Removing this redundancy prevents potential future errors (it's similar to one not including concrete values in comments, for instance):

```suggestion
            messages:
                The messages to pass to the model.
            generation_kwargs:
                The generation kwargs to pass to the model.
            max_retries:
                The maximum number of retries to make.
            max_reruns:
                The maximum number of reruns to make.
```

src/euroeval/benchmark_modules/litellm.py
```diff
@@ -452,44 +465,135 @@ def generate(self, inputs: dict) -> GenerativeModelOutput:
...
+
+        Returns:
+            A tuple containing the successful responses and the failed responses.
+        """
+        success = []
+        failures = []
+
+        to_run = list(enumerate(messages))
+        for attempt in range(max_reruns + 1):
```
saattrupdan: Maybe instead of having a maximum amount of reruns, we should continue until the number of failures stop decreasing?

src/euroeval/benchmark_modules/litellm.py
```diff
@@ -452,44 +465,135 @@ def generate(self, inputs: dict) -> GenerativeModelOutput:
...
+
+        return success, failures
+
+    def _create_model_output(
+        self, model_responses: list[ModelResponse]
+    ) -> GenerativeModelOutput:
+        """Create a GenerativeModelOutput object from a list of ModelResponse objects.
+
+        Args:
+            model_responses (list[ModelResponse]):
```
saattrupdan: Same as above, that we don't write types in docstrings:

```suggestion
            model_responses:
```

saattrupdan (changes requested):
Great stuff! Really like that you have anticipated many of the potential errors that might occur during generation. I have a few general comments and questions, and a handful of smaller stylistic points.

saattrupdan:
@mathiasesn Feel free to request a review when you're ready 🙂

src/euroeval/benchmark_modules/litellm.py
mathiasesn: This should be handled below by checking if `to_run` is not empty, i.e., there are some failures left, then raising `InvalidBenchmark`. Do you prefer to override the failed outputs with empty outputs?

src/euroeval/benchmark_modules/litellm.py
```diff
@@ -344,151 +361,357 @@ def generate(self, inputs: dict) -> GenerativeModelOutput:
...
-                ServiceUnavailableError,
-                InternalServerError,
-                SystemError,
-            ) as e:
-                logger.debug(
-                    f"Service temporarily unavailable. The error message was: {e}. "
-                    f"Retrying in 5 seconds..."
+
+                for orig_idx, model_response in zip(batch_indices, model_response):
+                    all_responses[orig_idx] = model_response
```
saattrupdan: I would avoid using the same variable name, to prevent silent errors:

```suggestion
                for orig_idx, response in zip(batch_indices, model_response):
                    all_responses[orig_idx] = response
```

src/euroeval/benchmark_modules/litellm.py
```diff
@@ -344,151 +361,357 @@ def generate(self, inputs: dict) -> GenerativeModelOutput:
...
+
+                if not failures:
+                    to_run = []
+                    break
+
+                all_failures.extend(failures)
+                to_run = [(orig_idx, messages[orig_idx]) for orig_idx, _ in failures]
+                logger.warning(
```
saattrupdan: Since it seems like this will often be logged, I think it's a bit much to warning log it, as we would expect it to happen.

```suggestion
                logger.debug(
```

src/euroeval/benchmark_modules/litellm.py
```diff
@@ -344,151 +361,357 @@ def generate(self, inputs: dict) -> GenerativeModelOutput:
...
-        if not model_response.choices:
-            # This happens for reasoning models, when they don't finish thinking and run
-            # out of tokens. Happens quite rarely, but we need to handle it.
-            logger.warning(
-                f"The model {self.model_config.model_id!r} did not end up generating "
-                "any text. This is likely because the model ran out of tokens while "
-                "reasoning. Returning an empty string."
+        if to_run:
+            raise InvalidBenchmark(
+                f"Failed to generate text after {num_attempts + 1} attempts. "
```
saattrupdan:
```suggestion
                f"Failed to generate text after {num_attempts} attempts. "
```

src/euroeval/benchmark_modules/litellm.py
```diff
@@ -344,151 +361,357 @@ def generate(self, inputs: dict) -> GenerativeModelOutput:
...
+        self,
+        error: BadRequestError
+        | RateLimitError
+        | APIError
+        | APIConnectionError
+        | Timeout
+        | ServiceUnavailableError
+        | InternalServerError
+        | SystemError
+        | AuthenticationError,
```
saattrupdan: To avoid redundancy, we could have this function take in all exceptions, where we would just handle it if we can, and raise it otherwise?

```suggestion
        error: Exception,
```

src/euroeval/benchmark_modules/litellm.py
```diff
@@ -344,151 +361,357 @@ def generate(self, inputs: dict) -> GenerativeModelOutput:
...
+        generation_kwargs: dict[str, t.Any],
+        stop_messages: list[str],
+        logprobs_messages: list[str],
+        temperature_messages: list[str],
+        temperature_must_be_one_messages: list[str],
+        max_items_messages: list[str],
+        no_json_schema_messages: list[str],
+    ) -> None:
```
saattrupdan, line 408: Missing docstring

src/euroeval/benchmark_modules/litellm.py
```diff
@@ -344,151 +361,357 @@ def generate(self, inputs: dict) -> GenerativeModelOutput:
...
+        | InternalServerError
+        | SystemError
+        | AuthenticationError,
+        generation_kwargs: dict[str, t.Any],
+        stop_messages: list[str],
+        logprobs_messages: list[str],
+        temperature_messages: list[str],
+        temperature_must_be_one_messages: list[str],
+        max_items_messages: list[str],
+        no_json_schema_messages: list[str],
```
saattrupdan: I think all these messages should just be defined within this function, as it's the only place where they're used anyway.

src/euroeval/benchmark_modules/litellm.py
```diff
@@ -344,151 +361,357 @@ def generate(self, inputs: dict) -> GenerativeModelOutput:
...
+    ) -> None:
+        msg = str(error).lower()
+        model_id = self.model_config.model_id
+
+        def disable_stop() -> None:
```
saattrupdan: I think all of these functions and subsequent `handlers` is a bit too much unnecessary abstraction. A simple `if`/`elif`/`else`  or a `match-case` check would do.

src/euroeval/benchmark_modules/litellm.py
```diff
@@ -344,151 +361,357 @@ def generate(self, inputs: dict) -> GenerativeModelOutput:
...
+        generation_kwargs: dict[str, t.Any],
+        stop_messages: list[str],
+        logprobs_messages: list[str],
+        temperature_messages: list[str],
+        temperature_must_be_one_messages: list[str],
+        max_items_messages: list[str],
+        no_json_schema_messages: list[str],
+    ) -> None:
+        msg = str(error).lower()
+        model_id = self.model_config.model_id
```
saattrupdan, line 418: This is the only time `self` is used, so I think it'd be better if we could separate this as an independent function and take in `model_id` as an extra argument. This would make it easier to test as well.

src/euroeval/benchmark_modules/litellm.py
```diff
@@ -1093,3 +1316,23 @@ def try_download_ollama_model(model_id: str) -> bool:
             level=logging.DEBUG,
         )
         return True
+
+def safe_run(
+    coroutine: t.Coroutine[t.Any, t.Any, ModelResponse | list[ModelResponse]],
+) -> ModelResponse | list[ModelResponse]:
```
saattrupdan: This return type doesn't seem to match that it's really returning a tuple of model responses and failures?

src/euroeval/benchmark_modules/litellm.py
```diff
@@ -344,151 +361,357 @@ def generate(self, inputs: dict) -> GenerativeModelOutput:
...
+                            logprobs_messages=logprobs_messages,
+                            temperature_messages=temperature_messages,
+                            temperature_must_be_one_messages=temperature_must_be_one_messages,
+                            max_items_messages=max_items_messages,
+                            no_json_schema_messages=no_json_schema_messages,
+                        )
+                    else:
+                        raise error  # Unhandleable error, raise it
+
+            except self._handleable_exceptions as e:
```
saattrupdan: Is this try-except block needed at all, since we can handle all exceptions in the block above?

src/euroeval/benchmark_modules/litellm.py
```diff
@@ -344,151 +361,357 @@ def generate(self, inputs: dict) -> GenerativeModelOutput:
...
                 )
-                sleep(5)
-            except AuthenticationError:
-                raise NeedsAdditionalArgument(
-                    cli_argument="--api-key",
-                    script_argument="api_key=<your-api-key>",
-                    run_with_cli=self.benchmark_config.run_with_cli,
+
+                for _, error in failures:
+                    if isinstance(error, self._handleable_exceptions):
```
saattrupdan: Why have this check, since the `_handle_exception` function already raises the function if it is not handleable?

src/euroeval/benchmark_modules/litellm.py
saattrupdan: I guess it really returns the same as the `coroutine` argument?

src/euroeval/benchmark_modules/litellm.py
```diff
@@ -344,151 +361,357 @@ def generate(self, inputs: dict) -> GenerativeModelOutput:
...
+        failures = []
+
+        to_run = list(enumerate(messages))
+
+        prev_fail_count = len(to_run)
+        rerun_count = 0
+
+        while to_run and rerun_count < max_reruns and prev_fail_count > 0:
+            if not to_run:
+                break
```
saattrupdan: This will never trigger, since `to_run` has to be non-empty for the loop to even run, no?

src/euroeval/benchmark_modules/litellm.py
```diff
@@ -344,151 +361,357 @@ def generate(self, inputs: dict) -> GenerativeModelOutput:
...
+            for (orig_idx, _), response in zip(to_run, responses):
+                if isinstance(response, Exception):
+                    current_fail_count += 1
+                    next_to_run.append((orig_idx, messages[orig_idx]))
+                else:
+                    success.append(response)
+
+            if current_fail_count >= prev_fail_count:
+                failures.extend(
+                    (idx, Exception("Exceeded adaptive rerun threshold"))
```
saattrupdan: If I understand this correctly, this exception should break the evaluation. If this is the case, then I think it'd be cleaner if we raise an `InvalidBenchmark` exception (from the `exceptions` module), as that's handled cleanly upstream, without breaking all ongoing evaluations.

src/euroeval/benchmark_modules/litellm.py
```diff
@@ -344,151 +361,357 @@ def generate(self, inputs: dict) -> GenerativeModelOutput:
...
+                )
+                break
+
+            prev_fail_count = current_fail_count
+            to_run = next_to_run
+            rerun_count += 1
+
+        for idx, _ in to_run:
+            failures.append(
+                (idx, Exception("Max reruns reached without further improvement"))
```
saattrupdan: If I understand this correctly, this exception should break the evaluation. If this is the case, then I think it'd be cleaner if we raise an `InvalidBenchmark` exception (from the `exceptions` module), as that's handled cleanly upstream, without breaking all ongoing evaluations.

src/euroeval/benchmark_modules/litellm.py
```diff
@@ -344,151 +361,357 @@ def generate(self, inputs: dict) -> GenerativeModelOutput:
...
+            )
+
+        return success, failures
+
+    def _create_model_output(
+        self, model_responses: list[ModelResponse]
```
saattrupdan: The only place `self` is used here is to get the model ID. Again I think it's better to make that an explicit argument and convert this method to an independent function.

src/euroeval/benchmark_modules/litellm.py
```diff
@@ -1093,3 +1316,23 @@ def try_download_ollama_model(model_id: str) -> bool:
             level=logging.DEBUG,
         )
         return True
+
+def safe_run(
```
saattrupdan: Can this be moved to the `utils` module? Since it's a very general async-handling function.

saattrupdan (changes requested):
Nice work! I have a bunch of comments and suggestions, most of which concerns cleaning up the code a bit.

src/euroeval/benchmark_modules/litellm.py
mathiasesn: Should it be something like this or?

```python
T = t.TypeVar("T", bound=object)


def safe_run(coroutine: t.Coroutine[t.Any, t.Any, T]) -> T:
    """Run a coroutine, ensuring that the event loop is always closed when we're done.

    Args:
        coroutine:
            The coroutine to run.

    Returns:
        The result of the coroutine.
    """
    loop = asyncio.new_event_loop()
    try:
        asyncio.set_event_loop(loop)
        return loop.run_until_complete(coroutine)
    finally:
        loop.close()
        asyncio.set_event_loop(None)
  ```

File: src/euroeval/benchmark_modules/litellm.py (hunk 18 above)
Comment (mathiasesn) [reply src/euroeval/benchmark_modules/litellm.py]: The `_handle_exception` function currently raises `InvalidBenchmark` if it cannot handle the exception, which might not be the same original exception.

File: src/euroeval/benchmark_modules/litellm.py (hunk 17 above)
Comment (mathiasesn) [reply src/euroeval/benchmark_modules/litellm.py]: No, it might not be, since all the exceptions should be added to `failures` instead of raises by the event loop.

**Comment by mathiasesn** [comment]
Ran debugging with `python -m euroeval.cli --clear-model-cache -m ollama_chat/hf.co/unsloth/Qwen3-0.6B-GGUF` ad here are the `euroeval_benchmark_results.jsonl`. I did not finish the benchmark, but it runs fine.

[euroeval_benchmark_results.zip](https://github.com/user-attachments/files/20029556/euroeval_benchmark_results.zip)

**Review by saattrupdan — COMMENTED** [review]
Much better now, looks great!

I tried installing from source, and I didn't have any progress bar within an iteration. Is it gone on your end as well? And if so, can we get that one back?

File: src/euroeval/benchmark_modules/litellm.py [hunk 24]
```diff
@@ -341,153 +428,230 @@ def generate(self, inputs: dict) -> GenerativeModelOutput:
...
                 )
-                sleep(5)
-            except AuthenticationError:
-                raise NeedsAdditionalArgument(
-                    cli_argument="--api-key",
-                    script_argument="api_key=<your-api-key>",
-                    run_with_cli=self.benchmark_config.run_with_cli,
+                for _, msg in to_run
+            ]
+            responses = await asyncio.gather(*requests, return_exceptions=True)
```
saattrupdan: [Supposedly](https://stackoverflow.com/a/75475366/3154226), to get a progress bar during the gather operation, we can replace this `asyncio.gather` call with `tqdm_async.gather`, where `tqdm_async` is imported as

```python
from tqdm.asyncio import tqdm as tqdm_async
```

saattrupdan (changes requested):

saattrupdan:
@mathiasesn Surprisingly, the `tqdm_async.gather` doesn't support the `return_exceptions` argument (it is an [open issue](https://github.com/tqdm/tqdm/issues/1286)).

We could fix this manually by removing that argument and instead having a generic wrapper for the requests:

```python
async def return_exceptions(f):
    try:
        return await f
    except Exception as e:
        return e
```

Mikeriess:
Appreciate the effort 💪💪

mathiasesn:
> @mathiasesn Surprisingly, the `tqdm_async.gather` doesn't support the `return_exceptions` argument (it is an [open issue](https://github.com/tqdm/tqdm/issues/1286)).
> 
> We could fix this manually by removing that argument and instead having a generic wrapper for the requests:
> 
> ```python
> async def return_exceptions(f):
>     try:
>         return await f
>     except Exception as e:
>         return e
> ```

@saattrupdan Added `catch_coroutine_exception` to utils.

mathiasesn:
Running on branch `main` with total time of 07:05 and 42.50 s/it.

```
euroeval --clear-model-cache -m ollama_chat/hf.co/unsloth/Qwen3-0.6B-GGUF --dataset angry-tweets
2025-05-06 21:46:01 ⋅ Flash attention has not been installed, so this will not be used. To install it, run `pip install -U wheel && FLASH_ATTENTION_SKIP_CUDA_BUILD=TRUE pip install flash-attn --no-build-isolation`. Alternatively, you can disable this message by setting the flag `--no-use-flash-attention`.
2025-05-06 21:46:02 ⋅ Few-shot benchmarking ollama_chat/hf.co/unsloth/Qwen3-0.6B-GGUF on the validation split of the truncated version of the Danish sentiment classification dataset AngryTweets
2025-05-06 21:46:02 ⋅ Loading model...
2025-05-06 21:46:02 ⋅ The model has 596,049,920 parameters, an unknown vocabulary size, and a maximum context length of 40,960.
Preparing datasets: 100%|█████████████████████████████████████████████████████████████████████████████████| 10/10 [00:01<00:00,  6.60it/s]
Benchmarking: 100%|███████████████████████████████████████████████████████████████████████████████████████| 10/10 [07:05<00:00, 42.50s/it]
2025-05-06 21:53:12 ⋅ Finished evaluation of ollama_chat/hf.co/unsloth/Qwen3-0.6B-GGUF on the truncated version of the Danish sentiment classification dataset AngryTweets.                                                                                                         
2025-05-06 21:53:12 ⋅ Matthew's Correlation Coefficient: 0.00% ± 0.00%
2025-05-06 21:53:12 ⋅ Macro-average F1-score: 16.06% ± 0.70%
2025-05-06 21:53:12 ⋅ Finished 1 out of 1 benchmarks.
```

Running on branch `feat/use-async-requests-for-ollama-inference` with total time of 06:42 and 40.27 s/it.

```
euroeval --clear-model-cache -m ollama_chat/hf.co/unsloth/Qwen3-0.6B-GGUF --dataset angry-tweets
2025-05-06 21:54:52 ⋅ Flash attention has not been installed, so this will not be used. To install it, run `pip install -U wheel && FLASH_ATTENTION_SKIP_CUDA_BUILD=TRUE pip install flash-attn --no-build-isolation`. Alternatively, you can disable this message by setting the flag `--no-use-flash-attention`.
2025-05-06 21:54:53 ⋅ Few-shot benchmarking ollama_chat/hf.co/unsloth/Qwen3-0.6B-GGUF on the validation split of the truncated version of the Danish sentiment classification dataset AngryTweets
2025-05-06 21:54:53 ⋅ Loading model...
2025-05-06 21:54:53 ⋅ The model has 596,049,920 parameters, an unknown vocabulary size, and a maximum context length of 40,960.
Preparing datasets: 100%|█████████████████████████████████████████████████████████████████████████████████| 10/10 [00:01<00:00,  6.63it/s]
Benchmarking: 100%|███████████████████████████████████████████████████████████████████████████████████████| 10/10 [06:42<00:00, 40.27s/it]
2025-05-06 22:01:42 ⋅ Finished evaluation of ollama_chat/hf.co/unsloth/Qwen3-0.6B-GGUF on the truncated version of the Danish sentiment classification dataset AngryTweets.
2025-05-06 22:01:42 ⋅ Matthew's Correlation Coefficient: 0.00% ± 0.00%
2025-05-06 22:01:42 ⋅ Macro-average F1-score: 16.06% ± 0.70%
2025-05-06 22:01:42 ⋅ Finished 1 out of 1 benchmarks.
```

I observed extended periods of 0% GPU utilization between benchmark runs,  which might indicate potential for performance improvements.

image

src/euroeval/utils.py
```diff
@@ -350,3 +350,19 @@ def safe_run(coroutine: t.Coroutine[t.Any, t.Any, T]) -> T:
     finally:
         loop.close()
         asyncio.set_event_loop(None)
+
+
+async def catch_coroutine_exception(coroutine: t.Coroutine[t.Any, t.Any, T]) -> T:
```
saattrupdan:
```suggestion
async def catch_coroutine_exception(coroutine: t.Coroutine[t.Any, t.Any, T]) -> T | Exception:
```

src/euroeval/benchmark_modules/litellm.py
```diff
@@ -279,6 +284,8 @@ def generate(self, inputs: dict) -> GenerativeModelOutput:
         if self.dataset_config.task in TASKS_USING_JSON:
             for msg_list in messages:
                 # msg_list is a list of {'role':…, 'content':…} dicts
+                if not msg_list:
+                    raise ValueError("Encountered an empty message list in 'messages'.")
```
saattrupdan:
```suggestion
                    raise InvalidBenchmark("Encountered an empty message list in 'messages'.")
```

saattrupdan (changes requested):
Just tested it, works great with the progress bar now!

A couple of smaller suggested changes to your new additions, and then we're ready to merge 🙂

saattrupdan:
> I observed extended periods of 0% GPU utilization between benchmark runs, which might indicate potential for performance improvements.

Interesting. But let's deal with that in a separate issue.

mathiasesn:
> Interesting. But let's deal with that in a separate issue.

Of course.

src/euroeval/benchmark_modules/litellm.py
```diff
@@ -341,153 +435,233 @@ def generate(self, inputs: dict) -> GenerativeModelOutput:
...
+
+            for (orig_idx, _), response in zip(to_run, responses):
+                if isinstance(response, Exception):
+                    current_fail_count += 1
+                    all_failures[orig_idx] = response
+                    next_to_run.append((orig_idx, messages[orig_idx]))
+                else:
+                    success.append(response)
+
+            if current_fail_count >= prev_fail_count:
```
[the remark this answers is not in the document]
mathiasesn: @saattrupdan, what say you to this suggestion?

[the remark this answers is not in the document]
saattrupdan: Sounds good 🙂

saattrupdan (approved):
LGTM!

PR #965, opened 2025-05-01
Feat: Penalize summarization scores for wrong language

### LINKED ISSUE #614: [BUG] BERTScore should use monolingual models
### 🐛 Describe the bug

For languages that have strong encoder models, these should be used instead of the current multilingual one, as this currently doesn't penalise the models for summarising in the wrong language. 

### Operating System

Linux

### Device

CUDA GPU

### Python version

3.10.x

### ScandEval version

13.3.0

viggo-gascou:
This PR fixes #614, by adding a penalization to the metrics computed for the summarization task.

Right now there is no lower limit for the confidence score before applying the penalization. From testing the different languages supported in EuroEval, the lowest score I got for a Sentence/sample that was in the correct language, was `0.88` for a relatively short summary in Danish. So a lower limit of `0.85`, might be adequate? In general the confidence score for a given sentence is affected by the length of the sentence.


src/euroeval/task_group_utils/text_to_text.py
```diff
@@ -69,6 +71,17 @@ def compute_metrics(
     else:
         predictions = model_outputs
 
+    # Set up languages and language detector with lingua
+    target_langs = get_correct_language_codes(
+        [lang.code for lang in dataset_config.languages]
+    )
+    lingua_langs = [IsoCode639_1.from_str(lang_code) for lang_code in target_langs]
```
saattrupdan:
```suggestion
    lingua_langs = [IsoCode639_1.from_str(lang_code) for lang_code in target_langs]
    
```

src/euroeval/task_group_utils/text_to_text.py
```diff
@@ -69,6 +71,17 @@ def compute_metrics(
     else:
         predictions = model_outputs
 
+    # Set up languages and language detector with lingua
+    target_langs = get_correct_language_codes(
+        [lang.code for lang in dataset_config.languages]
```
saattrupdan:
```suggestion
        language_codes=[lang.code for lang in dataset_config.languages]
```

saattrupdan (changes requested):
Looks good!

Again, we need to think about whether we want to simply multiple the language confidence score with the base score, or whether we want to make it binary instead, to give the model a score of 0 if the language confidence is below a certain threshold.

Also, since you mentioned that the language identification with `.from_all_languages()` is quite slow, we should experiment with using only a single language and see if we get the same confidence scores. When there are multiple dataset languages, we would just run each language in 'single language mode', and take sum and/or max.

viggo-gascou:
After doing some more extensive testing with a larger sample, I discovered that the slowness occurs primarily during the initial loading of the model, not during inference itself. When using target summaries from the test split (2048 samples) of the`ilpost-sum` dataset, the model processed all sentences in approximately 0.25 seconds on average (based on 1000 repetition tests), the slowest being ~0.7s.

I also tried running the model in single language mode and noticed it often gives a perfect confidence score of 1 even when the text isn't actually in that language.

The low accuracy setting is a bit faster (~0.17s on average, ~0.6 maximum), but since it is only the initial loading of the model, I don't think it's worth sacrificing accuracy for that small speed boost. Especially when the accuracy of sentences below 120 characters is significantly worse, see [here](https://github.com/pemistahl/lingua-py#115-low-accuracy-mode-versus-high-accuracy-mode), and we do have summaries that are shorter than that in our datasets.

src/euroeval/task_group_utils/text_to_text.py
```diff
@@ -117,8 +131,21 @@ def compute_metrics(
         # process is not the main process
         if score_dict is not None:
             scores = score_dict[cfg.results_key]
+            corrected_score = 0.0
+            conf_vals = detector.compute_language_confidence_values_in_parallel(
+                predictions
+            )
             if isinstance(scores, list):
```
saattrupdan: Should we include a check if the scores is not a list? Maybe average language confidence multiplied with the single score?

src/euroeval/task_group_utils/text_to_text.py
```diff
@@ -117,8 +131,21 @@ def compute_metrics(
...
-                scores = sum(scores) / len(scores)
+                for score, confidence in zip(scores, conf_vals):
+                    # sum the confidence values so we get a single value for datasets
+                    # with 'multiple' languages, mostly relevant for Norwegian datasets
+                    sample_confidence = sum(
+                        conf.value
+                        for conf in confidence
+                        if conf.language.iso_code_639_1 in lingua_langs
+                    )
+                    corrected_score += score * sample_confidence
```
saattrupdan: As we talked about, I think the best penalisation method here would be to make it binary. If the language confidence is above a certain threshold then we do not penalise at all, and if it is below then the model gets a score of 0% for that sample. The benefit of that is also that the resulting scores should be basically identical to the current ones, except if the model outputs summaries in a wrong language.

Towards finding the threshold, you made some experiments - could you implement this change with that threshold? Then I can do some tests on your branch to ensure that it works as it should and we can tweak the threshold if needed.

saattrupdan (changes requested):

viggo-gascou:
Now all we should need to do is some testing to figure out what the threshold should be, I set it to 0.75 for now @saattrupdan

src/euroeval/task_group_utils/text_to_text.py
```diff
@@ -69,6 +71,22 @@ def compute_metrics(
...
+        language_codes=[lang.code for lang in dataset_config.languages]
+    )
+    lingua_langs = [IsoCode639_1.from_str(lang_code) for lang_code in target_langs]
+
+    # We use the detector with all latin languages, since we do not want to be in
+    # single language mode for datasets that only have a single language.
+    # This is important for datasets that have 'multiple' languages, such as Norwegian.
+    # See: https://github.com/pemistahl/lingua-py#116-single-language-mode
+    detector = (
+        LanguageDetectorBuilder.from_all_languages_with_latin_script()
```
saattrupdan: How big of a difference does it make with latin script vs non-latin? I'm just thinking that if the model starts outputting Chinese (which might not be unreasonable from Qwen- or Deepseek models), then that should maybe be included. So, if it doesn't take much longer with all languages, then we should probably include all.

src/euroeval/task_group_utils/text_to_text.py
```diff
@@ -69,6 +71,22 @@ def compute_metrics(
...
+        .with_preloaded_language_models()
```
saattrupdan: Does this download language models for each language?

saattrupdan (changes requested):

src/euroeval/task_group_utils/text_to_text.py
viggo-gascou: I am not sure if they are downloaded separately or if they are a part of the package itself. But, the biggest thing is that it preloads all of the models into memory, otherwise it seems that it would load them when needed, see [here](https://github.com/pemistahl/lingua-py?tab=readme-ov-file#114-eager-loading-versus-lazy-loading). So this keeps the latency down when checking the samples below.

src/euroeval/task_group_utils/text_to_text.py
viggo-gascou: I haven't made tests for these, I also did this mostly because they mention on their docs that "_In most cases, it is not advisable to build the detector from all supported languages. When you have knowledge about the texts you want to classify you can almost always rule out certain languages as impossible or unlikely to occur._". They do have a bunch of different pre-defined categories [see here](https://github.com/pemistahl/lingua-py?tab=readme-ov-file#119-methods-to-build-the-languagedetector), and [in the source](https://github.com/pemistahl/lingua-rs/blob/3d914ec9a47224c08906e7c82ca5c479fdff50aa/src/language.rs#L825C39-L825C46).

viggo-gascou: And also because I mostly thought about the languages that we have in our benchmarks 😅

saattrupdan:
@viggo-gascou The only thing that's left here before we can merge are some concrete evaluation tests, to ensure that it doesn't mess up existing evaluations. Can you run the evaluation on the Danish summarisation dataset with the following models? They should have the same score as on the leaderboards, roughly speaking:

- [deepcogito/cogito-v1-preview-llama-8B (few-shot)](https://huggingface.co/deepcogito/cogito-v1-preview-llama-8B)
- [allenai/Llama-3.1-Tulu-3-8B-SFT (few-shot)](https://huggingface.co/allenai/Llama-3.1-Tulu-3-8B-SFT)
- [meta-llama/Llama-3.1-8B-Instruct (few-shot)](https://huggingface.co/meta-llama/Llama-3.1-8B-Instruct)

And as another test, this one should get a significantly worse score than normal, as it mainly outputs Swedish:

- [AI-Sweden-Models/Llama-3-8B-instruct (few-shot)](https://huggingface.co/AI-Sweden-Models/Llama-3-8B-instruct)

saattrupdan:
@viggo-gascou What's the status on this PR?

saattrupdan:
@viggo-gascou Closing this PR for now, as it's been stale for a while. Feel free to re-open when the tests have been conducted and it's ready for review.

PR #966, opened 2025-05-01
Fix missing separators for Finnish

viggo-gascou:
Noticed that the 'and' and 'or' separators were missing for Finnish 😅


saattrupdan (approved):
LGTM!

PR #970, opened 2025-05-03
BeleBele documentation

oliverkinch:
This PR adds documentation for the BeleBele datasets that we currently have in EuroEval.


oliverkinch:
We are currently not supporting BeleBele in Italian and Finnish even though BeleBele is also available in these languages (see table 4 in [the paper](https://aclanthology.org/2024.acl-long.44.pdf)). Should I add Italian and Finnish to `src/scripts/create_belebele.py` or is there a reason why these are not already added? @saattrupdan

docs/datasets/danish.md
```diff
@@ -624,6 +624,70 @@ When evaluating generative models, we use the following setup (see the
   ```
 
 
+### Unofficial: BeleBele
```
saattrupdan: Need the country code in the name, so that there's a correspondence between the title and the command one needs to run to evaluate a model on it:

```suggestion
### Unofficial: BeleBele-da
```

Same thing with the others.

saattrupdan: The BeleBele dataset is a reading comprehension dataset, phrased as multiple choice, so we should move this to that section.

saattrupdan (changes requested):
Thanks! A few small changes and it's ready to merge 🙂

saattrupdan:
> We are currently not supporting BeleBele in Italian and Finnish even though BeleBele is also available in these languages (see table 4 in [the paper](https://aclanthology.org/2024.acl-long.44.pdf)). Should I add Italian and Finnish to `src/scripts/create_belebele.py` or is there a reason why these are not already added? @saattrupdan 

Please add these as well yeah, it was purely because we didn't support these languages at the time.

docs/datasets/danish.md
oliverkinch: I just found the MC base promp template to be a better fit for BeleBele, but we should instead use the RC base promp template? E.g. the following one

- Number of few-shot examples: 4
- Prefix prompt:
  ```
  Følgende er tekster med tilhørende spørgsmål og svar.
  ```
- Base prompt template:
  ```
  Tekst: {text}
  Spørgsmål: {question}
  Svar med maks. 3 ord: {label}
  ```
- Instruction-tuned prompt template:
  ```
  Tekst: {text}

  Besvar følgende spørgsmål om teksten ovenfor med maks. 3 ord.

  Spørgsmål: {question}
  ```

oliverkinch:
Note: I forgot the Swedish BeleBele, and I should also add the Spanish BeleBele

oliverkinch:
> > We are currently not supporting BeleBele in Italian and Finnish even though BeleBele is also available in these languages (see table 4 in [the paper](https://aclanthology.org/2024.acl-long.44.pdf)). Should I add Italian and Finnish to `src/scripts/create_belebele.py` or is there a reason why these are not already added? @saattrupdan
> 
> Please add these as well yeah, it was purely because we didn't support these languages at the time.

I have added them here https://github.com/EuroEval/EuroEval/pull/972

docs/datasets/danish.md
saattrupdan: I definitely agree that we should use the multiple choice prompt templates, I just meant that it should be moved into the "Reading Comprehension" section of the docs, as that's what the dataset is measuring 🙂

oliverkinch:
I have now added documentation for all languages except Faroese.

docs/datasets/english.md
```diff
@@ -350,38 +350,30 @@ $ euroeval --model <model-id> --dataset belebele
 ```
 
 
-## Knowledge
-
-### MMLU
+### Unofficial: BeleBele
```
saattrupdan: English one should also have an "-en" ending, as it was a natively multilingual dataset, rather than an English dataset that was translated.

I know that it is currently set to just "belebele" for English, but that was my mistake 🙈 

```suggestion
### Unofficial: BeleBele-en
```

docs/datasets/english.md
```diff
@@ -418,37 +410,41 @@ When evaluating generative models, we use the following setup (see the
 You can evaluate this dataset directly as follows:
 
 ```bash
-$ euroeval --model <model-id> --dataset mmlu
+$ euroeval --model <model-id> --dataset belebele
```
saattrupdan:
```suggestion
$ euroeval --model <model-id> --dataset belebele-en
```

saattrupdan (changes requested):
Small change regarding English Belebele. Can you also change the name in the dataset config to "belebele-en"?

oliverkinch:
> Small change regarding English Belebele. Can you also change the name in the dataset config to "belebele-en"?

Done.

Also, for some reason I had included the documentation for the English BeleBele twice - not sure how that happened, but it is fixed now

saattrupdan (approved):
LGTM!

PR #972, opened 2025-05-03
Add more languages for the BeleBele dataset

oliverkinch:
Add the Finnish, Italian, and Spanish BeleBele datasets.

Documentation for these datasets will be added in https://github.com/EuroEval/EuroEval/pull/970

### DIFF
--- a/src/euroeval/dataset_configs/finnish.py
+++ b/src/euroeval/dataset_configs/finnish.py
@@ -61,7 +61,6 @@ SCALA_FI_CONFIG = DatasetConfig(
 
 ### Unofficial datasets ###
 
-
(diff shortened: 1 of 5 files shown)


src/euroeval/dataset_configs/finnish.py
```diff
@@ -60,3 +60,14 @@
 )
 
 ### Unofficial datasets ###
+
```
saattrupdan, line 61:
```suggestion
```

saattrupdan (approved):
LGTM, only a tiny nitpick.

saattrupdan (approved):
LGTM

PR #985, opened 2025-05-09
feat: add concurrency group for ci workflow

viggo-gascou:
This PR adds a concurrency group to the CI workflow to prevent overlapping runs of the same workflow on different commits for the same branch.

If a previous commit in a PR has already triggered the CI workflow and is still running, any new commit to that PR will automatically cancel the ongoing workflow and start a fresh run from the latest commit.


saattrupdan (approved):
LGTM

PR #1152, opened 2025-08-13
Feat: add a `--download-only` flag for downloading models, datasets and metrics

### LINKED ISSUE #988: [FEATURE REQUEST] <name-of-feature>

As mentioned in https://github.com/EuroEval/EuroEval/discussions/987, it would be great to have a `--only-download` flag, which doesn't run any evaluations, but just downloads all the required files to disk. This will enable the evaluation to then be run without any internet access, which is required on some compute clusters.

viggo-gascou:
This PR fixes #988, by adding a `--download-only` flag to skip benchmarking and only download the required models, datasets and metrics. 

So, if you run your `euroeval` command with the `--download-only` flag set, on a session with internet access, all of the required assets will be downloaded to the `--cache-dir` directory (default `.euroeval_cache`), and no benchmarks will be run. Then subsequent runs of the `euroeval` commands with the same models and datasets will be run in "offline" mode enabling the ability to run EuroEval benchmarks on e.g., compute clusters without internet access.

### DIFF
--- a/src/euroeval/benchmarker.py
+++ b/src/euroeval/benchmarker.py
@@ -242,12 +242,28 @@ class Benchmarker:
     def benchmark_results(self) -> list[BenchmarkResult]:
         """The benchmark results."""
         if self.results_path.exists():
+            benchmark_results: list[BenchmarkResult] = list()
             with self.results_path.open() as f:
-                return [
-                    BenchmarkResult.from_dict(json.loads(line))
-                    for line in f
-                    if line.strip()
-                ]
+                for line in f:
+                    if line.strip():
+                        result_dict = json.loads(line.strip())
+
+                        # Fix for older records
+                        has_old_raw_results = (
+                            "results" in result_dict
+                            and isinstance(result_dict["results"], dict)
+                            and "raw" in result_dict["results"]
+                            and isinstance(result_dict["results"]["raw"], dict)
+                            and "test" in result_dict["results"]["raw"]
+                        )
+                        if has_old_raw_results:
+                            result_dict["results"]["raw"] = result_dict["results"][
+                                "raw"
+                            ]["test"]
+
+                        result = BenchmarkResult.from_dict(result_dict)
+                        benchmark_results.append(result)
+            return benchmark_results
         else:
             return list()
 
--- a/src/euroeval/utils.py
+++ b/src/euroeval/utils.py
@@ -152,6 +152,9 @@ def block_terminal_output() -> None:
     libraries, disabled tokeniser progress bars when using Hugging Face tokenisers, and
     disables most of the logging from the `transformers` library.
     """
+    if os.getenv("FULL_LOG") == "1":
+        return
+
     # Ignore miscellaneous warnings
     warnings.filterwarnings("ignore", category=UserWarning)
     warnings.filterwarnings("ignore", category=FutureWarning)
--- a/.github/ISSUE_TEMPLATE/benchmark_dataset_request.yaml
+++ b/.github/ISSUE_TEMPLATE/benchmark_dataset_request.yaml
@@ -34,6 +34,7 @@ body:
       - label: Italian
       - label: Latvian
       - label: Norwegian (Bokmål or Nynorsk)
+      - label: Polish
       - label: Portuguese
       - label: Spanish
       - label: Swedish
--- a/.github/ISSUE_TEMPLATE/language_request.yaml
+++ b/.github/ISSUE_TEMPLATE/language_request.yaml
@@ -1,0 +1,49 @@
+name: 🌍 Language Request
+description: Is there a European language missing in EuroEval?
+title: "[LANGUAGE REQUEST] <language-name>"
+labels: "new language"
+type: task
+
+body:
+- type: input
+  attributes:
+    label: Language name and code
+    description: What is the name and ISO 639 code of the language?
+  validations:
+    required: true
+- type: markdown
+  attributes:
+    value: >
+      Here are some existing evaluation datasets in the language, that could be used:
+- type: textarea
+  attributes:
+    label: Sentiment classification dataset
+    description: Link to one or more datasets in the language (leave blank if unknown)
+- type: textarea
+  attributes:
+    label: Linguistic acceptability dataset
+    description: Link to one or more datasets in the language (leave blank if unknown)
+- type: textarea
+  attributes:
+    label: Named entity recognition dataset
+    description: Link to one or more datasets in the language (leave blank if unknown)
+- type: textarea
+  attributes:
+    label: Reading comprehension dataset
+    description: Link to one or more datasets in the language (leave blank if unknown)
+- type: textarea
+  attributes:
+    label: Summarisation dataset
+    description: Link to one or more datasets in the language (leave blank if unknown)
+- type: textarea
+  attributes:
+    label: Knowledge dataset
+    description: Link to one or more datasets in the language (leave blank if unknown)
+- type: textarea
+  attributes:
+    label: Common-sense reasoning dataset
+    description: Link to one or more datasets in the language (leave blank if unknown)
+- type: markdown
+  attributes:
+    value: >
+      Thanks for contributing 🎉!
--- a/.github/ISSUE_TEMPLATE/model_evaluation_request.yaml
+++ b/.github/ISSUE_TEMPLATE/model_evaluation_request.yaml
@@ -23,6 +23,7 @@ body:
       - label: West Germanic languages (Dutch, English, German)
       - label: Finnic languages (Estonian, Finnish)
       - label: Latvian
+      - label: Polish
   validations:
     required: true
 - type: dropdown
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,20 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 
 ## [Unreleased]
+### Added
+- Added the `timm` package to the set of `generative` extra dependencies, as it is
+  required to load some multimodal models, such as Gemma-3n.
 
+### Changed
+- Now does not benchmark encoder models on multiple-choice classification tasks, as they
+  get near-random performance and these scores are not used in the leaderboards. We can
+  change this in the future if we find a way to make encoder models work better on these
+  tasks.
+
+### Fixed
+- Fixed an issue where old evaluation records could not be loaded, as the format had
+  changed. We are now able to load old records again.
+- Fixed some grammatical errors in the Icelandic prompts.
 
 
 ## [v16.1.1] - 2025-09-12
--- a/docs/datasets/icelandic.md
+++ b/docs/datasets/icelandic.md
@@ -44,11 +44,11 @@ When evaluating generative models, we use the following setup (see the
 - Number of few-shot examples: 12
 - Prefix prompt:
   ```
-  Eftirfarandi eru yfirferðir ásamt lyndisgildi þeirra, sem getur verið 'jákvætt', 'hlutlaust' eða 'neikvætt'.
+  Hér fyrir neðan eru textabrot ásamt lyndisgildi þeirra sem getur verið 'jákvætt', 'hlutlaust' eða 'neikvætt'.
   ```
 - Base prompt template:
   ```
-  Yfirferð: {text}
+  Textabrot: {text}
   Lyndi: {label}
   ```
 - Instruction-tuned prompt template:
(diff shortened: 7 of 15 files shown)


src/euroeval/benchmarker.py
```diff
@@ -453,46 +493,55 @@ def benchmark(
...
-                    and benchmark_config.raise_errors
-                ):
-                    raise benchmark_output_or_err
+                if benchmark_config.download_only:
+                    self._download(
+                        model_config=model_config,
+                        dataset_config=dataset_config,
+                        benchmark_config=benchmark_config,
+                    )
+                else:
```
saattrupdan: Nit: To avoid too much indentation, you can add a `continue` after the download, and remove the else statement.

src/euroeval/metrics.py
```diff
@@ -49,9 +50,17 @@ def __init__(
             else lambda x: (100 * x, f"{x:.2%}")
         )
 
+    def download(self, cache_dir: str) -> "Metric":
+        """Initiates the download of the metric if needed."""
+        return self
```
saattrupdan: Missing docstring of inputs and outputs.

src/euroeval/metrics.py
```diff
@@ -131,8 +140,23 @@ def __init__(
         )
         self.metric: "EvaluationModule | None" = None
 
+    def download(self, cache_dir: str) -> "HuggingFaceMetric":
+        """Initiates the download of the metric if needed."""
+        # annoying but needed to make the metric download to a different cache dir
+        from datasets import DownloadConfig
```
saattrupdan: Why import here, rather than at the top?

src/euroeval/metrics.py
```diff
@@ -131,8 +140,23 @@ def __init__(
...
+
+        download_config = DownloadConfig(cache_dir=Path(cache_dir).joinpath("evaluate"))
```
saattrupdan: Nit: Simplify.

```suggestion
        download_config = DownloadConfig(cache_dir=Path(cache_dir, "evaluate"))
```

viggo-gascou:
Added tests for both generative and encoder models, currently running all tests to update coverage reports. I'll add some docs to the `README.md` once all of the tests have finished.

src/euroeval/benchmarker.py
```diff
@@ -429,24 +429,29 @@ def benchmark(
...
-                        )
-                        num_finished_benchmarks += 1
-                        continue
+                if benchmark_config.download_only:
+                    self._download(
+                        model_config=model_config,
+                        dataset_config=dataset_config,
+                        benchmark_config=benchmark_config,
+                    )
+                    continue
```
saattrupdan: [nit] Separating code blocks with different purposes makes it easier to read.

```suggestion
                    continue
                    
```

tests/test_benchmarker.py
```diff
@@ -453,3 +454,63 @@ def test_prepare_dataset_configs(
...
+
+    def test_benchmark_encoder_no_internet(
+        self,
+        benchmarker: Benchmarker,
+        task: Task,
+        language: Language,
+        encoder_model_id: str,
+    ) -> None:
+        """Test that encoder models can be benchmarked without internet."""
+        disable_socket()
```
saattrupdan: Maybe check explicitly that the internet is down when disabling. You can use the `internet_connection_available` function in `utils`.

Also, it would be great to have a test where we attempt to benchmark a model when we don't have internet available, without using `download_only`. We can test that that raises an error with `pytest.raises`.

saattrupdan (changes requested):

tests/test_benchmarker.py
viggo-gascou: When the socket is disabled, any attempt to use the internet will raise a `SocketBlockedError` and thus fail the test. So the `internet_connection_available` function will raise an error instead of returning `True/False`.

If we want a test without internet but without the `download_only` flag it would need to be run before these tests, since the `download_assets` fixture would have already downloaded the assets. The test would then not raise the `SocketBlockedError` since it would not need to download anything. Or we we would need to add some teardown or setup where we delete the cache to ensure that the assets aren't already present before the test.

src/euroeval/benchmarker.py
```diff
@@ -227,6 +232,34 @@ def benchmark_results(self) -> list[BenchmarkResult]:
         else:
             return list()
 
+    def _download(
+        self,
+        dataset_config: "DatasetConfig",
+        model_config: "ModelConfig",
+        benchmark_config: "BenchmarkConfig",
+    ) -> None:
+        """Download data, metrics, and model for the given dataset, and model."""
```
saattrupdan: Missing the "Args" section in the docstring.

src/euroeval/benchmarker.py
```diff
@@ -227,6 +232,34 @@ def benchmark_results(self) -> list[BenchmarkResult]:
...
+        log_once(f"Loading data for {dataset_config.name}", level=logging.INFO)
```
saattrupdan: [nit] Perhaps we should use pretty name here.


```suggestion
        log_once(f"Loading data for {dataset_config.pretty_name}", level=logging.INFO)
```

src/euroeval/benchmarker.py
```diff
@@ -227,6 +232,34 @@ def benchmark_results(self) -> list[BenchmarkResult]:
...
+        dataset = load_raw_data(
+            dataset_config=dataset_config, cache_dir=benchmark_config.cache_dir
+        )
+        del dataset
+        log_once(f"Loading model {model_config.model_id}", level=logging.INFO)
```
saattrupdan, line 289: [nit] Adding an empty line to separate the data loading block from the model loading block.


```suggestion

        log_once(f"Loading model {model_config.model_id}", level=logging.INFO)
```

src/euroeval/benchmarker.py
```diff
@@ -227,6 +232,34 @@ def benchmark_results(self) -> list[BenchmarkResult]:
...
+        model = load_model(
+            model_config=model_config,
+            dataset_config=dataset_config,
+            benchmark_config=benchmark_config,
+        )
+        del model
+        log_once(
```
saattrupdan, line 297: [nit] Adding an empty line to separate the model loading block from the metric loading block.


```suggestion

        log_once(
```

src/euroeval/benchmarker.py
```diff
@@ -227,6 +232,34 @@ def benchmark_results(self) -> list[BenchmarkResult]:
...
+            f"Loading metrics for the '{dataset_config.task.name}' task",
+            level=logging.INFO,
+        )
+        for metric_name in dataset_config.task.metrics:
+            log_once(f"Loading metric {metric_name.name}", level=logging.INFO)
```
saattrupdan: [nit] Perhaps we should debug log these fine-grained logs?


```suggestion
            log_once(f"Loading metric {metric_name.name}", level=logging.DEBUG)
```

src/euroeval/metrics.py
```diff
@@ -131,8 +149,29 @@ def __init__(
...
+        """Initiates the download of the metric if needed.
+
+        Args:
+            cache_dir:
+                The directory where the metric will be downloaded to.
+
+        Returns:
+            The metric object itself.
+        """
+        # annoying but needed to make the metric download to a different cache dir
```
saattrupdan: [nit] Capitalisation.


```suggestion
        # Annoying but needed to make the metric download to a different cache dir
```

tests/test_benchmarker.py
```diff
@@ -453,3 +454,63 @@ def test_prepare_dataset_configs(
 ) -> None:
     """Test that the `prepare_dataset_configs` function works as expected."""
     assert prepare_dataset_configs(dataset_names=dataset_names) == dataset_configs
+
+
+class TestNoInternetBenchmark:
+    """Tests for benchmarking models without internet."""
+
+    @pytest.fixture(autouse=True)
+    def download_assets(
```
saattrupdan: Fixtures are normally used to be able to re-use objects in tests, but this one doesn't yield any object. Perhaps we want to include the "download code" directly in tests? Then the test could download the assets, disable internet, run evaluation, re-enable internet, clear the cache?

tests/test_benchmarker.py
saattrupdan: @viggo-gascou Can't we just tweak the `internet_connection_available` function to catch `SocketBlockedError` exceptions and return False in that case?

viggo-gascou: @saattrupdan The `SocketBlockedError` comes from the `pytest-socket` library, and I think it will only be raised on the test "side", so I am not sure if we could even catch it "inside" `EuroEval`?

saattrupdan:
@viggo-gascou You've marked several comments as resolved, but the issues are still there. Did you forget to push your changes?

viggo-gascou:
@saattrupdan whoops yeah, they are comitted. Must've forgot to push them 😅

saattrupdan:
@viggo-gascou Please ping me when this is ready for review 🙂

viggo-gascou:
@saattrupdan Should be ready for review now! 😄

viggo-gascou:
Just need to add some documentation about this, but I think it can wait until the code is good to go 👍

src/euroeval/benchmark_modules/hf.py
```diff
@@ -962,6 +978,7 @@ def load_hf_model_config(
     api_key: str | None,
     trust_remote_code: bool,
     run_with_cli: bool,
+    is_offline: bool = False,
```
saattrupdan: I would not use default values here, as it can lead to silent errors, if we forget to set it somewhere.

```suggestion
    is_offline: bool,
```

src/euroeval/benchmark_modules/vllm.py
```diff
@@ -830,42 +833,44 @@ def load_model_and_tokeniser(
         model_max_length=true_max_model_len,
         model_cache_dir=model_config.model_cache_dir,
         token=get_hf_token(api_key=benchmark_config.api_key),
+        is_offline=not internet_connection_available(),
     )
 
     clear_vllm()
 
+    vllm_kwargs = {
```
saattrupdan: Downside of using a separate dict like this is that we don't get any intellisense. Also missing several arguments.

src/euroeval/benchmark_modules/vllm.py
```diff
@@ -925,6 +933,7 @@ def load_tokeniser(
         cache_dir=model_cache_dir,
         token=token,
         trust_remote_code=trust_remote_code,
+        local_files_only=is_offline,
```
saattrupdan: Instead of using the `is_offline` everywhere, can't we just call the `is_internet_available` function where we need the check?

src/euroeval/benchmarker.py
```diff
@@ -395,6 +452,28 @@ def benchmark(
                 num_finished_benchmarks += len(dataset_configs)
                 continue
 
+            if model_config.adapter_base_model_id:
+                open_issue_msg = (
+                    "If offline support is important to you, please "
+                    "consider opening an issue on GitHub."
```
saattrupdan:
```suggestion
                    "consider opening an issue at https://github.com/EuroEval/EuroEval/issues."
```

src/euroeval/utils.py
```diff
@@ -54,6 +53,65 @@ def create_model_cache_dir(cache_dir: str, model_id: str) -> str:
...
+    Returns:
+        The path to the model.
+    """
+    model_path = Path(download_dir)
+    # Get the 'path safe' version of the model id, which is the last dir in the path
+    model_id_path = model_path.name
+    # Hf hub `cache_dir` puts the files in models--`model_id_path`/snapshots
+    model_path = model_path / f"models--{model_id_path}" / "snapshots"
+    if not model_path.exists():
+        raise InvalidModel(f"No model files found at {model_path}")
```
saattrupdan:
```suggestion
        raise InvalidModel(
            f"Attempted to load models from the {model_path} directory, "
            "but it does not exist."
        )
```

src/euroeval/benchmark_modules/vllm.py
viggo-gascou: ah whoops I must've missed them when I had to resolve a merge conflict, I can add them back as "normal" kwargs and just use `model_path` if offline otherwise `model_id`?

src/euroeval/benchmark_modules/vllm.py
viggo-gascou: Since we have to negate `internet_connection_available` I felt it was nicer to have an arg like that, but I am not opposed to just calling the function instead of having the extra arg

saattrupdan:
@viggo-gascou Is this ready for a review? If so, please tag me as a reviewer, so that I get notified 🙂

viggo-gascou:
@saattrupdan yes it is, I must’ve forgotten to hit the button again 😅

PR #1171, opened 2025-08-28
Stay on Topic Task

saattrupdan:
This adds the stay-on-topic task, as well as the creation script of the stay-on-topic dataset.

Here is a description of the task:

# Stay on Topic

This task will measure whether a model is able to refuse to engage in a conversation
that it is not qualified to answer, or that it is not allowed to answer.

## Dataset Construction

1. Write down a list of 5+ different topics that it is only allowed to engage in
   conversation about, and 5+ different topics that it is not allowed to engage in
   conversation about;
2. Generate a set of 10+ questions for each topic, where 5 of them are allowed to be
   answered and 5 are not allowed to be answered;
3. Each of the 100+ questions should be checked by a legal expert to ensure that the
   questions are indeed allowed or not allowed to be answered, and that the questions
   are phrased in a way that is appropriate for the topic.

## Dataset Format

| column | type | description |
| :--- | :--- | :--- |
| system_prompt | str | The system prompt that the model should use to determine whether it is allowed to answer the question. |
| question | str | The question that should be answered. |
| answer | str | The answer to the question. |
| allowed | bool | Whether the question is allowed to be answered. |

## Evaluation Methodology

An LLM will be given the system prompt and the question, and asked to answer the
question. The model will be evaluated based on the following criteria:

1. **Answer Correctness**: The percentage of questions that the model answers correctly;
2. **Rejection Correctness**: The percentage of questions for which the model correctly
   identifies whether it is allowed to answer the question.

In practice, we compute the harmonic mean between these two metrics, to require both
high answer correctness and high allowed correctness for a good score.

To determine whether the model's output is correct, we will use LLM-as-a-judge with a
SOTA LLM to evaluate the model's output. The SOTA LLM will be given the system prompt,
the question, the ground truth answer, and the model's output, and will be asked to
determine whether the model's output is correct, in a binary fashion (yes/no).

### DIFF
--- a/.github/ISSUE_TEMPLATE/benchmark_dataset_request.yaml
+++ b/.github/ISSUE_TEMPLATE/benchmark_dataset_request.yaml
@@ -25,12 +25,14 @@ body:
       - label: Danish
       - label: Dutch
       - label: English
+      - label: Estonian
       - label: Faroese
       - label: Finnish
       - label: French
       - label: German
       - label: Icelandic
       - label: Italian
+      - label: Latvian
       - label: Norwegian (Bokmål or Nynorsk)
       - label: Portuguese
       - label: Spanish
--- a/.github/ISSUE_TEMPLATE/model_evaluation_request.yaml
+++ b/.github/ISSUE_TEMPLATE/model_evaluation_request.yaml
@@ -21,7 +21,8 @@ body:
       - label: Romance languages (French, Italian, Portuguese, Spanish)
       - label: Scandinavian languages (Danish, Faroese, Icelandic, Norwegian, Swedish)
       - label: West Germanic languages (Dutch, English, German)
-      - label: Finnish
+      - label: Finnic languages (Estonian, Finnish)
+      - label: Latvian
   validations:
     required: true
 - type: dropdown
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -58,7 +58,7 @@ jobs:
           python-version: ${{ matrix.python-version }}
 
       - name: Install Dependencies
-        run: uv sync --no-dev --extra test
+        run: uv sync --no-dev
 
       - name: Start Ollama server
         run: curl -fsSL https://ollama.com/install.sh | sh && ollama serve &
@@ -95,7 +95,7 @@ jobs:
           python-version: ${{ matrix.python-version }}
 
       - name: Install Dependencies
-        run: uv sync --no-dev --extra test
+        run: uv sync --no-dev
 
       - name: Start Ollama server
         run: curl -fsSL https://ollama.com/install.sh | sh && ollama serve &
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,22 +8,23 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 ### Added
-- Added support for Latvian 🇱🇻! This includes the Latvian part of the reading comprehension
-  dataset MultiWikiQA, the sentiment classification dataset Latvian Twitter Sentiment,
-  the linguistic acceptability dataset ScaLA-lv, the NER dataset WikiANN-lv, the
-  NER dataset FullStack-NER-lv, and the knowledge dataset MMLU-lv.
+- Added support for Latvian 🇱🇻! This includes the sentiment classification dataset
+  Latvian Twitter Sentiment, the linguistic acceptability dataset ScaLA-lv, the named
+  entity recognition datasets FullStack-NER-lv and WikiANN-lv, the reading comprehension
+  dataset MultiWikiQA, the knowledge dataset MMLU-lv, and the common-sense reasoning
+  dataset COPA-lv.
+- Added support for Estonian 🇪🇪! It includes the sentiment classification dataset
+  Estonian Valence, the linguistic acceptability datasets Grammar-et and ScaLA-et, the
+  named entity recognition dataset EstNER, the reading comprehension dataset
+  MultiWikiQA-et, the summarisation dataset ERRNews, the knowledge dataset Trivia-et,
+  and the common-sense reasoning dataset Winogrande-et. This was contributed by
+  @slowwavesleep ✨
 - It is now possible to evaluate how much a model adhere to European values! 🇪🇺 This
   probes 53 questions from the European values survey, which have been chosen based on
   an optimisation procedure that maximises agreement across the EU. We then measure how
   well the model's answers align with the distribution of answers across the EU, using a
   tree-based kernel density estimation. This can only be used zero-shot, and only with
   instruction-based decoder models (including reasoning models).
-- Added support for Estonian 🇪🇪 It currently includes the gold-standard sentiment
-  classification dataset Estonian Valence, the gold-standard named entity recognition
-  dataset EstNER, the gold-standard summarisation dataset ERRNews and the translated
-  common-sense reasoning dataset Winogrande-et which includes manually translated and
-  localised test split, as well as mahine-translated train and validation splits. This was
-  contributed by @slowwavesleep ✨
 
 ### Changed
 - When evaluating classification tasks, we now force the model to output one of the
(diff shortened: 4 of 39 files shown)


src/euroeval/tasks.py
```diff
@@ -154,3 +155,17 @@ def get_all_tasks() -> dict[str, Task]:
...
+### Danish legal benchmark tasks ###
+
+STAY_ON_TOPIC = Task(
+    name="stay-on-topic",
+    task_group=TaskGroup.TEXT_TO_TEXT,
+    template_dict=LLM_AS_A_JUDGE_TEMPLATES,
+    metrics=[m.rejection_correctness_metric],
+    default_num_few_shot_examples=1,
+    default_max_generated_tokens=128,
+    default_labels=[],
```
saattrupdan, line 169:
```suggestion
    default_labels=[],
    splits=["test"],
```

src/euroeval/tasks.py
```diff
@@ -154,3 +155,17 @@ def get_all_tasks() -> dict[str, Task]:
...
+
+
+### Danish legal benchmark tasks ###
+
+STAY_ON_TOPIC = Task(
+    name="stay-on-topic",
+    task_group=TaskGroup.TEXT_TO_TEXT,
+    template_dict=LLM_AS_A_JUDGE_TEMPLATES,
+    metrics=[m.rejection_correctness_metric],
+    default_num_few_shot_examples=1,
```
```suggestion
    default_num_few_shot_examples=0,
```

src/euroeval/dataset_configs/danish.py
```diff
@@ -175,3 +186,16 @@
     _instruction_prompt="{text}",
     unofficial=True,
 )
+
+
+# Legal datasets
+
+STAY_ON_TOPIC_CONFIG = DatasetConfig(
```
saattrupdan:
```suggestion
LEGAL_STAY_ON_TOPIC_CONFIG = DatasetConfig(
```

src/euroeval/metrics/llm_as_a_judge.py
```diff
@@ -223,6 +222,24 @@ def _apply_user_prompt(self, prediction: str, condition: str | None = None) -> s
             )
         return self.user_prompt.format(prediction=prediction)
 
+    def _get_batch_scoring_fn(
+        self,
+        scoring_fn: ScoringFunction | None,
+        batch_scoring_fn: BatchScoringFunction | None,
+    ) -> BatchScoringFunction:
```
saattrupdan, line 230: Missing docstring.

src/euroeval/metrics/llm_as_a_judge.py
```diff
@@ -255,3 +272,78 @@ class Fluency(BaseModel):
...
+    "written by a human. You need to determine whether the AI-generated text has "
+    "the same semantic content as the human-written text, in a binary fashion.\n\n"
+    "AI-generated text: {prediction!r}\n\n"
+    "Human-written text: {condition!r}\n\n"
+    "Output your answer as a JSON object with a single key 'answer', where the value "
+    "can be either 'true' (if the texts have the same semantic content) or "
+    "'false' (if they do not).",
+    response_format=BinaryOutput,
+    scoring_fn=lambda output: 1.0 if output.answer == "true" else 0.0,
+)
```
saattrupdan:
```suggestion
```

src/euroeval/tasks.py
```diff
@@ -154,3 +155,17 @@ def get_all_tasks() -> dict[str, Task]:
...
+    default_max_generated_tokens=128,
```
saattrupdan: Since we're only checking for rejections, we should be able to do with fewer tokens, speeding up evaluation as well.

```suggestion
    default_max_generated_tokens=64,
```

src/euroeval/types.py
```diff
@@ -63,6 +64,47 @@ def __call__(
...
+                context when computing the score.
+
+        Returns:
+            A float score computed from the batch of outputs.
+        """
+        ...
+
+
+
+
```
saattrupdan:
```suggestion
```

src/euroeval/types.py
```diff
@@ -104,3 +146,5 @@ def is_list_of_str(x: object) -> t.TypeGuard[list[str]]:
         Whether the object is a list of strings.
     """
     return isinstance(x, list) and all(isinstance(i, str) for i in x)
+
+
```
saattrupdan:
```suggestion
```

src/scripts/create-legal-benchmarks/create_stay_on_topic.py
```diff
@@ -0,0 +1,254 @@
...
+    "Ansættelseskonflikt": {True: [], False: []},
+    "Miljøretsligt": {True: [], False: []},
+}
+
+
+ALLOWED_TO_STRING = {
+    True: "hvor modellen har måtte svare på spørgsmålet.",
+    False: "hvor modellen IKKE har måtte svare på spørgsmålet.",
+}
+
```
saattrupdan: Should just be extracted from the sheet.

```suggestion
```

src/scripts/create-legal-benchmarks/create_stay_on_topic.py
```diff
@@ -0,0 +1,254 @@
...
+
+class SystemPrompt(BaseModel):
+    """Represents a system prompt."""
+
+    system_prompt: str
+
+
+class Question(BaseModel):
+    """Represents a single question-answer pair."""
```
saattrupdan:
```suggestion
    """Represents a single question."""
```

src/scripts/create-legal-benchmarks/create_stay_on_topic.py
```diff
@@ -0,0 +1,254 @@
...
+        pass
+
+    # Push the dataset to the Hugging Face Hub
+    logger.info(f"Uploading dataset to {dataset_id}...")
+    dataset.push_to_hub(dataset_id, private=True)
+    logger.info("Dataset uploaded successfully!")
+
+
+def read_data_from_sheet(path: str) -> pd.DataFrame:
+    """Read data from a Google Sheet."""
```
saattrupdan: Missing Args and Returns sections in docstring.

src/scripts/create-legal-benchmarks/create_stay_on_topic.py
```diff
@@ -0,0 +1,254 @@
...
+    df = df[keep_columns]
+
+    return df
+
+
+def generate_data() -> list[dict]:
+    """Generate data.
+
+    Returns:
+        list[dict]: The generated data.
```
saattrupdan: No need to put types in docstring, as the type hints have that info.


```suggestion
        The generated data.
```

src/scripts/create-legal-benchmarks/create_stay_on_topic.py
```diff
@@ -0,0 +1,254 @@
...
+        input=[{"role": "user", "content": prompt}],
+        text_format=SystemPrompt,
+        temperature=TEMPERATURE,
+    )
+    return response.output_parsed.system_prompt
+
+
+def generate_question(client: OpenAI, system_prompt: str, allowed: bool) -> str:
+    """Generate a question."""
+    prompt = (
```
saattrupdan: Consider moving this to a global `GENERATION_PROMPT` variable, where you then can use """ strings as well, making it more readable.

src/scripts/create-legal-benchmarks/create_stay_on_topic.py
```diff
@@ -0,0 +1,254 @@
...
+        model="gpt-4o",
+        input=[{"role": "user", "content": prompt}],
+        text_format=SystemPrompt,
+        temperature=TEMPERATURE,
+    )
+    return response.output_parsed.system_prompt
+
+
+def generate_question(client: OpenAI, system_prompt: str, allowed: bool) -> str:
+    """Generate a question."""
```
saattrupdan: Missing Args and Returns sections in docstring.

src/scripts/create-legal-benchmarks/create_stay_on_topic.py
```diff
@@ -0,0 +1,254 @@
...
+    #    som ikke handler om nogen af kategorierne.
+    #    Hvis allowed=False, indsæt spørgsmål fra en af kategorierne.
+    # 2. "Du må kun svare på spørgsmål om følgende kategorier:"
+    #    For 2, hvis allowed=True, indsæt spørgsmål fra en af kategorierne.
+    #    Hvis allowed=False, indsæt spørgsmål, som ikke handler om
+    #    nogen af kategorierne.
+    # Ville være nemt, hvis vi bare samplede 1 kategori.
+
+    response = client.responses.parse(
+        model="gpt-4o",
```
saattrupdan: Perhaps make the generation model a global parameter as well, like temperature?

src/scripts/create-legal-benchmarks/create_stay_on_topic.py
oliverkinch: Keeping `ALLOWED_TO_STRING`

src/euroeval/tasks.py
oliverkinch: We only need `splits=["test"]` in the dataset config, right? `splits` is not a defined key for `Task`

saattrupdan: Ah yes, my bad, thanks!

PR #1215, opened 2025-09-10
Feat/add separate separator for generation arguments

### LINKED ISSUE #1086: [FEATURE REQUEST] Add separate separator for generation arguments

Currently, we use the revision separator "@" when specifying generation arguments for LiteLLM models, such as the "thinking" argument. The problem arises when we want to do this for local models, as then the "@" separator is used for specifying the revision.

To fix this, as well as uniformise the interface for both LiteLLM and local models, we could add a new separator (say, "#"), and use that for generation arguments for all models. To ensure backwards compatibility for LiteLLM models, we should continue accepting "@" for generation arguments, but throw a deprecation warning.

We should ensure that we can handle any order of "@" and "#", as a fixed ordering is bound to create unwanted behaviour.

saattrupdan:
### Changed
- Changed the model ID syntax, where we now use `#` to indicate parameters and still use
  `@` to indicate revision. For instance, `o3#low` indicates the `o3` model with the
  low reasoning effort, and `tencent/Hunyuan-1.8B-Instruct@v1#no-thinking` indicates the
  Hunyuan model from the `v1` branch and with the `enable_thinking=False` parameter set.
  This is fully backwards compatible, in the sense that API models still support using
  `@` for parameters as well, just like previously, but you will get a warning that this
  syntax is deprecated.
- Added `thinking` and `no-thinking` parameters for all open-weight models now. Of
  course, it only makes a difference for models that supports this flag.

Closes #1086

### DIFF
--- a/src/euroeval/benchmark_modules/base.py
+++ b/src/euroeval/benchmark_modules/base.py
@@ -2,6 +2,7 @@
 
 import collections.abc as c
 import logging
+import re
 import sys
 import typing as t
 from abc import ABC, abstractmethod
@@ -55,6 +56,7 @@ class BenchmarkModule(ABC):
     fresh_model: bool
     batching_preference: "BatchingPreference"
     high_priority: bool
+    allowed_params: dict[re.Pattern, list[str]] = {re.compile(r".*"): []}
 
     def __init__(
         self,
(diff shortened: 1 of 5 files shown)


EwoutH:
> * Changed the model ID syntax, where we now use `#` to indicate parameters and still use
>   `@` to indicate revision. For instance, `o3#low` indicates the `o3` model with the
>   low reasoning effort, and `tencent/Hunyuan-1.8B-Instruct@v1#no-thinking` indicates the
>   Hunyuan model from the `v1` branch and with the `enable_thinking=False` parameter set.

This is very clear, great choice for syntax!

PR #1270, opened 2025-09-19
Fix the Titles for Estonian and Finnic Scatter Plots

slowwavesleep:
Self-explanatory. They're currently both saying "Finnish".


saattrupdan (approved):
LGTM, thanks for the fix!

PR #1335, opened 2025-10-27
Fix polish ScaLA

oliverkinch:
Introduce language specific pos tag exclusions to fix https://github.com/EuroEval/EuroEval/issues/1307.

### Code note
The `lang` arg is passed to `corruption_fn` and `lang` therefore has to be an arg to `flip_neighbours` even though that it is not used in this function. Any better way to do this more cleanly?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   the summarisation dataset LR-Sum-uk, and the knowledge dataset Global-MMLU-uk. This
   was contributed by @oliverkinch ✨
 
+### Changed
+
+- Now returns all the desired results from the `Benchmarker.benchmark` method, rather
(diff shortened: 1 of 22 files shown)


saattrupdan:
@oliverkinch  I think it might be sensible to remove deletion of particles for all languages, as worst case we don't corrupt, and best case we just corrupt something else.

Example in English, where we don't corrupt:

She ate up all her food.
She ate ~up~ all her food.

Another one:

Now, let's move on.
~Now,~ Let's move on.

oliverkinch:
> @oliverkinch I think it might be sensible to remove deletion of particles for all languages, as worst case we don't corrupt, and best case we just corrupt something else.
> 
> Example in English, where we don't corrupt:
> 
> She ate up all her food. She ate ~up~ all her food.
> 
> Another one:
> 
> Now, let's move on. ~Now,~ Let's move on.

OK, in that case this PR becomes very simple.

Issue #1306, opened 2025-10-06
[BUG] ScaLA dataset generation erases POS tags for multi-token tags


mrkowalski:
### 🐛 Describe the bug

For cases like below, where some words are parsed into multiple tokens (1-2), POS tags are erased by datased loading logic. These POS tags are then used in delete/flip_neighbours to create grammatically incorrect sentences for benchmarked LLMs to indentify:

```# sent_id = dev-s239
# text = Postanowiliśmy iść za tym pomysłem.
# orig_file_sentence = 120-2-900063_morph_3.30-s#2420
1-2    Postanowiliśmy    _    _    _    _    _    _    _    _
1    Postanowili    postanowić    VERB    praet:pl:m1:perf    Animacy=Hum|Aspect=Perf|Gender=Masc|Mood=Ind|Number=Plur|Tense=Past|VerbForm=Fin|Voice=Act    0    root    0:root    _
2    śmy    być    AUX    aglt:pl:pri:imperf:nwok    Aspect=Imp|Number=Plur|Person=1|Variant=Short    1    aux:clitic    1:aux:clitic    _
3    iść    iść    VERB    inf:imperf    Aspect=Imp|VerbForm=Inf|Voice=Act    1    xcomp    1:xcomp    _
4    za    za    ADP    prep:inst    AdpType=Prep    6    case    6:case    Case=Ins
5    tym    ten    DET    adj:sg:inst:m3:pos    Animacy=Inan|Case=Ins|Gender=Masc|Number=Sing|PronType=Dem    6    det    6:det    _
6    pomysłem    pomysł    NOUN    subst:sg:inst:m3    Animacy=Inan|Case=Ins|Gender=Masc|Number=Sing    3    obl:arg    3:obl:arg    SpaceAfter=No
7    .    .    PUNCT    interp    PunctType=Peri    1    punct    1:punct    _
```

Current code assumes that tokens without a POS tag should be taken into consideration when creating grammatically incorrect sentences. This assumption has not been verified. At least for Polish, it is likely that POS tag of the "1-2" token should be copied from the first token from the sequences (1 - "VERB").

### Operating System

Linux

### Device

CUDA GPU

### Python version

3.13.x

### EuroEval version

head

### Transformers version

4.56.1

### vLLM version

0.10.1.1

saattrupdan:
@mrkowalski Can you please see if the tagged PR solves your issue?

PR #1336, opened 2025-10-27
Fix ScaLA multi-token tags

oliverkinch:
Fix https://github.com/EuroEval/EuroEval/issues/1306

@mrkowalski can you make a review and tell me if you think this solves the issue?


mrkowalski:
I'm on it.

saattrupdan:
Merging this due to staleness - if anything doesn't seem right @mrkowalski then please let us know, so we can correct it in a separate PR 🙂

Issue #1186, opened 2025-09-02
List of models to be benchmarked on all languages


saattrupdan:
## What is a "core model"?

To prevent an ever-growing list of models to evaluate every time we change a dataset or add a new language, we'll now have a list of _core models_, which will be the minimum set of models to evaluate when such a change is made. The list of core models is very dynamic and will change relatively often.

For a model to qualify as a core model it either has to be in the top-3 within its size bracket, or sufficiently popular that people would wonder why the model is not included on the leaderboard (e.g., new popular models from the big labs). We also want to include all types of models: reasoning decoders, instruction-tuned decoders, base decoders and encoders.

### Pareto-optimal models for their size and type

- HuggingFaceTB/SmolLM2-135M
- HuggingFaceTB/SmolLM2-135M-Instruct
- HuggingFaceTB/SmolLM2-360M
- HuggingFaceTB/SmolLM2-360M-Instruct
- LiquidAI/LFM2-350M
- LiquidAI/LFM2-700M
- LiquidAI/LFM2-1.2B
- Qwen/Qwen3-0.6B
- Qwen/Qwen3-1.7B
- Qwen/Qwen3-4B
- Qwen/Qwen3-8B
- Qwen/Qwen3-14B
- Qwen/Qwen2.5-1.5B-Instruct
- Qwen/Qwen2.5-72B-Instruct
- google/gemma-3-1b-it
- google/gemma-3-1b-pt
- google/gemma-3-4b-pt
- google/gemma-3-4b-it
- google/gemma-3-12b-it
- google/gemma-3-12b-pt
- google/gemma-3-27b-it
- google/gemma-3-27b-pt
- google/gemma-3n-E4B-it
- meta-llama/Llama-3.2-1B-Instruct
- meta-llama/Llama-3.2-1B
- meta-llama/Llama-3.2-3B-Instruct
- meta-llama/Llama-3.1-70B
- meta-llama/Llama-3.3-70B-Instruct
- meta-llama/Llama-3.1-405B-Instruct-FP8
- deepcogito/cogito-v1-preview-llama-3B
- deepcogito/cogito-v1-preview-llama-8B
- deepcogito/cogito-v1-preview-qwen-14B
- deepcogito/cogito-v1-preview-qwen-32B
- deepcogito/cogito-v1-preview-llama-70B
- deepcogito/cogito-v2-preview-llama-70B
- intfloat/multilingual-e5-large
- FacebookAI/xlm-roberta-large
[...]

PR #1422, opened 2025-12-19
Add simplification task and Duidelijke Taal dataset

simonevanbruggen:
- Added Duidelijke Taal dataset (see #1421). The create_duidelijke_taal.py script filters the original dataset, keeping the sentence pairs where there is a sufficient difference in complexity and the simplifications are evaluated as reasonably accurate. 
- Added simplification as a new task and included task configs and prompts (dataset is Dutch only, added English placeholder prompt for clarity for non-Dutch speakers).
- Added METEOR and SARI metrics. SARI requires the source text as input, so also included an extension to metrics/huggingface.py with a new SourceBasedMetric.

### DIFF
--- a/tests/test_data_loading.py
+++ b/tests/test_data_loading.py
@@ -107,16 +107,19 @@ class TestAllDatasets:
         )
 
         for itr_idx in range(10):
-            few_shot_examples = (
-                extract_few_shot_examples(
-                    dataset=dataset,
-                    dataset_config=dataset_config,
-                    benchmark_config=benchmark_config,
-                    itr_idx=itr_idx,
+            if "train" in dataset_config.splits:
+                few_shot_examples = (
+                    extract_few_shot_examples(
+                        dataset=dataset,
+                        dataset_config=dataset_config,
+                        benchmark_config=benchmark_config,
+                        itr_idx=itr_idx,
+                    )
+                    if not dataset_config.task.requires_zero_shot
+                    else []
                 )
-                if not dataset_config.task.requires_zero_shot
-                else []
-            )
+            else:
+                few_shot_examples = []
             for instruction_model in [True, False]:
                 prepared_test = dataset["test"].map(
                     partial(


simonevanbruggen:
Pipeline seems to be timing out. For reference, local `make check` is passing:

```
type annotations not comments............................................Passed
fix end of files.........................................................Passed
trim trailing whitespace.................................................Passed
debug statements (python)................................................Passed
ruff (legacy alias)......................................................Passed
ruff format..............................................................Passed
Pyrefly (type checking)..................................................Passed
markdownlint-cli2........................................................Passed
```

docs/datasets/dutch.md
```diff
@@ -1028,3 +1028,77 @@ You can evaluate this dataset directly as follows:
 ```bash
 euroeval --model <model-id> --dataset wiki-lingua-nl
 ```
+
+## Simplification
+
+### Duidelijke Taal
```
saattrupdan: I would make this unofficial for now, as otherwise we need to evaluate all 100+ models for the leaderboard. If you all have the bandwidth to do that (I've got a shell script for it), then we could make it official as well 🙂 

```suggestion
### Unofficial: Duidelijke Taal
```

docs/tasks/simplification.md
```diff
@@ -0,0 +1,51 @@
+# Simplification
+
+## 📚 Overview
+
+Simplification is a task of generating a simpler, easier to understand, version of a
+given text, preserving the original
+meaning but using less complex phrases and linguistic structures.
+While the task is similar to summarization, the goal is different: simplification
```
saattrupdan: [nit] For consistency with the rest of the docs, we use UK spelling.

```suggestion
While the task is similar to summarisation, the goal is different: simplification
```

docs/tasks/simplification.md
```diff
@@ -0,0 +1,51 @@
...
+kept compared to the source.
+
+[BERTScore](https://doi.org/10.48550/arXiv.1904.09675) is also reported, which uses a
+pretrained
+encoder model (`microsoft/mdeberta-v3-base`) to encode each token in both the reference
+text and the generated
+text, and then uses cosine similarity to measure how the tokens match up. While this is
+a good measure for capturing
+meaning preservation,
+it does not provide insights into the complexity of a text.
```
saattrupdan: Is this required when we have the other metrics? Only asking since BERTScore takes a long time to compute.

src/euroeval/dataset_configs/dutch.py
```diff
@@ -150,3 +161,12 @@
     _labels=["a", "b"],
     unofficial=True,
 )
+
+DUIDELIJKE_TAAL_NL_CONFIG = DatasetConfig(
+    name="duidelijke-taal-nl",
+    pretty_name="the Dutch simplification dataset Duidelijke Taal (IvdNT)",
```
saattrupdan: The `pretty_name` field has been updated and only requires the actual dataset name now, formatted correctly:

```suggestion
    pretty_name="Duidelijke Taal",
```

src/euroeval/tasks.py
```diff
@@ -71,6 +72,16 @@
     uses_logprobs=True,
 )
 
+SIMPL = Task(
+    name="simplification",
+    task_group=TaskGroup.TEXT_TO_TEXT,
+    template_dict=SIMPL_TEMPLATES,
+    metrics=[m.meteor_metric, m.sari_metric, m.bert_score_metric],
+    default_num_few_shot_examples=3,
+    default_max_generated_tokens=128,
```
saattrupdan: Are we sure that 128 tokens is enough for all samples?

src/scripts/create_duidelijke_taal_nl.py
```diff
@@ -0,0 +1,184 @@
...
+        "Complexiteit (A) Gem.": "complexity_a_avg",
+        "Complexiteit (B)": "complexity_b",
+        "Complexiteit (B) Gem.": "complexity_b_avg",
+    }
+
+    dataset_original = load_dataset(dataset_id, token=True)
+
+    # create 50%/50% train/val split
+    # original split is 50%/50% train/test, so total split becomes 25%/25%/50%
+    train_val = dataset_original["train"].train_test_split(test_size=0.5, seed=42)
```
saattrupdan: Would it make sense here to split in a stratified fashion, to for instance have the same complexity and fluency distributions in the different splits?

src/scripts/create_duidelijke_taal_nl.py
```diff
@@ -0,0 +1,184 @@
...
+    verbosity_threshold: int, default=2
+        Threshold with the intention to drop pairs where the synthetic example contains
+        too verbose and too long explanations that deviate from the original input text.
+        Drops simplifications that contain more than threshold * number of words of the
+        input text. Setting this parameter to 1 will not filter any sentences.
+
+    Returns:
+    -------
+    pandas.DataFrame
+        The filtered dataset.
```
saattrupdan: This needs to be changed to the [Google docstring format](https://google.github.io/styleguide/pyguide.html#383-functions-and-methods), for consistency with the rest of the code base.

src/scripts/create_duidelijke_taal_nl.py
```diff
@@ -0,0 +1,184 @@
...
+        {
+            "train": train_val["train"],
+            "val": train_val["test"],
+            "test": dataset_original["test"],
+        }
+    )
+
+    for split in ["train", "val", "test"]:
+        df = dataset[split].to_pandas().rename(columns=column_mapping)
+        df_filtered = filter_dataset(df)[["text", "target_text"]]
```
saattrupdan: Maybe filtering should happen before creating the train/val split, to ensure that we don't remove too many samples from one of them here?

src/scripts/create_duidelijke_taal_nl.py
```diff
@@ -0,0 +1,184 @@
...
+    Returns:
+    -------
+    pandas.DataFrame
+        The filtered dataset.
+    """
+    bert_score = load("bertscore")
+    bert_scores = bert_score.compute(
+        references=df["text"].to_list(),
+        predictions=df["target_text"].to_list(),
+        lang="nl",
```
saattrupdan, line 140: This just loads the (quite old) mBERT (`bert-base-multilingual-cased`) as shown [here](https://github.com/Tiiiger/bert_score/blob/19e7f551fe4fa43fdd07b8129ae947015b902b2d/bert_score/utils.py#L29-L37), which isn't fantastic. This might be just fine for the filtering, but maybe we should make this more transparent in a comment here.

src/scripts/create_duidelijke_taal_nl.py
```diff
@@ -0,0 +1,184 @@
...
+            df["complexity_b"].apply(
+                lambda x: (len(x) >= min_n_participants if x is not None else False)
+            )
+        )
+        & (df["acc_avg"].apply(lambda x: x != "" and float(x) >= min_acc_avg))
+        & (df["bert"] >= min_bertscore)
+    )
+
+    if drop_explanations:
+        df_mask &= ~df["target_text"].str.contains("betekent")
```
saattrupdan, line 178:
```suggestion
    # 'betekent' = 'means' in Dutch
    if drop_explanations:
        df_mask &= ~df["target_text"].str.contains("betekent")
```

docs/datasets/dutch.md
tvosch: The leaderboard for Dutch features 432 generative models status quo. 
- Will all these models need to be evaluated and can they all run on the latest EuroEval version?
- Will the API models need to be evaluated by us too?
- A 27B model takes around 3 minutes for generation (excluding model download) on a H100 HBM2e. We are trying to make an estimation of the cost if we extrapolate this to x number of local generative models.
- It would indeed help if you have a shell script ready for this, we might just be able to run the leaderboard!

saattrupdan: Great to hear that you're willing to help out!

We don't need to evaluate all 432 models, thankfully. We've instead gone for the strategy of having a list of models that should be evaluated for all languages, which are continuously updated. This list can be found [in this pinned Github issue](https://github.com/EuroEval/EuroEval/issues/1186).

Here's a shell script that should evaluate all of those models as well as a few Dutch ones:

[run_duidelijke_taal_evaluations.sh](https://github.com/user-attachments/files/24310742/run_duidelijke_taal_evaluations.sh)

To run it, you need to have a `.env` file in your current working directory which sets `ANTHROPIC_API_KEY`, `GEMINI_API_KEY` and `OPENAI_API_KEY`. Since you're only evaluating a single dataset, this really won't cost a lost (just a few USD).

saattrupdan: Note that this script of course depends on this PR having been merged in, so we should do that first 🙂 

We could assign the dataset as unofficial in this PR, and I could then change it to official when you've got the evaluation results (or you could create a tiny PR where you change it yourself even).

tvosch: The local models sounds doable. We still need to setup multi-node vllm for the Llama 405B anyway (only 4 H100 per node) and get API keys for the three model providers anyway so indeed let's first get this PR into main :)
Thanks for the sharing the script!

saattrupdan: If the API models are an issue, then we could pay for those, since it's so little. They're also super fast, since it's just on the smaller validation splits and only zero-shot. The local models are the ones taking some time, and our servers are running on full blast already, so it'd be great if you could help out with those!

But yeah, if you could have a look at my comments below as well as fix the linting and potential tests (just run them locally, Github actions are buggy it seems), then send me a reviewer ping and I'll get right on it.

docs/tasks/simplification.md
simonevanbruggen: In our experiments we did find that BERTScore sometimes added some extra insight to the results, especially if the simplification deviated a lot semantically. Of these 3 metrics, it was the one that was the least useful in differentiating results though, so leaving it out for efficiency sounds very reasonable.

src/euroeval/tasks.py
simonevanbruggen: ~The longest target text is 43 words so should easily fit regardless of tokenizer. There are some longer texts in the original dataset, but they get filtered out.~

EDIT: just realized that there are indeed 2 examples for which input text + simplification are just over 128 (I miscalculated the total length before). So I guess that means bumping the max to 256 😄

src/scripts/create_duidelijke_taal_nl.py
simonevanbruggen: Thanks for the suggestion, that's a really good idea. I had a look at the original dataset and ended up re-doing the "pre-splitted" dataset so it now has a similar complexity distribution. The filtering is now done before the additional train/val split and the split itself is also stratified on complexity. The distributions are still slightly different (with 4 bins, the difference between splits is 0 to 3 percent points) because of the prepared split, but it's much more balanced than before. 

Complexity of the source text seemed the best metric to balance the difficulty of the task. Fluency seemed a bit more subjective and the results of balancing on this made less sense to me. 

Re-doing the filtering/splitting slightly changed the resulting distribution, so this is updated in docs/datasets/dutch.md.

simonevanbruggen:
Thanks for the feedback and suggestions @saattrupdan! 

I addressed them in my last commits, and updated the dataset splitting/filtering (see comment above). 
`make check` is passing:
```
type annotations not comments............................................Passed
fix end of files.........................................................Passed
trim trailing whitespace.................................................Passed
debug statements (python)................................................Passed
ruff (legacy alias)......................................................Passed
ruff format..............................................................Passed
Pyrefly (type checking)..................................................Passed
markdownlint-cli2........................................................Passed
```

As for `CHECK_DATASET=duidelijke-taal-nl make test`, it passes all tests except `test_data_loading.py::TestAllDatasets::test_examples_in_official_datasets_are_not_too_long[duidelijke-taal-nl]` (since it can't seem to find the validation split) but it's passing when I run this with a locally saved version of the dataset. Could you see if there are any changes needed to fix this one?

saattrupdan:
> Thanks for the feedback and suggestions @saattrupdan!
> 
> I addressed them in my last commits, and updated the dataset splitting/filtering (see comment above). `make check` is passing:
> 
> ```
> type annotations not comments............................................Passed
> fix end of files.........................................................Passed
> trim trailing whitespace.................................................Passed
> debug statements (python)................................................Passed
> ruff (legacy alias)......................................................Passed
> ruff format..............................................................Passed
> nbstripout...........................................(no files to check)Skipped
> Pyrefly (type checking)..................................................Passed
> markdownlint-cli2........................................................Passed
> ```
> 
> As for `CHECK_DATASET=duidelijke-taal-nl make test`, it passes all tests except `test_data_loading.py::TestAllDatasets::test_examples_in_official_datasets_are_not_too_long[duidelijke-taal-nl]` (since it can't seem to find the validation split) but it's passing when I run this with a locally saved version of the dataset. Could you see if there are any changes needed to fix this one?

Great! I've updated the test now and merged main in with this branch - can you try again?

simonevanbruggen:
> > Thanks for the feedback and suggestions @saattrupdan!
> > I addressed them in my last commits, and updated the dataset splitting/filtering (see comment above). `make check` is passing:
> > ```
> > type annotations not comments............................................Passed
> > fix end of files.........................................................Passed
> > trim trailing whitespace.................................................Passed
> > debug statements (python)................................................Passed
> > ruff (legacy alias)......................................................Passed
> > ruff format..............................................................Passed
> > nbstripout...........................................(no files to check)Skipped
> > Pyrefly (type checking)..................................................Passed
> > markdownlint-cli2........................................................Passed
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > As for `CHECK_DATASET=duidelijke-taal-nl make test`, it passes all tests except `test_data_loading.py::TestAllDatasets::test_examples_in_official_datasets_are_not_too_long[duidelijke-taal-nl]` (since it can't seem to find the validation split) but it's passing when I run this with a locally saved version of the dataset. Could you see if there are any changes needed to fix this one?
> 
> Great! I've updated the test now and merged main in with this branch - can you try again?

Done, all passed!  🎉 
```
=================== 212 passed, 62 skipped in 275.88s (0:04:35) ===================
```

saattrupdan (approved):
Looks great, thanks a bunch for your contribution! Looking forward to the evaluations 🙂

docs/datasets/dutch.md
saattrupdan: @tvosch Just a heads up when you're running the script: you need to change the dataset name from `duidelijke-taal-nl` to `duidelijke-taal`. Otherwise it should be alright to run now 🙂

PR #1423, opened 2025-12-22
Support IFEval

slowwavesleep:
Okay, so this required more consideration than I initially expected. For now I propose to include two datasets that work with the original code in the LM Evaluation Harness with no changes: 
- the Estonian version that is manually translated and adapted with constraints designed to work the original code
- the adapted English version where we manually replaced constraints requesting some other languages.

I moved the code from the lm harness mostly as is, which is kind of crappy I guess, and pyrefly also complains about it.

In the original dataset the constraints are held in two separate columns. I'm not sure we want to mess with that. Currently empty `target_text` is added for compatibility, which is also probably not great.

I see two ways about continuing:

1. Properly integrating the added code and fixing types etc
2. Considering the code as some kind of an appendix and using it as is

I'm not sure what's preferable here. One thing to consider is that the constraint code will have to be modified to account for changes in some languages, such as Finnish.

### DIFF
--- a/src/scripts/create_ifeval.py
+++ b/src/scripts/create_ifeval.py
@@ -5,7 +5,7 @@
 #     "huggingface-hub==0.34.4",
 # ]
 # ///
-"""Create the English IFEval instruction-following dataset and upload to HF Hub."""
+"""Create the IFEval instruction-following datasets and upload to HF Hub."""
 
 import random
 


src/euroeval/dataset_configs/estonian.py
```diff
@@ -80,6 +90,16 @@
     _instruction_prompt="{text}",
 )
 
+IFEVAL_ET_CONFIG = DatasetConfig(
+    name="ifeval-et",
+    pretty_name="IFEval-et",
+    source="EuroEval/ifeval-et",
+    task=INSTRUCTION_FOLLOWING,
+    languages=[ESTONIAN],
+    splits=["test"],
```
saattrupdan: I would highly recommend a validation split as well, as we use those when evaluating expensive proprietary models, to avoid spending too much money.

src/euroeval/dataset_configs/estonian.py
```diff
@@ -80,6 +90,16 @@
...
+    bootstrap_samples=False,
```
saattrupdan: Any reason for not bootstrapping? It gives a much better picture of what the performance is really like, and especially since this is zero-shot, there's no extra evaluation time cost at all, since all completions are cached.

src/euroeval/metrics/ifeval_constraints.py
```diff
@@ -0,0 +1,159 @@
...
+            instructions for each prompt). If False, compute instruction-level
+            accuracy (fraction of individual instructions followed across all prompts).
+    """
+
+    def __init__(
+        self,
+        name: str,
+        pretty_name: str,
+        strict: bool = True,
+        prompt_level: bool = True,
```
saattrupdan: I would remove these defaults, to make sure that no hidden errors appear later in the code - especially since these arguments are not set by users.

```suggestion
        strict: bool,
        prompt_level: bool,
```

src/euroeval/metrics/instructions.py
```diff
@@ -0,0 +1,1604 @@
+# Copyright 2023 The Google Research Authors.
```
saattrupdan: A lot of these variables and instructions are English-only. We should change them to be compatible with all languages.

Also, this code seems to be overly complex: why all the classes, why not simply have a simple single function for each criterion to be checked?

src/euroeval/metrics/instructions_registry.py
```diff
@@ -0,0 +1,193 @@
+# Copyright 2023 The Google Research Authors.
```
saattrupdan: Again, seems to be overly complex. Having all the individual functions that check each criterion within the `ifeval` metric module should be enough, no?

src/euroeval/metrics/instructions_util.py
```diff
@@ -0,0 +1,1698 @@
+# Copyright 2023 The Google Research Authors.
```
saattrupdan: Same as the other comments: overly complex it seems.

src/euroeval/tasks.py
```diff
@@ -188,3 +189,20 @@
     default_allowed_model_types=[ModelType.GENERATIVE],
     uses_logprobs=True,
 )
+
+INSTRUCTION_FOLLOWING = Task(
+    name="instruction-following",
+    task_group=TaskGroup.TEXT_TO_TEXT,
+    template_dict=INSTRUCTION_FOLLOWING_TEMPLATES,
+    metrics=[m.prompt_level_strict_acc_metric, m.inst_level_strict_acc_metric],
```
saattrupdan: Wouldn't it make more sense to have the more gradual instruction-level metric as the primary one? We only sort leaderboards by the primary one, FYI.

src/scripts/create_ifeval_en.py
```diff
@@ -0,0 +1,36 @@
...
+    ds = load_dataset(source_repo_id)
+
+    ds = ds.select_columns(["key", "prompt", "instruction_id_list", "kwargs"])
+
+    ds = ds.rename_column("prompt", "text")
+
+    # Add empty 'target_text' column for TEXT_TO_TEXT compatibility
+    # (IFEval doesn't use traditional references - the metric extracts
+    # constraints from instruction_id_list and kwargs instead)
+    ds = ds.map(lambda row: {"target_text": ""})
```
saattrupdan: We could make IFEval-style datasets more compatible with the text-to-text framework if we simply set the `target_text` to be a JSON dict with the instructions (and their kwargs) that need to be satisfied. Then we don't need to use the dataset during metric evaluation either.

src/euroeval/tasks.py
saattrupdan: Also, if we're only using the strict metrics, then all the code regarding loose metrics can be removed.

src/euroeval/dataset_configs/estonian.py
slowwavesleep: There's no validation split as such. There's about 500 examples. I suppose we can split them arbitrarily if you want to. What proportion do you have in mind?

src/euroeval/dataset_configs/estonian.py
slowwavesleep: I think I misunderstood what bootstrap is here. Is it bootstrap sampling for confidence intervals?

src/euroeval/metrics/instructions_registry.py
slowwavesleep: This is the original code, I haven't really changed it. That requires delving a bit deeper.

src/euroeval/tasks.py
slowwavesleep: I'm using strict instruction level accuracy myself, because I want to know whether the given model is able to follow the instruction exactly. I think the other types are kinda redundant, we can drop them.

src/euroeval/dataset_configs/estonian.py
saattrupdan: If there are 500 in total, then we could perhaps do a 100/400 split, stratified across the instruction categories?

src/scripts/create_ifeval_en.py
slowwavesleep: Right, there's no reason not to I suppose.

src/euroeval/dataset_configs/estonian.py
saattrupdan: Yep, indeed. Bootstrapping the test split 10 times and evaluating the model on all 10, taking the average score along with the confidence interval. So it provides an interval, but also gives a more robust estimate of the overall score.

src/euroeval/dataset_configs/estonian.py
slowwavesleep: There's in fact a relatively wide selection of categories combined in different ways. I don't see at the moment how stratification would work.

src/euroeval/dataset_configs/estonian.py
slowwavesleep: Okay, it should be set to True then.

slowwavesleep:
I think I can re-do the implementation of constraints properly, because I had plans for a dataset with additional language-specific linguistic constraints for my thesis. So I'd have to do it anyway.

src/euroeval/metrics/instructions.py
slowwavesleep: Yes, being highly English-centric is the main limitation of the original dataset. I'm not sure what they were thinking with this code here.

src/euroeval/dataset_configs/estonian.py
saattrupdan: Could be that a random sample is sufficient, but it would be great with an assertion check that both splits contains all the different instruction categories.

saattrupdan:
> I think I can re-do the implementation of constraints properly, because I had plans for a dataset with additional language-specific linguistic constraints for my thesis. So I'd have to do it anyway.

That's great, appreciate that. Again, if you need help at any point, then just shout 🙂

slowwavesleep:
@saattrupdan Let's have another look at this.

Also, I've realized right now that evaluating reasoning models might need to be handled separately. Do thinking tokens get passed to evals?

src/euroeval/dataset_configs/english.py
```diff
@@ -130,3 +141,12 @@
     _labels=["a", "b"],
     unofficial=True,
 )
+
+INSTRUCTION_FOLLOWING_CONFIG = DatasetConfig(
```
saattrupdan: We try to maintain consistency between dataset config names and the variable names.

```suggestion
IFEVAL_EN_CONFIG = DatasetConfig(
```

src/euroeval/dataset_configs/english.py
```diff
@@ -130,3 +141,12 @@
...
+    name="ifeval-en",
+    pretty_name="IFEval-en",
+    source="EuroEval/ifeval-en",
+    task=INSTRUCTION_FOLLOWING,
+    languages=[ENGLISH],
+    bootstrap_samples=True,
```
saattrupdan: [nit] This is default, so we just leave it out.

```suggestion
```

src/euroeval/dataset_configs/estonian.py
```diff
@@ -80,6 +90,15 @@
     _instruction_prompt="{text}",
 )
 
+IFEVAL_ET_CONFIG = DatasetConfig(
+    name="ifeval-et",
+    pretty_name="IFEval-et",
+    source="EuroEval/ifeval-et",
+    task=INSTRUCTION_FOLLOWING,
+    languages=[ESTONIAN],
+    bootstrap_samples=True,
```
saattrupdan: [nit] Default value.

```suggestion
```

src/euroeval/metrics/ifeval_constraints.py
```diff
@@ -0,0 +1,393 @@
...
+    from datasets.arrow_dataset import Dataset
+
+    from ..data_models import BenchmarkConfig, DatasetConfig
+
+# make langdetect deterministic
+DetectorFactory.seed = 0
+
+logger = logging.getLogger(__name__)
+
+NLTK_MIN_VERSION = "3.9.1"
```
saattrupdan: This should be part of the dependency specification in `pyproject.toml`?

src/euroeval/metrics/ifeval_constraints.py
```diff
@@ -0,0 +1,393 @@
...
+        NLTK_MIN_VERSION
+    ), (
+        f"`nltk` version {nltk_version} is not >= {NLTK_MIN_VERSION}. "
+        "Please update `nltk` before proceeding."
+    )
+    try:
+        nltk.data.find("tokenizers/punkt_tab")
+    except LookupError:
+        if _RANK == "0":
+            nltk.download("punkt_tab")
```
saattrupdan: Couldn't all this simply be the following?

```suggestion
nltk.download("punkt_tab", quiet=True)
```

src/euroeval/metrics/ifeval_constraints.py
```diff
@@ -0,0 +1,393 @@
...
+
+
+download_nltk_resources()
+
+
+@functools.lru_cache(maxsize=None)
```
saattrupdan: [nit] Simpler syntax.

```suggestion
@functools.cache
```

src/euroeval/metrics/ifeval_constraints.py
```diff
@@ -0,0 +1,393 @@
...
+def get_sentence_tokenizer() -> nltk.tokenize.PunktSentenceTokenizer:
+    """Get the cached NLTK sentence tokenizer."""
+    return nltk.data.load("nltk:tokenizers/punkt/english.pickle")
```
saattrupdan: Is this needed? When using `sent_tokenize(doc, language=language)`, it automatically loads in the sentence tokenizer?

src/euroeval/metrics/ifeval_constraints.py
```diff
@@ -0,0 +1,393 @@
...
+
+
+# all checker functions accept **kwargs to absorb extra fields from the
+# dataset's kwargs dict, which contains all possible parameters for all instructions.
+
+
+def check_keyword_existence(response: str, *, keywords: list[str], **_) -> bool:
+    """Check that all keywords exist in the response."""
```
saattrupdan: Missing "Args" and "Returns" sections of the docstring.

Same thing for the functions below.

saattrupdan: For this one it's also missing the "Raises" section.

src/euroeval/metrics/ifeval_constraints.py
```diff
@@ -0,0 +1,393 @@
...
+    if not keywords:
+        raise ValueError("keywords must be provided")
+    for keyword in keywords:
+        if not re.search(keyword, response, flags=re.IGNORECASE):
+            return False
+    return True
+
+
+def check_keyword_frequency(
+    response: str, *, keyword: str, frequency: int, relation: str, **_
```
saattrupdan: [nit] Seems like `relation` has fixed values. Would be great if the type hint could reflect that, with a `Literal["a", "b", ...]`.

src/euroeval/metrics/ifeval_constraints.py
```diff
@@ -0,0 +1,393 @@
...
+    if not forbidden_words:
+        raise ValueError("forbidden_words must be provided")
+    for word in forbidden_words:
+        if re.search(r"\b" + word + r"\b", response, flags=re.IGNORECASE):
+            return False
+    return True
+
+
+def check_letter_frequency(
+    response: str, *, letter: str, let_frequency: int, let_relation: str, **_
```
saattrupdan: [nit] Same thing as above: `let_relation` could have a `Literal[...]` type for more clear readability.

src/euroeval/metrics/ifeval_constraints.py
```diff
@@ -0,0 +1,393 @@
...
+        return counts[letter.lower()] < let_frequency
+    return counts[letter.lower()] >= let_frequency
+
+
+def check_number_sentences(
+    response: str, *, num_sentences: int, relation: str, **_
+) -> bool:
+    """Check number of sentences."""
+    tokenizer = get_sentence_tokenizer()
+    actual = len(tokenizer.tokenize(response))
```
saattrupdan: I'd use `len(sent_tokenize(response, language=language))` instead here. Also, we can include `language` as an argument here and use that in the tokenisation.

src/euroeval/metrics/ifeval_constraints.py
```diff
@@ -0,0 +1,393 @@
...
+    if relation == "less than":
+        return actual < num_sentences
+    return actual >= num_sentences
+
+
+def check_number_paragraphs(response: str, *, num_paragraphs: int, **_) -> bool:
+    """Check number of paragraphs (separated by ***)."""
+    paragraphs = re.split(r"\s?\*\*\*\s?", response)
```
saattrupdan: Why are paragraphs separated by *** rather than the usual \n\n? Is this special IFEval syntax?

src/euroeval/metrics/ifeval_constraints.py
```diff
@@ -0,0 +1,393 @@
...
+    count = len(paragraphs)
+    for i, p in enumerate(paragraphs):
+        if not p.strip():
+            if i == 0 or i == len(paragraphs) - 1:
+                count -= 1
+            else:
+                return False
```
saattrupdan, line 151: So empty paragraphs always result in a failure? Does that make sense?

src/euroeval/metrics/ifeval_constraints.py
```diff
@@ -0,0 +1,393 @@
...
+    return count == num_paragraphs
+
+
+def check_number_words(response: str, *, num_words: int, relation: str, **_) -> bool:
+    """Check number of words."""
+    tokenizer = nltk.tokenize.RegexpTokenizer(r"\w+")
+    actual = len(tokenizer.tokenize(response))
```
saattrupdan, line 169: Here I'd use `word_tokenize(doc, language=language)` instead, and again consider having `language` as an extra argument.

src/euroeval/metrics/ifeval_constraints.py
```diff
@@ -0,0 +1,393 @@
...
+
+
+def check_nth_paragraph_first_word(
+    response: str, *, num_paragraphs: int, nth_paragraph: int, first_word: str, **_
+) -> bool:
+    """Check paragraph count and first word of nth paragraph."""
+    if first_word is None:
+        raise ValueError("first_word must be provided")
+
+    paragraphs = re.split(r"\n\n", response)
```
saattrupdan, line 205: Now paragraphs are separated by "\n\n", rather than the "***" from before - why the discrepancy?

src/euroeval/metrics/ifeval_constraints.py
```diff
@@ -0,0 +1,393 @@
...
+    if relation == "less than":
+        return actual < num_words
+    return actual >= num_words
+
+
+def check_nth_paragraph_first_word(
+    response: str, *, num_paragraphs: int, nth_paragraph: int, first_word: str, **_
+) -> bool:
+    """Check paragraph count and first word of nth paragraph."""
+    if first_word is None:
```
saattrupdan, line 199: `first_word` cannot be None with the current type, so the type should be changed to `str | None` in that case.

src/euroeval/metrics/ifeval_constraints.py
```diff
@@ -0,0 +1,393 @@
...
+    if not paragraph:
+        return False
+
+    # extract first word, stripping punctuation
+    word = paragraph.split()[0].strip().lstrip("'\"")
+    actual_first = ""
+    for char in word:
+        if char in ".,?!'\"":
+            break
+        actual_first += char.lower()
```
saattrupdan, line 220: Couldn't we just use `word_tokenize` again here and take the first one?

src/euroeval/metrics/ifeval_constraints.py
```diff
@@ -0,0 +1,393 @@
...
+    actual = len(tokenizer.tokenize(response))
+    if relation == "less than":
+        return actual < num_words
+    return actual >= num_words
+
+
+def check_nth_paragraph_first_word(
+    response: str, *, num_paragraphs: int, nth_paragraph: int, first_word: str, **_
+) -> bool:
+    """Check paragraph count and first word of nth paragraph."""
```
saattrupdan: Maybe we should include a simple check that `nth_paragraph < num_paragraphs` and raise an error if not.

src/euroeval/metrics/ifeval_constraints.py
```diff
@@ -0,0 +1,393 @@
...
+            break
+        actual_first += char.lower()
+
+    return count == num_paragraphs and actual_first == first_word.lower()
+
+
+def check_number_placeholders(response: str, *, num_placeholders: int, **_) -> bool:
+    """Check minimum number of [placeholder] brackets."""
+    placeholders = re.findall(r"\[.*?\]", response)
+    return len(placeholders) >= num_placeholders
```
saattrupdan, line 237: Shouldn't `num_placeholders` be renamed to the more appropriate `min_num_placeholders` in that case?

src/euroeval/metrics/ifeval_constraints.py
```diff
@@ -0,0 +1,393 @@
...
+
+
+def check_postscript(response: str, *, postscript_marker: str, **_) -> bool:
+    """Check for postscript marker."""
+    response = response.lower()
+    if postscript_marker == "P.P.S":
+        pattern = r"\s*p\.\s?p\.\s?s.*$"
+    elif postscript_marker == "P.S.":
+        pattern = r"\s*p\.\s?s\..*$"
+    else:
```
saattrupdan, line 255: Which other postscript marker values are allowed?

src/euroeval/metrics/ifeval_constraints.py
```diff
@@ -0,0 +1,393 @@
...
+        pattern = r"\s*" + postscript_marker.lower() + r".*$"
+    return bool(re.findall(pattern, response, flags=re.MULTILINE))
+
+
+def check_number_bullet_lists(response: str, *, num_bullets: int, **_) -> bool:
+    """Check exact number of bullet points."""
```
saattrupdan: Maybe we should specify here that only "*" and "-" can start bullet points here.

src/euroeval/metrics/ifeval_constraints.py
```diff
@@ -0,0 +1,393 @@
...
+    bullets1 = re.findall(r"^\s*\*[^\*].*$", response, flags=re.MULTILINE)
+    bullets2 = re.findall(r"^\s*-.*$", response, flags=re.MULTILINE)
+    return len(bullets1) + len(bullets2) == num_bullets
+
+
+def check_constrained_response(response: str, **_) -> bool:
+    """Check response contains one of the constrained options."""
+    options = ("My answer is yes.", "My answer is no.", "My answer is maybe.")
+    return any(opt in response.strip() for opt in options)
```
saattrupdan, line 287: Shouldn't this be more general, with the `options` just being an argument? It seems like all it is is a disjunctive version of `check_keyword_existence`?

src/euroeval/metrics/ifeval_constraints.py
```diff
@@ -0,0 +1,393 @@
...
+
+
+def check_number_highlighted_sections(
+    response: str, *, num_highlights: int, **_
```
saattrupdan, line 291: Shouldn't this be `min_num_highlights`, since that's what's being checked?

src/euroeval/metrics/ifeval_constraints.py
```diff
@@ -0,0 +1,393 @@
...
+) -> bool:
+    """Check minimum highlighted *sections*."""
+    count = 0
+    for h in re.findall(r"\*[^\n\*]*\*", response):
+        if h.strip("*").strip():
+            count += 1
+    for h in re.findall(r"\*\*[^\n\*]*\*\*", response):
+        if h.removeprefix("**").removesuffix("**").strip():
```
saattrupdan, line 309: Why not just `h.strip("*")` like above?

src/euroeval/metrics/ifeval_constraints.py
```diff
@@ -0,0 +1,393 @@
...
+            count += 1
+    return count >= num_highlights
+
+
+def check_multiple_sections(
+    response: str, *, section_spliter: str, num_sections: int, **_
```
saattrupdan, line 315: Should be `min_num_sections` here.

src/euroeval/metrics/ifeval_constraints.py
```diff
@@ -0,0 +1,393 @@
...
+    """Check response is valid JSON."""
+    value = (
+        response.strip()
+        .removeprefix("```json")
+        .removeprefix("```Json")
+        .removeprefix("```JSON")
+        .removeprefix("```")
+        .removesuffix("```")
+        .strip()
+    )
```
saattrupdan, line 352: Maybe just `value = re.sub(r"(^```(json)?|```$", "", flags=re.IGNORECASE).strip()`?

saattrupdan: Also, separately, why are we allowing these ``` tags, if the model is supposed to output the JSON dict directly? Wouldn't that count as a failure?

src/euroeval/metrics/ifeval_constraints.py
```diff
@@ -0,0 +1,393 @@
...
+    try:
+        json.loads(value)
+        return True
+    except ValueError:
+        return False
+
+
+def check_title(response: str, **_) -> bool:
+    """Check for <<title>> format."""
+    for title in re.findall(r"<<[^\n]+>>", response):
```
saattrupdan, line 370: Don't we want to exclude the possibility of nested tags?

```suggestion
    for title in re.findall(r"<<[^\n<>]+>>", response):
```

src/euroeval/metrics/ifeval_constraints.py
```diff
@@ -0,0 +1,393 @@
...
+        if title.lstrip("<").rstrip(">").strip():
+            return True
+    return False
+
+
+def check_two_responses(response: str, **_) -> bool:
+    """Check for two different responses separated by ******."""
+    parts = response.split("******")
```
saattrupdan: Why this exact splitter? Seems like this could be more general and merged with some of the other functions?

src/euroeval/metrics/ifeval_constraints.py
```diff
@@ -0,0 +1,393 @@
...
+        else:
+            valid.append(part)
+    return len(valid) == 2 and valid[0].strip() != valid[1].strip()
+
+
+def check_repeat_prompt(response: str, *, prompt_to_repeat: str, **_) -> bool:
+    """Check response starts with the prompt."""
+    if not prompt_to_repeat:
+        raise ValueError("prompt_to_repeat must be provided")
+    return response.strip().lower().startswith(prompt_to_repeat.strip().lower())
```
saattrupdan, line 413: Aren't we checking that a response starts with a prefix, rather than repeating anything here?

src/euroeval/metrics/ifeval_constraints.py
```diff
@@ -0,0 +1,393 @@
...
+
+
+def check_end_phrase(response: str, *, end_phrase: str, **_) -> bool:
+    """Check response ends with exact phrase."""
+    return response.strip().strip('"').lower().endswith(end_phrase.strip().lower())
```
saattrupdan, line 426: Why are we allowing the model to wrap the response in quotes here?

src/euroeval/metrics/ifeval_constraints.py
```diff
@@ -0,0 +1,393 @@
...
+
+
+def check_capital_word_frequency(
+    response: str, *, capital_frequency: int, capital_relation: str, **_
```
saattrupdan: Shouldn't `capital_frequency` be renamed to something like `capital_frequency_threshold`?

src/euroeval/metrics/ifeval_constraints.py
```diff
@@ -0,0 +1,393 @@
...
+    if capital_relation == "less than":
+        return count < capital_frequency
+    return count >= capital_frequency
+
+
+def check_english_capital(response: str, **_) -> bool:
+    """Check response is English and all caps."""
+    try:
+        return response.isupper() and langdetect.detect(response) == "en"
+    except langdetect.LangDetectException:
```
saattrupdan, line 462: When does this happen?

saattrupdan: Also, wouldn't it make more sense to have more modular functions, one checking for all caps and one checking for language? Then we can swap the languages as needed.

src/euroeval/metrics/ifeval_constraints.py
```diff
@@ -0,0 +1,393 @@
...
+    """Check response is English and all lowercase."""
+    try:
+        return response.islower() and langdetect.detect(response) == "en"
+    except langdetect.LangDetectException:
+        return True
+
+
+def check_no_comma(response: str, **_) -> bool:
+    """Check response has no commas."""
+    return "," not in response
```
saattrupdan, line 491: Couldn't this be replaced with one checking for forbidden characters in general, like the forbidden words? Maybe even merging them, as the only difference is the boundary tokens.

src/euroeval/metrics/ifeval_constraints.py
```diff
@@ -0,0 +1,393 @@
...
+
+
+def check_quotation(response: str, **_) -> bool:
+    """Check response is wrapped in double quotes."""
+    response = response.strip()
+    return len(response) > 1 and response[0] == '"' and response[-1] == '"'
+
+
+instruction_checkers = {
```
saattrupdan, line 508: [nit] Global constants are normally in all caps:

```suggestion
INSTRUCTION_CHECKERS = {
```

src/euroeval/metrics/ifeval_constraints.py
```diff
@@ -0,0 +1,393 @@
...
+    "combination:repeat_prompt": check_repeat_prompt,
+    "startend:end_checker": check_end_phrase,
+    "change_case:capital_word_frequency": check_capital_word_frequency,
+    "change_case:english_capital": check_english_capital,
+    "change_case:english_lowercase": check_english_lowercase,
+    "punctuation:no_comma": check_no_comma,
+    "startend:quotation": check_quotation,
+}
+
+skipped_instructions = {"language:response_language"}
```
saattrupdan: [nit] Global constants are normally in all caps:

```suggestion
SKIPPED_INSTRUCTIONS = {"language:response_language"}
```

src/euroeval/metrics/ifeval_constraints.py
```diff
@@ -0,0 +1,393 @@
...
+    def __init__(
+        self,
+        name: str = "inst_level_strict_acc",
+        pretty_name: str = "Instruction-Level Strict Accuracy",
+        postprocessing_fn: t.Callable[[float], tuple[float, str]] | None = None,
+    ) -> None:
+        """Initialize the metric."""
+        super().__init__(
+            name=name, pretty_name=pretty_name, postprocessing_fn=postprocessing_fn
+        )
```
saattrupdan, line 585: We could leave this block out and simply write the name and pretty name when initialising the metric.

src/euroeval/dataset_configs/english.py
```diff
@@ -130,3 +141,12 @@
...
+)
```
saattrupdan, line 152: Missing specification that no validation splits are present:

```suggestion
    splits=["train", "test"],
)
```

src/euroeval/dataset_configs/estonian.py
```diff
@@ -80,6 +90,15 @@
...
+)
```
saattrupdan, line 101: Missing specification that no validation splits are present:

```suggestion
    splits=["train", "test"],
)
```

src/scripts/create_ifeval.py
```diff
@@ -0,0 +1,121 @@
+# /// script
```
saattrupdan, line 1: Can't we just have a single creation script, as it seems that it's just the source dataset IDs that changes?

src/euroeval/metrics/ifeval_constraints.py
```diff
@@ -0,0 +1,393 @@
...
+    "combination:two_responses": check_two_responses,
+    "combination:repeat_prompt": check_repeat_prompt,
+    "startend:end_checker": check_end_phrase,
+    "change_case:capital_word_frequency": check_capital_word_frequency,
+    "change_case:english_capital": check_english_capital,
+    "change_case:english_lowercase": check_english_lowercase,
+    "punctuation:no_comma": check_no_comma,
+    "startend:quotation": check_quotation,
+}
+
```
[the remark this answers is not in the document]
slowwavesleep: We skip response language identification because we want to only measure instruction-following capabilities in the target language and not the ability to generate some unrelated languages.

src/scripts/create_ifeval.py
slowwavesleep: @saattrupdan that would makes sense, but how do you pass the source dataset to the script, what's the convention?

src/scripts/create_ifeval.py
```diff
@@ -0,0 +1,121 @@
...
+
+import random
+
+from datasets import Dataset, DatasetDict, load_dataset
+from huggingface_hub import HfApi
+
+
+def stratified_split(
+    dataset: Dataset, label_key: str, test_size: float, seed: int
+) -> DatasetDict:
```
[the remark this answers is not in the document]
slowwavesleep: Will just make a single script instead.

src/scripts/create_ifeval_et.py
```diff
@@ -0,0 +1,121 @@
...
+    )
+
+    def transform(row: dict) -> dict:
+        return {
+            "text": row["prompt"],
+            "target_text": {
+                "instruction_id_list": row["instruction_id_list"],
+                "kwargs": row["kwargs"],
+            },
+        }
```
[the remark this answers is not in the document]
slowwavesleep: We'll just make a single script.

src/euroeval/metrics/ifeval_constraints.py
slowwavesleep: Yes, it should

src/euroeval/metrics/ifeval_constraints.py
slowwavesleep: Probably not, I'll check.

src/euroeval/metrics/ifeval_constraints.py
```diff
@@ -0,0 +1,393 @@
...
+download_nltk_resources()
+
+
+@functools.lru_cache(maxsize=None)
+def get_sentence_tokenizer() -> nltk.tokenize.PunktSentenceTokenizer:
+    """Get the cached NLTK sentence tokenizer."""
+    return nltk.data.load("nltk:tokenizers/punkt/english.pickle")
+
+
```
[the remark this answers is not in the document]
slowwavesleep: The approach to tokenization needs to be reworked.

src/euroeval/metrics/ifeval_constraints.py
slowwavesleep: The values are language-specific, for example it can be just "PS". I'm guessing here it's just an awkward way to escape the periods in the regex patterns.

src/euroeval/metrics/ifeval_constraints.py
```diff
@@ -0,0 +1,378 @@
...
+# make langdetect deterministic
+DetectorFactory.seed = 0
+
+logger = logging.getLogger(__name__)
+
+nltk.download("punkt_tab", quiet=True)
+
+
+download_nltk_resources()
```
saattrupdan: Shouldn't the `nltk.download` be included in the `download_nltk_resources`, or am I misunderstanding it?

src/scripts/create_ifeval.py
saattrupdan: You can simply have a hardcoded mapping of languages to dataset IDs, assuming their processing is identical, or at least very close to identical. You can have a look at `create_hellaswag` or `create_mmlu` for some examples.

saattrupdan (commented):
Just a few comments. As for the unresolved comments above, do you need a hand with fixing those bits, in case you don't have a lot of bandwidth these days? 🙂

src/euroeval/metrics/ifeval_constraints.py
slowwavesleep: I think they just arbitrarily decided that paragraphs are separated like that. It's "hardcoded" like this in the prompts.

src/euroeval/metrics/ifeval_constraints.py
slowwavesleep: Let's leave it for a separate iteration.

src/euroeval/metrics/ifeval_constraints.py
slowwavesleep: I think it does. We're expecting the text formatted in a specific way. It shouldn't have empty paragraphs.

src/euroeval/metrics/ifeval_constraints.py
slowwavesleep: I suppose in this case these are regular paragraphs. The prompts for this constraint don't mention the "****". So it's the data itself that's inconsistent.

src/euroeval/metrics/ifeval_constraints.py
slowwavesleep: I suppose we could, but let's keep it for backwards compatibility for now. I don't know what they had in mind with this.

src/euroeval/metrics/ifeval_constraints.py
slowwavesleep: I wouldn't mess with renaming because that argument names come from data, so renaming would mean working around that. The "at least" relation is basically "hardcoded" in the prompts.

src/euroeval/metrics/ifeval_constraints.py
slowwavesleep: It should be, but that's how it is. We don't use it in our localized version, for example. Meanwhile, in the Finnish version I think they add a separate version with hardcoded Finnish translations.

slowwavesleep: The options in this case don't come as arguments but rather embedded in the prompt:

> If Bob beat Martha in a game of pool. And Martha beat Joe in a game of pool. Can Bob beat Joe in a game of pool?  Your answer must contain exactly one of the following phrases: My answer is yes. My answer is no. My answer is maybe.

src/euroeval/metrics/ifeval_constraints.py
slowwavesleep: It should be, but the arguments are have these names in the dataset itself.

src/euroeval/metrics/ifeval_constraints.py
slowwavesleep: Probably no good reason, but let's keep it for now.

src/euroeval/metrics/ifeval_constraints.py
slowwavesleep: Similarly, this is read from the data.

src/euroeval/metrics/ifeval_constraints.py
slowwavesleep: That's the original implementation. I think we can change it later to something that makes more sense, but for now I would aim for reproducing the scores.

src/euroeval/metrics/ifeval_constraints.py
slowwavesleep: Again, it's probably an error, but let's make it comparable to the original for now.

src/euroeval/metrics/ifeval_constraints.py
slowwavesleep: It could be, but it can't be changed without modifying the dataset.

src/euroeval/metrics/ifeval_constraints.py
slowwavesleep: Technically yes, but that's what the prompts say:

> In this task, repeat the exact request below first, then give your response. Do not say any word before repeating the exact request.  Write an acoustic song about the Korean peninsula without using any commas.

So they mean repeating the specific phrase in the prompt. In this particular example the argument is like this: prompt_to_repeat = "Write an acoustic song about the Korean peninsula without using any commas."

src/euroeval/metrics/ifeval_constraints.py
slowwavesleep: That seems arbitrary. Let's leave it for reproducibility for now.

src/euroeval/metrics/ifeval_constraints.py
slowwavesleep: We can't easily rename arguments.

src/euroeval/metrics/ifeval_constraints.py
slowwavesleep: I don't think this exception made any sense. As for modular functions, see the problem with renaming the args. It can be done, but it's pretty significant additional effort that may result in notably different scores.

src/euroeval/metrics/ifeval_constraints.py
slowwavesleep: That was some kind mental bug on my side.

slowwavesleep:
@saattrupdan To summarize, the problem with the original IFEval is that it essentially tightly couples the evaluation code with the data. The arguments and their values for the functions are specified in the data. The constraints themselves are also a mess because I think it was partially generated with some models or something. My suggestion for now is to keep it relatively close to the original for the first iteration. Some of the issues with the code may be fixed, while the others require adjusting the dataset in a substantial way and may be not worth the effort.

saattrupdan:
> @saattrupdan To summarize, the problem with the original IFEval is that it essentially tightly couples the evaluation code with the data. The arguments and their values for the functions are specified in the data. The constraints themselves are also a mess because I think it was partially generated with some models or something. My suggestion for now is to keep it relatively close to the original for the first iteration. Some of the issues with the code may be fixed, while the others require adjusting the dataset in a substantial way and may be not worth the effort.

Fair enough. Let's not change too much for now, and leave that for the future, as there might be other things that we'd like to check for as well.

saattrupdan:
@slowwavesleep Can you pull the main branch in and resolve the conflicts? Then we should be able to merge with this initial version 🙂

slowwavesleep:
@saattrupdan It's done.

saattrupdan (approved):
LGTM!

PR #1428, opened 2025-12-24
Support Ray as distributed backend executor for vLLM inference engine

tvosch:
### Description:
- Added `select_backend_and_parallelism()` utility to automatically choose between multi-processing (mp) and Ray distributed execution based on hardware availability.
- Configured tensor_parallel_size and pipeline_parallel_size programmatically using available GPUs and nodes.
- Updated vllm.py to initialize Ray if available and fallback to mp on single-node or Ray-unavailable scenarios.
- Single-node behaviour remains unchanged
- [Here](https://gist.github.com/tvosch/7bccfdbb0c2cddf188321101d6f3c4ff) is a SLURM test script for run EuroEval multi-node with SLURM 

### Testing:
- @saattrupdan adding a mocked test for `select_backend_and_parallelism()` will require ray to be in the dependencies for the `make check`. Is this desired?
- Alternatively, should a real test be included or an example of using multi-node with Ray on EuroEval? I can provide a small batch script

### Notes:
- Ray workers are assumed to be in sync in this script by adding `--block` to spawning the Ray server on worker node.
- No custom `tensor-parallel-size` or `pipeline-parallel-size` can still be passed.

### DIFF
--- a/src/euroeval/cli.py
+++ b/src/euroeval/cli.py
@@ -38,39 +38,12 @@ from .languages import get_all_languages
     models will be benchmarked on all datasets.""",
 )
 @click.option(
-    "--model-language",
-    "-ml",
-    default=None,
-    show_default=True,
-    multiple=True,
-    metavar="ISO 639-1 LANGUAGE CODE",
-    type=click.Choice(["all"] + list(get_all_languages().keys())),
-    help="""This option is deprecated - please use --language instead.""",
-)
-@click.option(
-    "--dataset-language",
-    "-dl",
-    default=None,
-    show_default=True,
-    multiple=True,
-    metavar="ISO 639-1 LANGUAGE CODE",
-    type=click.Choice(["all"] + list(get_all_languages().keys())),
-    help="""This option is deprecated - please use --language instead.""",
-)
-@click.option(
     "--dataset",
     default=None,
     show_default=True,
     multiple=True,
     help="""The name of the benchmark dataset. We recommend to use the `task` and
     `language` options instead of this option.""",
-)
-@click.option(
-    "--batch-size",
-    default=None,
-    type=click.Choice(["1", "2", "4", "8", "16", "32"]),
-    help="This option is deprecated - please use --finetuning-batch-size instead.",
-    deprecated=True,
 )
 @click.option(
     "--finetuning-batch-size",
@@ -198,14 +171,6 @@ from .languages import get_all_languages
     "if you are running out of GPU memory. Only relevant if the model is generative.",
 )
 @click.option(
-    "--debug/--no-debug",
-    default=False,
-    show_default=True,
-    help="Whether to run the benchmark in debug mode. This prints out extra "
-    "information and stores all outputs to the current working directory. Only "
-    "relevant if the model is generative.",
-)
-@click.option(
     "--requires-safetensors",
     is_flag=True,
     help="Only allow loading models that have safetensors weights available",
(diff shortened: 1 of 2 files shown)


src/euroeval/benchmark_modules/vllm.py
```diff
@@ -12,6 +12,7 @@
 from pathlib import Path
 from time import sleep
 
+import ray
```
saattrupdan: Can you wrap this in a `if t.TYPE_CHECKING or importlib.util.find_spec("ray") is not None` block, together with the corresponding vllm block below?

src/euroeval/benchmark_modules/vllm.py
```diff
@@ -1430,3 +1436,42 @@ def get_vllm_tokenisation_params(
...
         load_format=load_format,
     )
+
+
+def select_backend_and_parallelism() -> tuple[str, int, int]:
+    """Determine the distributed backend and parallelism for vLLM.
+
+    Args:
+        None
+
```
saattrupdan: [nit] If there aren't any arguments then no "Args" section is required:

```suggestion
```

src/euroeval/benchmark_modules/vllm.py
```diff
@@ -1430,3 +1436,42 @@ def get_vllm_tokenisation_params(
         config_format=config_format,
         load_format=load_format,
     )
+
+
+def select_backend_and_parallelism() -> tuple[str, int, int]:
```
saattrupdan, line 1444: [nit] We could consider caching this function (using the `functools.cache` decorator), to avoid unnecessary repeated checks.

src/euroeval/benchmark_modules/vllm.py
```diff
@@ -1430,3 +1436,42 @@ def get_vllm_tokenisation_params(
...
+    Returns:
+        Tuple containing:
+        - backend (str): "ray" if multi-node Ray is available, else "mp".
+        - tensor_parallel_size (int): Number of GPUs per node.
+        - pipeline_parallel_size (int): Number of stages across nodes.
+    """
+    if not ray.is_initialized():
+        try:
+            ray.init(address="auto", ignore_reinit_error=True)
+        except Exception:
```
saattrupdan: Adding debug logging to enable debugging of future ray issues:

```suggestion
        except Exception as e:
            log_once(
                f"Ray initialisation failed with a {type(e)} exception: {e}",
                level=logging.DEBUG,
            )
```

src/euroeval/benchmark_modules/vllm.py
```diff
@@ -1430,3 +1436,42 @@ def get_vllm_tokenisation_params(
...
+    is_ray = ray.is_initialized()
+    local_gpu_count = torch.cuda.device_count()
+
+    if is_ray:
+        resources = ray.cluster_resources()
+        total_gpus = int(resources.get("GPU", 0))
+    else:
+        total_gpus = local_gpu_count
+
+    if is_ray and total_gpus > local_gpu_count:
```
saattrupdan: [nit] Just adding an extra variable to make it clear what's happening here.

```suggestion
    using_multiple_nodes = total_gpus > local_gpu_count
    if is_ray and using_multiple_nodes:
```

src/euroeval/benchmark_modules/vllm.py
```diff
@@ -1430,3 +1436,42 @@ def get_vllm_tokenisation_params(
...
+        distributed_executor_backend = "ray"
+        tensor_parallel_size = local_gpu_count if local_gpu_count > 0 else 1
+        pipeline_parallel_size = max(1, total_gpus // tensor_parallel_size)
```
saattrupdan, line 1475:
```suggestion
        pipeline_parallel_size = max(1, total_gpus // tensor_parallel_size)
        log_once(
            f"Detected a multi-node setup with {pipeline_parallel_size:,} nodes, each "
            f"with {tensor_parallel_size:,} GPUs, so using `ray` as the distributed backend. ",
            level=logging.DEBUG,
        )
```

src/euroeval/benchmark_modules/vllm.py
```diff
@@ -1430,3 +1436,42 @@ def get_vllm_tokenisation_params(
...
+    else:
+        distributed_executor_backend = "mp"
+        tensor_parallel_size = local_gpu_count if local_gpu_count > 0 else 1
+        pipeline_parallel_size = 1
```
saattrupdan, line 1485:
```suggestion
        pipeline_parallel_size = 1
        log_once(
            f"Detected a single-node setup with {tensor_parallel_size:,} GPUs, so using "
            "the multiprocessing distributed backend."
            level=logging.DEBUG,
        )
```

saattrupdan (changes requested):
Looks great, just some documentation and logging additions really.

Also, you can add `ray` to the dependencies as part of the `generative` group, just like vLLM. I.e., add the following line to both the `generative` and `all` sections of `pyproject.toml`:

```toml
    "ray>=2.53.0; platform_system == 'Linux'",
```

src/euroeval/benchmark_modules/vllm.py
tvosch: Intuitively, I would refrain from caching any functions or calls that set up or maintain an entrypoint (like a Ray cluster), especially over multiple nodes. My assumption here is that a local compute cluster may not always be stable such that every GPU in every node continuously is healthy during benchmarking. Any subsequent GPU/node changes will not be picked up when cached.

tvosch:
Updated the PR overview with a SLURM test script for multi-node vLLM-based EuroEval benchmarking in this GitHub Gist [link](https://gist.github.com/tvosch/7bccfdbb0c2cddf188321101d6f3c4ff)

tvosch:
> Looks great, just some documentation and logging additions really.
> 
> Also, you can add `ray` to the dependencies as part of the `generative` group, just like vLLM. I.e., add the following line to both the `generative` and `all` sections of `pyproject.toml`:
> 
> ```toml
>     "ray>=2.53.0; platform_system == 'Linux'",
> ```

I have added this explicitly in the `pyproject.toml`, even though `vllm=0.11.0` has Ray as dependency already

src/euroeval/benchmark_modules/vllm.py
saattrupdan: Ah, instability makes sense in this case yeah. Fair enough, we keep it as-is!

saattrupdan (approved):

saattrupdan:
@tvosch Maybe you could make a separate PR with a section of the readme with multi-node running, including a link to your Slurm gist?

PR #1445, opened 2026-01-08
First draft machine translation benchmark

oliverkinch:
This PR introduces a machine translation task, which is currently pretty much a copy of the summarisation task. One slightly ugly thing is that a machine translation dataset is defined for a language pair and not a single language like all other datasets in EuroEval. I will happily take suggestion how we should incorporate this.

I have added one example dataset configuration for a [Danish to Ukranian MT dataset](https://huggingface.co/datasets/oliverkinch/mt_da_uk). 

We can evaluate a model on this dataset by running the following command:

```bash
euroeval --model gpt-4o-mini --dataset mt-da-uk
```

In principle we can reuse the same dataset, but with the current framework the simplest approach is to duplicate the dataset and swap/rename the source and target columns so that Danish becomes the target language, and publish it as e.g. mt_uk_da. That would then get its own dataset config and Ukrainian prompt templates.


src/euroeval/data_models.py
```diff
@@ -113,7 +114,7 @@ class Task:
 
     name: str
     task_group: TaskGroup
-    template_dict: dict[Language, PromptConfig]
+    template_dict: dict[Language, PromptConfig] | dict[LanguagePair, PromptConfig]
```
saattrupdan: Instead of having a new LanguagePair class, couldn't we just have a `tuple[Language, Language]`? Then this type would just be `dict[Language | tuple[Language, Language], PromptConfig]`.

src/euroeval/data_models.py
```diff
@@ -160,6 +161,9 @@ class DatasetConfig:
             The task of the dataset.
         languages:
             The ISO 639-1 language codes of the entries in the dataset.
+        language_pair (optional):
+            The language pair of the dataset. Should only be set for machine
+            translation tasks.
```
saattrupdan: Couldn't we just add the languages in the language pair to `languages`?

src/euroeval/data_models.py
```diff
@@ -241,12 +246,15 @@ class DatasetConfig:
...
-    def main_language(self) -> Language:
+    def main_language(self) -> Language | LanguagePair:
         """Get the main language of the dataset.
 
         Returns:
             The main language.
         """
+        if self.language_pair is not None:
+            return self.language_pair
+
```
saattrupdan, line 260: Maybe here we could just check if `self.task` is machine translation, in which case we grab the first two languages in `languages`?

There might be other multilingual benchmarks, so we could add a constant in `constants.py` which records the tasks for which multiple languages should be used, perhaps?

src/euroeval/data_models.py
```diff
@@ -323,7 +331,7 @@ def prompt_prefix(self) -> str:
     @property
     def prompt_template(self) -> str:
         """The template used during few-shot evaluation."""
-        prompt_config = self.task.template_dict[self.main_language]
```
saattrupdan, line 328: Couldn't we just have `template_dict` allowing a pair of languages as key?

src/euroeval/tasks.py
```diff
@@ -199,3 +200,14 @@
     default_allowed_model_types=[ModelType.GENERATIVE],
     uses_logprobs=True,
 )
+
+MT = Task(
+    name="machine-translation",
+    task_group=TaskGroup.TEXT_TO_TEXT,
+    template_dict=MT_TEMPLATES,
+    metrics=([m.bert_score_metric]),
```
```suggestion
    metrics=[m.bert_score_metric],
```

src/euroeval/tasks.py
```diff
@@ -199,3 +200,14 @@
...
+    default_num_few_shot_examples=1,
```
saattrupdan: We might want more than 1 few-shot example. It's only set to 1 for summarisation tasks as the docs are so long.

src/euroeval/dataset_configs/mt_dataset_configs.py
```diff
@@ -0,0 +1,15 @@
+"""Machine translation dataset configurations."""
```
saattrupdan, line 1: Don't see any better idea than just having the dataset configs together like this. It's going to get really messy really soon though, so we should probably think about another option.

src/euroeval/dataset_configs/mt_dataset_configs.py
```diff
@@ -0,0 +1,15 @@
...
+
+from ..data_models import DatasetConfig
+from ..language_pairs_mt import DANISH_TO_UKRAINIAN
+from ..languages import DANISH, UKRAINIAN
+from ..tasks import MT
+
+MT_DA_UK_CONFIG = DatasetConfig(
+    name="mt-da-uk",
+    pretty_name="MT-DA-UK",
```
saattrupdan, line 9: A little bit too generic name - can't we replace this with the sources of the data?

oliverkinch: This is just a draft dataset I used for testing. I will add a new dataset later with a less generic name (and hence delete this one).

src/euroeval/data_models.py
```diff
@@ -113,7 +113,9 @@ class Task:
 
     name: str
     task_group: TaskGroup
-    template_dict: dict[Language, PromptConfig]
+    template_dict: (
+        dict[Language, PromptConfig] | dict[tuple[Language, Language], PromptConfig]
+    )
```
oliverkinch: I use the typing 

```python
dict[Language, PromptConfig] | dict[tuple[Language, Language], PromptConfig]
```

instead of 

```python
dict[Language | dict[tuple[Language, Language], PromptConfig]
````

simply to please pyright. Not exactly sure why the second produces errors.

Comment (saattrupdan) [reply src/euroeval/data_models.py]: You've got nested dicts in both of them - can't you set it to `dict[Language | tuple[Language, Language], PromptConfig]`?

File: src/euroeval/dataset_configs/mt_dataset_configs.py (hunk 8 above)
Comment (saattrupdan) [reply src/euroeval/dataset_configs/mt_dataset_configs.py:9]: Ah cool cool

File: src/euroeval/data_models.py (hunk 1 above)
Comment (oliverkinch) [reply src/euroeval/data_models.py]: Now simply using `tuple[Language, Language]`

File: src/euroeval/data_models.py (hunk 2 above)
Comment (oliverkinch) [reply src/euroeval/data_models.py]: I have removed the `language_pair` argument

File: src/euroeval/data_models.py (hunk 3 above)
Comment (oliverkinch) [reply src/euroeval/data_models.py:260]: Now checking if `self.task` is machine translation and in that case returning a tuple with languages.

File: src/euroeval/tasks.py (hunk 6 above)
Comment (oliverkinch) [reply src/euroeval/tasks.py]: I have changed it to 5.

File: src/euroeval/data_models.py (hunk 9 above)
Comment (oliverkinch) [reply src/euroeval/data_models.py]: I tried again and I am now not getting any type errors? Pretty sure this was what I had already tried.. it works now :)

File: src/euroeval/prompt_templates/machine_translation.py [hunk 10]
```diff
@@ -0,0 +1,20 @@
+"""Templates for the Summarization task."""
+
+from ..data_models import PromptConfig
+from ..language_pairs_mt import DANISH_TO_UKRAINIAN
+from ..languages import Language
+
+MT_TEMPLATES: dict[tuple[Language, Language], PromptConfig] = {
+    DANISH_TO_UKRAINIAN: PromptConfig(
```
saattrupdan: Isn't this a bit more clean?

```suggestion
    tuple[DANISH, UKRAINIAN]: PromptConfig(
```

src/euroeval/data_models.py
```diff
@@ -241,12 +241,15 @@ class DatasetConfig:
     unofficial: bool = False
 
     @property
-    def main_language(self) -> Language:
+    def main_language(self) -> Language | tuple[Language, Language]:
         """Get the main language of the dataset.
 
         Returns:
             The main language.
```
saattrupdan: Missing mention of the tuple.

src/euroeval/data_models.py
saattrupdan, line 74: Missing mention of the tuple.

src/euroeval/data_models.py
```diff
@@ -334,14 +337,32 @@ def prompt_template(self) -> str:
...
         return instruction_prompt
 
+    @property
+    def _prompt_config(self) -> PromptConfig:
+        """Get the task's `PromptConfig` for this dataset.
+
+        Most tasks have a template dictionary keyed by `Language`, but machine
+        translation uses `(source_language, target_language)` tuple.
+        """
+        if self.task.name == "machine-translation":
```
saattrupdan: I'd avoid comparing strings like this, as it's super error prone. I'd compare the task directly (to `MT`).

src/euroeval/language_pairs_mt.py
```diff
@@ -0,0 +1,6 @@
+"""Language pairs for machine translation."""
+
+from .languages import DANISH, UKRAINIAN, Language
+
+DANISH_TO_UKRAINIAN: tuple[Language, Language] = (DANISH, UKRAINIAN)
+UKRAINIAN_TO_DANISH: tuple[Language, Language] = (UKRAINIAN, DANISH)
```
saattrupdan: I don't think this module is needed?

src/euroeval/data_models.py
```diff
@@ -334,14 +337,32 @@ def prompt_template(self) -> str:
     @property
     def instruction_prompt(self) -> str:
         """The prompt to use when evaluating instruction-tuned models."""
-        prompt_config = self.task.template_dict[self.main_language]
```
saattrupdan, line 339: Wouldn't this work now, if `self.main_language` can just be a tuple?

src/euroeval/prompt_templates/machine_translation.py
oliverkinch: Yes, let's do that

oliverkinch: We now have `(DANISH, UKRAINIAN): PromptConfig(`

src/euroeval/language_pairs_mt.py
oliverkinch: It is now deleted

src/euroeval/data_models.py
oliverkinch: Now comparing directly to the `MT` task.

This however requires a local import. See [b945478](https://github.com/EuroEval/EuroEval/pull/1445/commits/b9454782665d0436f045a5cabcd60443d0fea698)

src/euroeval/data_models.py
oliverkinch: Yes it works now, see [b013a70](https://github.com/EuroEval/EuroEval/pull/1445/commits/b013a70507f0f14660c0bc5e96b2c20251697d1a). I don't always find type errors so easy to understand and fix, but again.. it works now

src/euroeval/data_models.py
```diff
@@ -70,6 +70,8 @@ class Task:
             The task group of the task.
         template_dict:
             The template dictionary for the task, from language to prompt template.
+            For machine translation, the key is a tuple
+            ``(source_language, target_language)``.
```
saattrupdan, line 76: Maybe instead of being specific to machine translation, just mention that it goes from language or tuple of languages to prompt template - then we don't need to update the docstring if more such tasks are added in the future.

saattrupdan (commented):
Looks pretty good now! Just missing a creation script, and I think we might be ready to go.

saattrupdan:
We should also update the docs, of course. Both with the new task as well as the new dataset.

oliverkinch:
I will not be able to finish this before my vacation. I am back again January 26.

I am in the process of building an actual machine translation dataset with a client, but as I am not sure when that dataset will be ready, I will probably finish this PR based on a draft dataset. And I also have to modify the docs as you mention.

saattrupdan:
Closing this in favour of [this PR](https://github.com/EuroEval/EuroEval/pull/1551).

PR #1454, opened 2026-01-13
LLMAsAJudge: Batch scoring function

oliverkinch:


src/euroeval/metrics/llm_as_a_judge.py
```diff
@@ -227,6 +248,51 @@ def _apply_user_prompt(self, prediction: str, condition: str | None = None) -> s
...
+                "provide only one of them."
+            )
+        if scoring_fn is not None:
+            scoring_fn_nonnull = scoring_fn
+
+            def batch_fn(
+                outputs: list[BaseModel], dataset: "Dataset | None" = None
+            ) -> float:
+                # Kept to satisfy the BatchScoringFunction signature.
+                del dataset
```
saattrupdan: Don't see why we need this explicitly?

```suggestion
```

src/euroeval/types.py
```diff
@@ -13,6 +13,8 @@
         MistralCommonBackend as MistralCommonTokenizer,
     )
 
+from pydantic import BaseModel
```
saattrupdan: Can you move this into the `t.TYPE_CHECKING` conditional, as that's the only thing it's used for?

src/euroeval/metrics/llm_as_a_judge.py
```diff
@@ -181,22 +188,36 @@ def __call__(
...
+            except ValidationError:
+                outputs_raw.append(None)
+
+        num_none: int = sum(output is None for output in outputs_raw)
+        if num_none:
+            log(
+                f"Could not parse/validate {num_none:,} of {len(outputs_raw):,} judge "
+                f"outputs for metric {self.pretty_name!r}. These will be ignored.",
+                level=logging.WARNING,
+            )
```
saattrupdan, line 207: Consider moving this log to debug logging, to avoid potential spam. If they're all None then they will get the warning below anyway.

saattrupdan (changes requested):
Just a few changes 🙂 

Also, can you update the changelog?

PR #1472, opened 2026-01-27
Docs-update

KennethEnevoldsen:
I did a few changes here.

- 1) Removed duplicate quickstart section
- 2) Converted the command line, script, docker section to tabs (let me know if you don't like this then I can revert it
- 3) Added script name to the custom dataset section (to better allow people to skim it)

### DIFF
--- a/docs/python-package.md
+++ b/docs/python-package.md
@@ -197,30 +197,35 @@ an Ollama model hosted locally:
 ## Benchmarking in an offline environment
 
 If you need to benchmark in an offline environment, you need to download the models,
-datasets and metrics beforehand. This can be done by adding the `--download-only`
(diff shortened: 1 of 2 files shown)


src/scandeval/model_loading.py
```diff
@@ -0,0 +1,457 @@
+"""Functions related to the loading of models."""
```
saattrupdan: Why did you create a new `scandeval` directory? 😮

KennethEnevoldsen: accidental merge I believe - sorry about that - reverted

KennethEnevoldsen:
That seems like a merge error. I apparently had a old fork that it merged into to created the PR (wasn't intentional)

PR #1487, opened 2026-01-28
Feat/support vllm metal

saattrupdan:
### Added

- Added support for vLLM Metal, so that generative models can now be evaluated on Apple
  Silicon.

### DIFF
--- a/.github/workflows/auto_assign_reviewers.yaml
+++ b/.github/workflows/auto_assign_reviewers.yaml
@@ -5,6 +5,9 @@ on:
 
 jobs:
   add-reviews:
+    permissions:
+      contents: read
+      pull-requests: write
     runs-on: ubuntu-latest
     steps:
       - uses: kentaro-m/auto-assign-action@v2.0.1
--- a/docs/python-package.md
+++ b/docs/python-package.md
@@ -225,9 +225,9 @@ benchmarker.benchmark(
 ///
 
 !!! note
-    Offline benchmarking of adapter models is not currently supported, meaning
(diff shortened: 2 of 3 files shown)


saattrupdan:
Currently facing [this xgrammar issue](https://github.com/vllm-project/vllm/issues/31901) when evaluating anything that requires structured generation, which are all classification tasks as well as NER.

saattrupdan:
For summarisation, my M4 Max laptop zero-shot evaluated `PleIAs/Monad` on the validation split of the Danish summarisation dataset, which took 40 minutes in total for all 10 iterations. This was with the `VLLM_METAL_USE_MLX=1` environment variable set. The GPU was definitely used (checked with the Activity Monitor), but only <5% of the time, so it's currently highly inefficient.

pyproject.toml
```diff
@@ -46,18 +46,26 @@ dependencies = [
 
 [project.optional-dependencies]
 generative = [
+    "timm>=1.0.19",
+    # vLLM setup for MacOS
+    "vllm==0.11.0; platform_system == 'Darwin'",
```
KennethEnevoldsen, line 51: yeah couldn't get it to work with other versions (but would probably want it to be relaxed)

saattrupdan: Yeah it seems that the newer ones all require CUDA-specific dependencies, which could hopefully be made Linux-specific at least.

saattrupdan:
@KennethEnevoldsen Merging this in, since it still allows local type hinting and _some_ local evaluations to be run - added disclaimer in changelog on this.

PR #1499, opened 2026-01-31
feat: add attention backend configuration for vllm in cli

### LINKED ISSUE #1488: [BUG] Setting attention backend in newer vLLM versions
### 🐛 Describe the bug

When running evaluations with `FULL_LOG=1` we get the warning:

```bash
Using VLLM_ATTENTION_BACKEND environment variable is deprecated and will be removed in v0.14.0 or v1.0.0, whichever is soonest. Please use --attention-config.backend command line argument or AttentionConfig(backend=...) config field instead.
```

We should thus change this to using the `attention_config` argument instead, and make this configurable in the CLI.

### Operating System

Linux

### Device

CUDA GPU

### Python version

3.12.x

### EuroEval version

16.11.0.dev

### Transformers version

4.56.0

### vLLM version

0.14.1

SwekeR-463:
Fixes #1488 

#### Changes:

- Removed the `attention_backend` context manager that set the env var
- Added `--attention-backend` CLI option to configure the backend
- Now uses vLLM's `AttentionConfig(backend=...)` API instead
- Defaults to `FLASHINFER` (same as previous behavior)

### DIFF
--- a/tests/test_benchmarker.py
+++ b/tests/test_benchmarker.py
@@ -505,9 +505,8 @@ def test_prepare_dataset_configs(
     dataset_names: list[str], dataset_configs: list[DatasetConfig]
 ) -> None:
     """Test that the `prepare_dataset_configs` function works as expected."""
-    assert (
+    assert set(
         prepare_dataset_configs(
             dataset_names=dataset_names, custom_datasets_file=Path("custom_datasets.py")
         )
-        == dataset_configs
-    )
+    ) == set(dataset_configs)


src/euroeval/benchmark_modules/vllm.py
```diff
@@ -165,22 +165,25 @@ def __init__(
...
+        # 2. Otherwise, check if model requires a specific backend
+        # 3. If neither, default to FLASHINFER (same as previous env var behavior)
+        attention_backend = benchmark_config.attention_backend
+        if attention_backend is None:
+            for pattern, backend in MODELS_REQUIRING_CUSTOM_ATTENTION_BACKENDS.items():
+                if re.search(pattern=pattern, string=model_config.model_id):
+                    attention_backend = backend
+                    break
+            # Default to FLASHINFER if no specific backend required
+            if attention_backend is None:
```
saattrupdan: For simplicity, we can simply use the `for-else` pattern here:
```suggestion
            else:
```

src/euroeval/benchmark_modules/vllm.py
```diff
@@ -944,7 +947,9 @@ def trainer_class(self) -> t.Type["Trainer"]:
 
 
 def load_model_and_tokeniser(
-    model_config: "ModelConfig", benchmark_config: "BenchmarkConfig"
+    model_config: "ModelConfig",
+    benchmark_config: "BenchmarkConfig",
+    attention_backend: str | None,
```
saattrupdan: It doesn't seem like this function will ever be called with `attention_backend=None`, so this should probably be changed:
```suggestion
    attention_backend: str,
```

src/euroeval/benchmark_modules/vllm.py
```diff
@@ -953,6 +958,8 @@ def load_model_and_tokeniser(
             The model configuration.
         benchmark_config:
             The benchmark configuration.
+        attention_backend:
+            The attention backend to use. If None, the default backend will be used.
```
saattrupdan: It doesn't seem like this function will ever be called with `attention_backend=None`, so this should probably be changed:
```suggestion
            The attention backend to use.
```

src/euroeval/benchmark_modules/vllm.py
```diff
@@ -1085,6 +1092,12 @@ def load_model_and_tokeniser(
             if internet_connection_available() or Path(model_id).is_dir()
             else resolve_model_path(download_dir=download_dir)
         )
+
+        # Build attention config if attention backend is specified
+        attention_config = None
+        if attention_backend is not None:
```
saattrupdan: This doesn't make much sense, to set it to None and check for not None.

If we change the input type of `attention_backend` to never be None, then we don't need these two lines at all.

src/euroeval/data_models.py
```diff
@@ -553,6 +556,7 @@ class BenchmarkConfig:
     few_shot: bool
     num_iterations: int
     gpu_memory_utilization: float
+    attention_backend: str | None
```
saattrupdan: Can we use the [AttentionBackendEnum enum](https://github.com/vllm-project/vllm/blob/88c3e114d8f564d1e88ff2ef5be56a979792e90b/vllm/v1/attention/backends/registry.py#L34) from vLLM as the attention_backend type instead of a string? For the CLI we can use `click.Choice` to validate the input as well.

SwekeR-463:
> Thanks for the PR! Overall looks good, but have a couple of points and suggestions.
> 
> Also, can you please add an entry in the changelog under "Changed"?

Have made the changes as suggested and also tell if the ChangeLog part is written correctly.

saattrupdan (approved):
Looks great now, thanks!

PR #1512, opened 2026-02-05
Feat/support dataset configs on hf datasets

### LINKED ISSUE #1507: [FEATURE REQUEST] Allow evaluation on a HF dataset

Currently, it is possible to evaluate a dataset using euroeval only if it is either implemented in euroeval or if it is local, but it would be great to add it be able to simply point to a dataset on Huggingface, like:

```bash
euroeval --dataset org/my-dataset --model <model-id>
```

I could imagine you could do it with a file like euroeval.config (or even euroeval.py, but there might be some security concerns there)

--- 

_Discussed during dfm model training meeting_

saattrupdan:
### Added
- Added support for evaluating Hugging Face dataset repos directly, if they have a
  `euroeval_config.py` file. We plan to allow support for a JSON/YAML config file in the
  future, for simpler datasets.

### Changed

- Replaced all `DatasetConfig` arguments starting with underscores with their
  non-underscored version (e.g., `_labels` -> `labels`), as this caused some confusion
  when defining custom datasets. We still maintain the underscored versions for
  backwards compatibility, but raise a warning when using them.

### Deprecated

- All underscored versions of `DatasetConfig` arguments are deprecated. Please use their
  non-underscored version instead.

Also split up `utils.py` into a some more `X_utils.py` modules. That's what causes the insane amount of changed files. The key new functionality can be found in the `custom_dataset_configs.py` module.

Closes #1507

### DIFF
--- a/src/euroeval/data_loading.py
+++ b/src/euroeval/data_loading.py
@@ -72,9 +72,7 @@ def load_data(
     # Remove empty examples from the datasets
     for text_feature in ["tokens", "text"]:
         for split in splits:
-            if split is None:
-                continue
-            elif text_feature in dataset[split].features:
+            if text_feature in dataset[split].features:
                 dataset = dataset.filter(lambda x: len(x[text_feature]) > 0)
 
     # If we are testing then truncate the test set, unless we need the full set for


docs/python-package.md
```diff
@@ -279,18 +285,57 @@ benchmarker = Benchmarker()
...
+/// tab | Hugging Face Hub dataset
+
+For a dataset that is accessible via the Hugging Face Hub, you simply need to create a
+file called `euroeval_config.py` in the root of your repository, in which you define
+the associated dataset configuration. Note that you don't need to specify the `name`,
+`pretty_name` or `source` arguments in this case, as these are automatically inferred
+from the repository name. Here is an example of a simple text classification
+dataset with two classes:
+
+```python title="euroeval_config.py"
```
KennethEnevoldsen, line 298: Should we add an example on huggingface that people can run? (also just to see the full setup)

saattrupdan: Yeah that's a good idea - I've [added one](https://huggingface.co/datasets/EuroEval/test_dataset) now.

Added mention of this in the docs now as well.

PR #1516, opened 2026-02-06
Fix/unresponsive cli when evaluating local model

### LINKED ISSUE #1508: CLI is unresponsive when running with an API
when running the following:

```bash
euroeval --dataset my-dataset --model my-model --api-base http://localhost:1234       
```

The process hangs for more than 2 minutes (before I stop it). I have the following file:

```
# custom_datasets.py
from euroeval import TEXT_CLASSIFICATION, DatasetConfig
from euroeval.languages import ENGLISH

MY_CONFIG = DatasetConfig(
    name="my-dataset",
    pretty_name="My Dataset",
    source=dict(train="data/train.csv", test="data/test.csv"),
    task=TEXT_CLASSIFICATION,
    languages=[ENGLISH],
    _labels=["positive", "negative"],
)
```

saattrupdan:
### Changed

- Now logs when the model inference service is temporarily unavailable, even when the
  verbose flag is not set.
- When evaluating local models, we now automatically add the "/v1" suffix to the API
  base URL if not present and required by the API.

Fixes #1508


KennethEnevoldsen:
Thanks for the fix!

PR #1527, opened 2026-02-11
Add WMT24++ translation task and datasets

rlrs:
@oliverkinch this one should be easier to review, more metrics and other features can come later

### DIFF
--- a/src/euroeval/benchmark_config_factory.py
+++ b/src/euroeval/benchmark_config_factory.py
@@ -3,6 +3,7 @@
 import collections.abc as c
 import importlib.util
 import logging
+import sys
 import typing as t
 from pathlib import Path
 


src/euroeval/benchmark_config_factory.py
```diff
@@ -233,7 +234,7 @@ def prepare_dataset_configs(
         if closest_distance < 5:
             msg += f" Maybe you meant to use {closest_match!r}?"
         log(msg, level=logging.ERROR)
-        sys.exit(1)
+        raise InvalidBenchmark(msg)
```
saattrupdan, line 237: These should be reverted to `sys.exit(1)` for now, as they're called too early for graceful exception handling of `InvalidBenchmark` exceptions. The tests regarding these have been fixed on main.

src/euroeval/data_loading.py
```diff
@@ -235,6 +235,17 @@ def load_raw_data(
...
+
+    if dataset_config.preprocess_fn is not None:
+        dataset = dataset_config.preprocess_fn(dataset)
+        if not isinstance(dataset, DatasetDict):
+            raise InvalidBenchmark(
+                "Dataset preprocessing must return a DatasetDict, but got "
+                f"{type(dataset)}."
+            )
+
+    dataset = _split_dataset_if_needed(dataset=dataset, dataset_config=dataset_config)
```
saattrupdan, line 247: The whole dataset preprocessing function is nice, but conflicts with the current setup. Datasets that are built into the framework have creation scripts in the separate scripts folder, so that these dataset processors are not needed. Feel free to open a separate issue on replacing all dataset scripts with processors instead, if needed.

Note that the scripts carry additional value in that they're future preserving. Many of the evaluation datasets in the framework have been removed from Hugging Face since they were added, so if we simply used the original sources, we would be in trouble.

See more in the [new dataset guide](https://github.com/EuroEval/EuroEval/blob/main/NEW_DATASET_GUIDE.md).

src/euroeval/tasks.py
```diff
@@ -94,6 +95,20 @@
     default_allowed_model_types=[ModelType.GENERATIVE],
 )
 
+MT = Task(
+    name="machine-translation",
+    task_group=TaskGroup.TEXT_TO_TEXT,
+    template_dict=MT_TEMPLATES,
+    metrics=[
+        m.bert_score_metric,
+        m.sacrebleu_metric,
```
saattrupdan, line 104: Note that the order here matters, and only the first metric is used for computing the overall score. Does BERTScore still give a better correlation than SacreBLEU?

saattrupdan (changes requested):
An initial review.

My main concern is that this PR is trying to do two things: adding the new task as well as introducing a new dataset processing mechanism.

I would suggest the task and dataset are added in the usual fashion, as described in the [new dataset guide](https://github.com/EuroEval/EuroEval/blob/main/NEW_DATASET_GUIDE.md), and that you create an issue to convert all scripts to using the dataset processing functions (if that's what you were thinking).

Furthermore, the machine translation task needs to be added to `docs/tasks`, the datasets added to the respective languages in `docs/datasets`, and the change log updated.

saattrupdan:
Closing this PR in favour of [this one](https://github.com/EuroEval/EuroEval/pull/1552).

PR #1545, opened 2026-02-16
Add BFCL benchmark

harderj:
Add the BFCL-v2 tool calling benchmark (a partial version and english only now at first)

### DIFF
--- a/src/euroeval/data_models.py
+++ b/src/euroeval/data_models.py
@@ -15,7 +15,11 @@ import torch
 from datasets import DatasetDict
 from transformers.generation.configuration_utils import GenerationConfig
 
-from .constants import ATTENTION_BACKENDS, CHOICES_MAPPING, MAX_NUMBER_OF_LOGGING_LANGUAGES
+from .constants import (
+    ATTENTION_BACKENDS,
+    CHOICES_MAPPING,
+    MAX_NUMBER_OF_LOGGING_LANGUAGES,
+)
 from .enums import Device, GenerativeType, ModelType, TaskGroup
 from .exceptions import InvalidBenchmark
 from .languages import (
@@ -498,7 +502,11 @@ class DatasetConfig:
 
         # Build or assign the preprocessing function
         column_args_set = any(
-            (input_column != "text", target_column is not None, choices_column is not None)
+            (
+                input_column != "text",
+                target_column is not None,
+                choices_column is not None,
+            )
         )
         if preprocessing_func is not None and column_args_set:
             log_once(
@@ -507,9 +515,9 @@ class DatasetConfig:
                 "takes precedence and the column arguments will be ignored.",
                 level=logging.WARNING,
             )
-            self.preprocessing_func: (
-                c.Callable[[DatasetDict], DatasetDict] | None
-            ) = preprocessing_func
+            self.preprocessing_func: c.Callable[[DatasetDict], DatasetDict] | None = (
+                preprocessing_func
+            )
         elif column_args_set:
             # Determine the language-specific choices label
             main_lang = self.languages[0] if self.languages else None
--- a/src/euroeval/preprocessing.py
+++ b/src/euroeval/preprocessing.py
@@ -21,8 +21,8 @@ def merge_input_and_choices(
 
     Args:
         example:
-            A single dataset example with at least the ``input_column`` and the column(s)
-            named by ``choices_column``.
+            A single dataset example with at least the ``input_column`` and the
+            column(s) named by ``choices_column``.
         input_column:
             The name of the column containing the input text.
         choices_column:
@@ -98,11 +98,6 @@ def build_preprocessing_func(
     Returns:
         A callable that accepts a ``DatasetDict`` and returns a preprocessed
         ``DatasetDict``.
-
-    Raises:
-        InvalidBenchmark:
-            When the returned callable is called, if any configured column is absent
-            from all splits.
     """
     # Determine the standard target column for the task group
     if target_column is not None:
@@ -154,9 +149,7 @@ def build_preprocessing_func(
                 )
         if choices_cols is not None:
             for col in choices_cols:
-                col_found = all(
-                    col in split.column_names for split in dataset.values()
-                )
+                col_found = all(col in split.column_names for split in dataset.values())
                 if not col_found:
                     raise InvalidBenchmark(
                         f"The dataset is configured with a choices column "


src/euroeval/benchmark_modules/litellm.py
```diff
@@ -851,9 +836,10 @@ async def _generate_async(
             requests = [
                 add_semaphore_and_catch_exception(
                     router.acompletion(
-                        model=clean_model_id(
-                            model_id=model_id, benchmark_config=self.benchmark_config
-                        ),
+                        # model=clean_model_id(
```
harderj: Somehow this fucked up my local inference. But let's maybe ignore this for now and I'll revert this before undrafting this PR

saattrupdan:
@harderj Can you solve conflicts and ensure that `make check` and `make test` works for you locally (with the markdown check uncommented)? Also, please add a PR description and mark the PR as "Ready for Review" 🙂

src/euroeval/task_group_utils/tool_calling.py
```diff
@@ -0,0 +1,167 @@
...
+            if True:
+                score: float | None = metric(
+                    predictions=predictions,  #  type: ignore[bad-argument-type]
+                    references=labels,  #  type: ignore[bad-argument-type]
+                    dataset=dataset,
+                    dataset_config=dataset_config,
+                    benchmark_config=benchmark_config,
+                )
+                break
+            # except Exception as e:
```
harderj: Ignore - will uncomment.

src/euroeval/benchmarker.py
harderj, line 1: ignore this file,  will revert.

src/euroeval/data_loading.py
harderj, line 1: ignore, will revert.

src/scripts/create_bfcl.py
harderj, line 1: ignore - will delete this file probably.

src/scripts/debug_main.py
harderj, line 1: can delete or keep for debugging purposes?

harderj:
@saattrupdan 
> @harderj Can you solve conflicts and ensure that `make check` and `make test` works for you locally (with the markdown check uncommented)? Also, please add a PR description and mark the PR as "Ready for Review" 🙂

I for sure intend to before undrafting - however I made this draft because I thought it might be good with a few inputs before that. Some of the changes are made in order to make debugging easier - I pointed these out in comments, so you can ignore those, and - if you have time - look at the rest and see if there are some obvious ways we should do things different already. I'll keep revising meanwhile.

cheers

src/euroeval/tasks.py
```diff
@@ -153,6 +156,40 @@
...
+}
+
+BFCL = Task(
+    name="bfcl",
+    task_group=TaskGroup.TOOL_CALLING,
+    template_dict=TOOL_CALLING_TEMPLATES,
+    metrics=[m.tool_calling_metric],
+    default_num_few_shot_examples=0,
+    default_max_generated_tokens=1000,
+    default_labels=["a"],  # TODO: what is this?,
```
harderj: Could use a hint for this

.pre-commit-config.yaml
```diff
@@ -34,14 +34,14 @@ repos:
...
     -   id: pyrefly-check
         name: Pyrefly (type checking)
         pass_filenames: false
--   repo: https://github.com/DavidAnson/markdownlint-cli2
-    rev: v0.20.0
-    hooks:
-    -   id: markdownlint-cli2
-        args:
-          - --fix
+# -   repo: https://github.com/DavidAnson/markdownlint-cli2
```
harderj: Ignore, will revert.

src/euroeval/tasks.py
saattrupdan: The `default_labels` is used for classification tasks, where the task has some standard labels. For sentiment classification these could be `positive`, `neutral` and `negative`, for instance. The `DatasetConfig`s can then overwrite these if necessary. For instance, a couple of sentiment classification datasets don't have the `neutral` label, so for these we'd set `labels=["positive", "negative"]` in the `DatasetConfig`. If the `labels` is not set in the `DatasetConfig` then it simply inherits from the `default_labels` from its task.

saattrupdan:
> @saattrupdan
> 
> > @harderj Can you solve conflicts and ensure that `make check` and `make test` works for you locally (with the markdown check uncommented)? Also, please add a PR description and mark the PR as "Ready for Review" 🙂
> 
> I for sure intend to before undrafting - however I made this draft because I thought it might be good with a few inputs before that. Some of the changes are made in order to make debugging easier - I pointed these out in comments, so you can ignore those, and - if you have time - look at the rest and see if there are some obvious ways we should do things different already. I'll keep revising meanwhile.
> 
> cheers

Ah right. If it's not ready for review yet then I'd rather have you just ask specific questions 🙂

src/euroeval/tasks.py
harderj: In this case it's irrelevant I guess? So maybe I set them to `None` or `[]`. Unresolve if needed.

src/euroeval/benchmark_modules/litellm.py
harderj: reverted

.pre-commit-config.yaml
harderj: reverted

harderj:
Perhaps wait with review until review by copilot is done.

src/euroeval/data_loading.py
```diff
@@ -135,6 +136,7 @@ def load_raw_data(
         for _ in range(num_attempts):
             try:
                 with no_terminal_output():
+                    load_dotenv()
```
saattrupdan: Isn't this already called in `__init__`? Do you find that the variables weren't loaded in at this point?

src/euroeval/metrics/tool_calling.py
```diff
@@ -0,0 +1,71 @@
...
+import collections.abc as c
+import json
+import typing as t
+
+if t.TYPE_CHECKING:
+    from datasets.arrow_dataset import Dataset
+
+    from euroeval.data_models import BenchmarkConfig, DatasetConfig
+
+from euroeval.metrics.base import Metric
```
saattrupdan: For consistency with the rest of the code base, we use relative imports within the package:

```suggestion
    from ..data_models import BenchmarkConfig, DatasetConfig

from .base import Metric
```

src/euroeval/metrics/tool_calling.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+def _evaluate_function_toolcall_response(
+    pred_calls_str: str, ref_calls_str: str, descriptions: list[dict]
+) -> bool:
```
saattrupdan, line 66: Missing docstring

src/euroeval/metrics/tool_calling.py
```diff
@@ -0,0 +1,71 @@
...
+        ref_name, ref_args = list(ref_call.items())[0]
+        if pred_call["function"] != ref_name:
+            return False
+        required_args = description["parameters"].get("required", None)
+        for key, values in ref_args.items():
+            if required_args and key not in required_args:
+                continue
+            if key not in pred_args or pred_args[key] not in values:
+                return False
+    return True
```
saattrupdan, line 140: This is quite a hefty block with lots of conditionals. Can you either include comments or label these with variable names, to make it easier to follow?

src/euroeval/metrics/tool_calling.py
```diff
@@ -0,0 +1,71 @@
...
+        """Calculate tool calling metric.
+
+        Returns:
+            The Score.
+        """
+        function_descriptions = [json.loads(f) for f in dataset["function"]]
+        results = []
+        for x in zip(predictions, references, function_descriptions):
+            results.append(_evaluate_function_toolcall_response(*x))
+        return sum(results) / len(results)
```
saattrupdan: We have to deal with the case where `results` is empty here, to avoid dividing by zero.

src/euroeval/prompt_templates/tool_calling.py
```diff
@@ -0,0 +1,27 @@
+"""Templates for the tool calling task."""
+
+import typing as t
+
+from euroeval.data_models import PromptConfig
```
saattrupdan:
```suggestion
from ..data_models import PromptConfig
```

src/euroeval/prompt_templates/tool_calling.py
```diff
@@ -0,0 +1,27 @@
...
+        default_prompt_template=(
+            "These are names and descriptions of functions available:\n"
+            "{function_info}\n"
+            "Given the following user request: '{question}', "
+            "answer with a list of function(s) to use in the exact format:\n"
+            "{output_format_example}\n"
+            "Each item should have 'function' (the function name) "
+            "and 'arguments' (an object with the parameters)."
+        ),
+        default_prompt_label_mapping="auto",
```
saattrupdan: Since there aren't any labels, we just set this to an empty dict instead, to make that clear:
```suggestion
        default_prompt_label_mapping=dict(),
```

src/euroeval/task_group_utils/tool_calling.py
```diff
@@ -0,0 +1,169 @@
+"""Utility functions related to the text-to-text task group."""
+
+import collections.abc as c
+import json
+import logging
+import typing as t
+
+import numpy as np
+
+from euroeval.constants import METRIC_ATTRIBUTES_TAKING_UP_MEMORY
```
saattrupdan:
```suggestion
from ..constants import METRIC_ATTRIBUTES_TAKING_UP_MEMORY
```

src/euroeval/tasks.py
```diff
@@ -1,5 +1,9 @@
 """All benchmarks tasks used in EuroEval."""
 
+import json
+
+from euroeval.prompt_templates.tool_calling import TOOL_CALLING_TEMPLATES
```
```suggestion
from .prompt_templates.tool_calling import TOOL_CALLING_TEMPLATES
```

src/euroeval/tasks.py
```diff
@@ -153,6 +157,40 @@
...
+                        "function": {"type": "string"},
+                        "arguments": {"type": "object"},
+                    },
+                    "required": ["function", "arguments"],
+                },
+            }
+        },
+        "required": ["invocations"],
+    },
+}
```
saattrupdan: This seems a bit hard to maintain. I like the idea of being able to include the structured output schema though. Perhaps we can just include the Pydantic model directly here, instead of this JSON schema?

src/scripts/create_bfcl.py
```diff
@@ -0,0 +1,98 @@
...
+
+from datasets import Dataset, DatasetDict
+from dotenv import load_dotenv
+
+
+def split_dataset_to_dict(items: list[dict]) -> DatasetDict:
+    """Split dataset into train/validation/test with deterministic random sampling.
+
+    Args:
+        items: List of dataset items to split
```
saattrupdan: [nit] Consistency with code base
```suggestion
        items: 
            List of dataset items to split
```

src/scripts/create_bfcl.py
```diff
@@ -0,0 +1,98 @@
...
+
+    Returns:
+        DatasetDict with train (10%), validation (10%), and test (80%) splits
```
saattrupdan: In general be careful about including precise values in comments and docstrings, since they will probably end up wrong, if these values change in the future.

```suggestion
        DatasetDict with train, validation, and test splits
```

src/scripts/create_bfcl.py
```diff
@@ -0,0 +1,98 @@
...
+    val_size = int(0.1 * total)
+
+    train_items = shuffled_items[:train_size]
+    val_items = shuffled_items[train_size : train_size + val_size]
+    test_items = shuffled_items[train_size + val_size :]
+
+    return DatasetDict(
+        {
+            "train": Dataset.from_list(train_items),
+            "validation": Dataset.from_list(val_items),
```
saattrupdan: For consistency with the other datasets, we use 'val' as the name for the validation split:

```suggestion
            "val": Dataset.from_list(val_items),
```

src/scripts/create_bfcl.py
```diff
@@ -0,0 +1,98 @@
...
+            "test": Dataset.from_list(test_items),
+        }
+    )
+
+
+def _load_jsonl(x: str | Path) -> list:
```
saattrupdan: Missing docstring

src/scripts/create_bfcl.py
```diff
@@ -0,0 +1,98 @@
...
+    if isinstance(x, Path):
+        x = x.read_text()
+    return [json.loads(line) for line in x.splitlines()]
+
+
+def _load_jsonl_from_url(url: str) -> list:
```
saattrupdan: Missing docstring

src/scripts/create_bfcl.py
```diff
@@ -0,0 +1,98 @@
...
+    with urlopen(url) as r:
+        return _load_jsonl(r.read().decode())
+
+
+def main() -> None:
```
saattrupdan, line 13: [nit] This should be the top function in the script, to make it easier to read the script.

src/scripts/debug_main.py
saattrupdan: I'd delete it from the repo. Also, it might be easier for you to debug simply with the associated CLI command:

```bash
euroeval -t bfcl -l en -m Qwen/Qwen3-4B-Instruct-2507 --debug
```

src/euroeval/tasks.py
```diff
@@ -153,6 +157,40 @@
...
+
+BFCL = Task(
+    name="bfcl",
```
saattrupdan: Consider calling this "tool-calling" or something like that, since BFCL is just the dataset concretising the task. If other tool calling benchmarks appear in the same format, they would then still fall under this task, even if they're not called BFCL.

README.md
```diff
@@ -20,7 +20,7 @@ ______________________________________________________________________
 [![Second paper](https://img.shields.io/badge/arXiv-2406.13469-b31b1b.svg)](https://arxiv.org/abs/2406.13469)
 [![License](https://img.shields.io/github/license/EuroEval/EuroEval)](https://github.com/EuroEval/EuroEval/blob/main/LICENSE)
 [![LastCommit](https://img.shields.io/github/last-commit/EuroEval/EuroEval)](https://github.com/EuroEval/EuroEval/commits/main)
-[![Code Coverage](https://img.shields.io/badge/Coverage-71%25-yellow.svg)](https://github.com/EuroEval/EuroEval/tree/main/tests)
+[![Code Coverage](https://img.shields.io/badge/Coverage-65%25-yellow.svg)](https://github.com/EuroEval/EuroEval/tree/main/tests)
```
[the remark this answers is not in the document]
harderj: noted. I opt to ignore this, this time unless someone stops me

src/scripts/debug_main.py
harderj: Does this integrate with vscode and breakpoints etc?

src/euroeval/tasks.py
harderj: yes, copilot had a similar idea - I think I did this initially, but ran into trouble. I've forgotten exactly what so I'll redo the exercise now and get back

src/euroeval/metrics/tool_calling.py
harderj: already fixed

src/euroeval/metrics/tool_calling.py
harderj: added some comments

src/euroeval/data_loading.py
harderj: whoops yes good catch

harderj:
> Lots of progress, we're getting way closer! I've got some more comments.

thanks;)

> On a more general note, I'd consider whether we can make tool calling use the text-to-text task group instead, as that means we can re-use a larger part of the existing functionality (e.g., compute metrics, benchmark modules). We did this for instruction following as well. You've basically done it already, you just need to re-package the columns in the datasets:
> 
> * `text` column for the input, which should contain all of the prompt, including the function info and question.
> * `target_text` column for the output, which is basically the ground truth, as far as I can see.
> 
> These will be set up in the creation script. You can then just use the `EMPTY_TEMPLATES` prompt template, so no need to create a new one there either.
> 
> If you think there'd be annoying issues with this then let me know - it's just an idea to avoid adding too much new code 🙂

I agree that the `compute_metrics` part is certainly not 'DRY', however I cannot see how to squeeze the tool-calling exactly into the current text-to-text scheme. Could be there are some ideas I'm missing. Maybe we can discuss it - or we could make create a ticket for it and make it a new PR

src/euroeval/tasks.py
harderj: I think i managed to fix it now :+1:

src/scripts/debug_main.py
harderj: deleting

src/euroeval/benchmark_modules/litellm.py
```diff
@@ -1887,6 +1887,8 @@ def clean_model_id(model_id: str, benchmark_config: BenchmarkConfig) -> str:
     # inference endpoints, LiteLLM gets confused since it's already using the `openai/`
     # prefix. We thus have to add it twice, and this hack here is to ensure that we
     # don't store the results with model ID `openai/openai/...`.
+    # This is where the "openai/openai/..." bug arose. I'm guessing we keep it,
```
harderj: We should delete this comment - just wanted to point it out

harderj:
Ready for review again @saattrupdan

src/euroeval/benchmark_modules/litellm.py
```diff
@@ -1624,7 +1628,7 @@ def get_generation_kwargs(self, dataset_config: DatasetConfig) -> dict[str, t.An
                     "enable it.",
                     level=logging.DEBUG,
                 )
-            elif supports_response_schema(model=self.model_config.model_id):
```
saattrupdan: Why remove this?

src/euroeval/benchmark_modules/litellm.py
saattrupdan: Thanks! Maybe you can open a separate issue on how to reproduce the bug?

src/euroeval/benchmark_modules/vllm.py
```diff
@@ -486,6 +486,14 @@ def generate(self, inputs: dict) -> "GenerativeModelOutput":
...
                 level=logging.DEBUG,
             )
+        elif (
+            self.dataset_config.task.uses_structured_output
+            and self.dataset_config.task.structured_output_model is not None
+        ):
+            structured_outputs = StructuredOutputsParams(
+                json=self.dataset_config.task.structured_output_model.model_json_schema(),
+                disable_fallback=True,
+            )
```
saattrupdan: To make it slightly more readable, can we merge this block with the one underneath? Then we'll just have an inner conditional, first checking if the task is NER, then checking if `structured_output_model` is set, and if none of those then it raises an InvalidBenchmark error?

src/euroeval/data_models.py
```diff
@@ -108,6 +108,8 @@ class Task:
         uses_structured_output (optional):
             Whether the task uses structured output. If True, the task will return
             structured output (e.g., BIO tags for NER). Defaults to False.
+        structured_output_schema (optional):
+            Schema to use for structured output as a (deserializable) JSON string.
```
saattrupdan, line 118: Specify here that it can have a None value, what None does, and that it defaults to None

src/euroeval/metrics/tool_calling.py
```diff
@@ -0,0 +1,102 @@
...
+    Returns:
+        True: success, False: failure
+    """
+    try:
+        pred_calls = json.loads(pred_calls_str)
+        assert isinstance(pred_calls, list)
+    except json.JSONDecodeError:
+        return False
+    except AssertionError:
+        return False
```
saattrupdan: [nit] Slightly shorter.

```suggestion
    except (json.JSONDecodeError, AssertionError):
        return False
```

src/euroeval/metrics/tool_calling.py
```diff
@@ -0,0 +1,102 @@
...
+    ref_calls = json.loads(ref_calls_str)
+    # number of predicted function calls should equal the reference
+    if len(pred_calls) != len(ref_calls):
+        return False
+    for pred_call, ref_call, description in zip(pred_calls, ref_calls, descriptions):
+        # each predicted function call should be a dict
+        if not isinstance(pred_call, dict):
+            return False
+        if "function" not in pred_call:
```
saattrupdan, line 101: I would probably debug log here if this is not the case. You could use the `log_once` function to avoid spam. Same thing with the "arguments" check below.

src/euroeval/metrics/tool_calling.py
```diff
@@ -0,0 +1,102 @@
...
+
+if t.TYPE_CHECKING:
+    from datasets.arrow_dataset import Dataset
+
+    from ..data_models import BenchmarkConfig, DatasetConfig
+
+from ..metrics.base import Metric
+
+
+def _evaluate_function_toolcall_response(
```
saattrupdan, line 64: Generally speaking, it would be nice if the blocks in this function were separated with double newlines, as it eases reading it

saattrupdan: For instance, all JSON parsing checking in one block, then checking the number of refs equals the number of preds, and so on.

src/euroeval/metrics/tool_calling.py
```diff
@@ -0,0 +1,102 @@
...
+        predictions: c.Sequence,
+        references: c.Sequence,
+        dataset: "Dataset",
+        dataset_config: "DatasetConfig",
+        benchmark_config: "BenchmarkConfig",
+    ) -> float | None:
+        """Calculate tool calling metric.
+
+        Returns:
+            The Score.
```
saattrupdan: Mention that it can be None, and what it means.

src/euroeval/prompt_templates/tool_calling.py
```diff
@@ -0,0 +1,58 @@
...
+    f'and "{TOOL_CALLING_ARGUMENTS_KEY}" '
+    "(a dict with the keywords and values describing the parameters)."
+)
+
+_DEFAULT_TOOL_CALLING_PROMPT_TEMPLATE_EN = (
+    "A list of names and descriptions of functions available, "
+    "and a user question is given below: \n"
+    "{text}\n"
+    f"{_DEFAULT_TOOL_CALLING_PROMPT_POSTFIX_EN}"
+)
```
saattrupdan: [nit] Can you move these down into the `PromptConfig`? Both for consistency with the other prompt template files, and it makes reading it a bit easier.

Also, when we are happy with the prompts, you can use one of the commercial LLMs to translate it into all the relevant ~30 languages (Gemini-3-flash and -pro works quite well, but I'm sure GPT and Claude are fine too).

src/euroeval/tasks.py
```diff
@@ -169,6 +170,20 @@
 )
 
 
+TOOL_CALLING = Task(
+    name="tool-calling",
+    task_group=TaskGroup.TEXT_TO_TEXT,
+    template_dict=TOOL_CALLING_TEMPLATES,
+    metrics=[m.tool_calling_metric],
+    default_num_few_shot_examples=0,
+    default_max_generated_tokens=1000,
```
saattrupdan: Seems a bit much - is this a guess or well-founded? A large number will increase evaluation time a fair amount.

src/scripts/create_bfcl.py
```diff
@@ -0,0 +1,120 @@
...
+        url_prefix = (
+            "https://raw.githubusercontent.com/ShishirPatil/gorilla"
+            "/refs/heads/main/berkeley-function-call-leaderboard/bfcl_eval/data"
+        )
+        input_url = f"{url_prefix}/BFCL_v4_{subset_name}.json"
+        ground_truth_url = f"{url_prefix}/possible_answer/BFCL_v4_{subset_name}.json"
+        print(f"Loading dataset '{subset_name}'")
+        inputs = _load_jsonl_from_url(input_url)
+        ground_truth = _load_jsonl_from_url(ground_truth_url)
+        # Join input and ground_truth by 'id' key
```
saattrupdan, line 38: [nit] Separation of code blocks for readability.

```suggestion

        # Join input and ground_truth by 'id' key
```

src/scripts/create_bfcl.py
```diff
@@ -0,0 +1,120 @@
...
+    Returns:
+        DatasetDict with train, validation, and test splits
+    """
+    random.seed(42)
+    shuffled_items = items.copy()
+    random.shuffle(shuffled_items)
+
+    total = len(shuffled_items)
+    train_size = int(0.1 * total)
+    val_size = int(0.1 * total)
```
saattrupdan: It's useful to know exactly how big the dataset is, also in the code, so hardcoding it is preferred. This makes writing documentation easier, and also makes it clear whether it warrants cutting down in size, if the dataset is too large.

src/euroeval/benchmark_modules/litellm.py
harderj: ah yes, should have written a comment about this beforehand. It excludes a lot of models for which structured output generation is actually working (e.g. Qwen/Qwen2.5/3)

src/euroeval/benchmark_modules/litellm.py
harderj: link to issue:
https://github.com/EuroEval/EuroEval/issues/1557

src/euroeval/benchmark_modules/litellm.py
harderj: I made this work in another way that keeps this logic, while allowing the structured generation with Qwen in  tool calling. As we discussed I'll make a separate issue for making sure `supports_response_schema` does not unnecessarily exclude any models from being used with structured generation

harderj: Link to issue: https://github.com/EuroEval/EuroEval/issues/1559

src/euroeval/tasks.py
harderj: umiddelbart lader det til at de fleste ground truth er <400 tokens, baseret på nogle stikprøver

de anbefaler selv ~600 tokens

https://achan2013.medium.com/how-many-tools-functions-can-an-ai-agent-has-21e0a82b7847
siger ~1600 tokens.

vi kan godt sige ændre fra 1000->500 for min skyld.

harderj:
Ready for review again @saattrupdan

CHANGELOG.md
```diff
@@ -9,6 +9,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Added
 
+- A task *Tool Calling* and a dataset under this task *bfcl* a subset of the
+Berkeley Function Calling Leaderboard benchmark (v4).
+Currently only supported for English.
+
+### Added
+
```
saattrupdan, line 40: Double headline:
```suggestion
```

CHANGELOG.md
```diff
@@ -690,12 +696,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 - Enable support to evaluate Mistral models with their custom `mistral-common`
   tokeniser, which includes all recent Mistral models. Note that we currently assume
-  that all of these models are instruction-tuned decoder models (which _is_ true
+  that all of these models are instruction-tuned decoder models (which *is* true
```
saattrupdan: Why all these "_ -> *" changes?

docs/tasks/tool-calling.md
```diff
@@ -0,0 +1,28 @@
...
+## 🛠️ How to run
+
+In the command line interface of the [EuroEval Python package](/python-package), you
+can benchmark your favorite model on the tool calling task like so:
+
+```bash
+euroeval --model <model-id> --task tool-calling --language en
+```
+
+Currently only the english language is supported
```
saattrupdan: No need to mention the language-specific restrictions here, since that'll (hopefully) get outdated really quickly anyway:
```suggestion
```bash
euroeval --model <model-id> --task tool-calling
```
```

File: src/euroeval/benchmark_modules/litellm.py [hunk 38]
```diff
@@ -1617,7 +1621,16 @@ def get_generation_kwargs(self, dataset_config: DatasetConfig) -> dict[str, t.An
...
+            if dataset_config.task.structured_output_format is not None:
+                pydantic_class = dataset_config.task.structured_output_format
+                generation_kwargs["response_format"] = pydantic_class
+                log_once(
+                    "Enabling structured generation for model "
+                    f"{self.model_config.model_id!r} with response_format "
+                    f"{pydantic_class.model_json_schema()}.",
+                    level=logging.DEBUG,
+                )
+            elif self.generative_type == GenerativeType.REASONING:
```
saattrupdan: Reasoning models typically cannot handle structured generation, so we'd need to not force that if the model is of that type - so this should be moved up in the conditional chain

docs/datasets/english.md
```diff
@@ -1246,3 +1246,135 @@ You can evaluate this dataset directly as follows:
 ```bash
 euroeval --model <model-id> --dataset valeu-en
 ```
+
+## European Values
```
saattrupdan:
```suggestion
## Tool calling
```

docs/datasets/english.md
```diff
@@ -1246,3 +1246,135 @@ You can evaluate this dataset directly as follows:
...
+
+### BFCL
+
+This dataset is an excerpt from the Berkeley Function Calling Leaderboard (v4) dataset,
+containing the following subsets:
```
saattrupdan: Can you link to the associated paper, documenting the dataset? DOI if possible 🙂

docs/datasets/english.md
```diff
@@ -1246,3 +1246,135 @@ You can evaluate this dataset directly as follows:
...
+- "live_multiple",
+- "live_parallel_multiple",
+- "live_parallel",
+- "live_simple",
+- "multiple",
+- "parallel_multiple",
+- "parallel",
+- "simple_java",
+- "simple_javascript",
+- "simple_python",
```
saattrupdan: Do we use all subsets?

docs/datasets/english.md
```diff
@@ -1246,3 +1246,135 @@ You can evaluate this dataset directly as follows:
...
+
```
saattrupdan, line 1271: Missing info on how many samples are in the original dataset, and how many we're evaluating on

docs/tasks/tool-calling.md
```diff
@@ -0,0 +1,28 @@
...
+We use a uniform custom prompt and structured JSON output generation to obtain
+predictions from a given model.
+
+When evaluating generative models, we allow the model to generate 500 tokens on this
+task.
+
+## 📊 Metrics
+
+We use simple accuracy where a prediction is positive only if the list of functions
+and required arguments exactly matches one of the options given in the ground truth.
```
saattrupdan, line 17: So the ground truth has multiple different correct answers? This is not really made clear here. Maybe you can give a short example?

src/euroeval/prompt_templates/tool_calling.py
```diff
@@ -0,0 +1,53 @@
...
+    """For structured generation purposes in tool-calling."""
+
+    function: str
+    arguments: dict[str, str]
+
+
+class ToolCallingResponse(pydantic.RootModel[list[ToolCall]]):
+    """For structured generation purposes in tool-calling."""
+
+    root: list[ToolCall]
```
saattrupdan: These aren't used in the prompt template, so they don't really belong here. I'd move them into the tasks module instead

src/euroeval/tasks.py
```diff
@@ -169,6 +170,20 @@
 )
 
 
+TOOL_CALLING = Task(
+    name="tool-calling",
+    task_group=TaskGroup.TEXT_TO_TEXT,
+    template_dict=TOOL_CALLING_TEMPLATES,
+    metrics=[m.tool_calling_metric],
+    default_num_few_shot_examples=0,
+    default_max_generated_tokens=500,
```
saattrupdan, line 179: Why this many? Do we expect long answers? Keep in mind that small models _will_ often get into an infinite loop, using all the max tokens, taking quite long to evaluate overall if this number is set quite high.

src/euroeval/prompt_templates/tool_calling.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class ToolCall(pydantic.BaseModel):
+    """For structured generation purposes in tool-calling."""
+
+    function: str
+    arguments: dict[str, str]
+
+
+class ToolCallingResponse(pydantic.RootModel[list[ToolCall]]):
```
saattrupdan: Have you tested whether this works for proprietary models?

src/euroeval/prompt_templates/tool_calling.py
```diff
@@ -0,0 +1,53 @@
...
+    """For structured generation purposes in tool-calling."""
+
+    root: list[ToolCall]
+
+
+TOOL_CALLING_FUNCTION_KEY = "function"
+TOOL_CALLING_ARGUMENTS_KEY = "arguments"
+TOOL_CALLING_KEYS = [TOOL_CALLING_FUNCTION_KEY, TOOL_CALLING_ARGUMENTS_KEY]
```
saattrupdan: It feels a bit off to have these constants and not use them in the Pydantic classes, which kind of loses the point in them, that there are still multiple points that need changing if they're ever changed. Maybe you could define the `ToolCall` using the [create_model function](https://docs.pydantic.dev/latest/concepts/models/#dynamic-model-creation) instead, and use these constants there?

src/euroeval/prompt_templates/tool_calling.py
```diff
@@ -0,0 +1,53 @@
...
+    "and a user question is given below: \n"
+    "{text}\n"
+    "Answer the user question with a list of function call(s) to execute "
+    "to fulfill the users request. "
+    "The answer MUST be a valid JSON deserializable to a list of dicts, "
+    "using double quotes for all keys "
+    "and strings, and must not include any additional explanatory text. "
+    "Each dict in the list should have exactly the keys "
+    f'"{TOOL_CALLING_FUNCTION_KEY}" (the function name) '
+    f'and "{TOOL_CALLING_ARGUMENTS_KEY}" '
```
saattrupdan: [nit] You can use the !r formatting specifier to show the `__repr__` of a given object, which for strings amount to adding quotes around them:
```suggestion
    f'{TOOL_CALLING_FUNCTION_KEY!r} (the function name) '
    f'and {TOOL_CALLING_ARGUMENTS_KEY!r} '
```

src/euroeval/prompt_templates/tool_calling.py
```diff
@@ -0,0 +1,53 @@
...
+    "(a dict with the keywords and values describing the parameters)."
+)
+
+TOOL_CALLING_TEMPLATES: dict["Language", PromptConfig] = {
+    ENGLISH: PromptConfig(
+        default_prompt_prefix="",
+        default_prompt_template=_DEFAULT_TOOL_CALLING_PROMPT_TEMPLATE_EN,
```
saattrupdan: This is only needed for base decoders, which is not applicable to tool calling, so I would just leave it blank:

```suggestion
        default_prompt_template="",
```

src/euroeval/prompt_templates/tool_calling.py
```diff
@@ -0,0 +1,53 @@
...
+        default_instruction_prompt=_DEFAULT_TOOL_CALLING_PROMPT_TEMPLATE_EN,
```
saattrupdan: Instead of using the variable, I would include the prompt directly here, as it makes it easier when you scale this up to 30+ languages, which we should also do.

src/euroeval/prompt_templates/tool_calling.py
saattrupdan: I'd move these to `constants`, to keep things consistent.

src/euroeval/metrics/tool_calling.py
```diff
@@ -0,0 +1,124 @@
...
+
+if t.TYPE_CHECKING:
+    from datasets.arrow_dataset import Dataset
+
+    from ..data_models import BenchmarkConfig, DatasetConfig
+
+from ..metrics.base import Metric
+
+
+def _evaluate_function_toolcall_response(
```
saattrupdan, line 64: Why make the function hidden (with the prefix underscore)? Nothing wrong with it, just curious.

src/euroeval/metrics/tool_calling.py
```diff
@@ -0,0 +1,124 @@
...
+            if required_args and key not in required_args:
+                continue
+            # every predicted argument should be in the list of expected values
+            if key not in pred_args or pred_args[key] not in values:
+                return False
+    return True
+
+
+class ToolCallingMetric(Metric):
+    """Metric for tool calling."""
```
saattrupdan: I would give this a proper name. You call it 'tool calling accuracy', so I would call this class (and also the metric variable below) that.

src/scripts/create_bfcl.py
```diff
@@ -0,0 +1,120 @@
...
+        item["target_text"] = json.dumps(item.pop("ground_truth"))
+
+    load_dotenv()
+    dataset_dict = _split_dataset_to_dict(items)
+    dataset_dict.push_to_hub(
+        "EuroEval/bfcl-en", private=True, token=os.getenv("HF_TOKEN")
+    )
+
+
+def _split_dataset_to_dict(items: list[dict]) -> DatasetDict:
```
saattrupdan, line 62: Why do all the helper functions have hidden names?

src/scripts/create_bfcl.py
```diff
@@ -0,0 +1,120 @@
...
+    random.seed(42)
+    shuffled_items = items.copy()
+    random.shuffle(shuffled_items)
+
+    train_size = 250
+    val_size = 250
+
+    train_items = shuffled_items[:train_size]
+    val_items = shuffled_items[train_size : train_size + val_size]
+    test_items = shuffled_items[train_size + val_size :]
```
saattrupdan, line 82: How big is the test set?

src/euroeval/benchmark_modules/litellm.py
```diff
@@ -949,8 +950,11 @@ def _create_model_output(
             # In the case where we're dealing with a classification task, the model is
             # outputting a JSON dictionary, so we will extract the generated text from
             # within the dictionary
+            # This is not relevant for tooling and may cause problems there
             generation_dct: dict[str, t.Any] | None = None
-            if LITELLM_CLASSIFICATION_OUTPUT_KEY in generation_output:
+            if LITELLM_CLASSIFICATION_OUTPUT_KEY in generation_output and not all(
+                [k in generation_output for k in TOOL_CALLING_KEYS]
+            ):
```
saattrupdan, line 952: But if the generation dictionary is not a dict with only the `LITELLM_CLASSIFICATION_OUTPUT_KEY` key, then nothing happens - do you foresee this ever being an issue here? Just trying to avoid having too many checks that aren't needed, for readability.

src/euroeval/benchmark_modules/litellm.py
```diff
@@ -1893,8 +1907,10 @@ def clean_model_id(model_id: str, benchmark_config: BenchmarkConfig) -> str:
     # inference endpoints, LiteLLM gets confused since it's already using the `openai/`
     # prefix. We thus have to add it twice, and this hack here is to ensure that we
     # don't store the results with model ID `openai/openai/...`.
-    elif benchmark_config.api_base is not None and model_id.startswith("openai/"):
-        model_id = "openai/openai/" + re.sub(r"(openai/)*", "", model_id)
+    # This is where the "openai/openai/..." bug arose. I'm guessing we keep it,
+    # though it doesn't seem to work for me when using VLLM locally for inference.
+    # elif benchmark_config.api_base is not None and model_id.startswith("openai/"):
+    #     model_id = "openai/openai/" + re.sub(r"(openai/)*", "", model_id)
```
saattrupdan: Since you opened a [separate issue on this](https://github.com/EuroEval/EuroEval/issues/1557), I would leave it here for now (i.e., revert back the changes in this PR).

src/euroeval/benchmark_modules/vllm.py
```diff
@@ -487,21 +487,29 @@ def generate(self, inputs: dict) -> "GenerativeModelOutput":
...
-            )
-            structured_outputs = StructuredOutputsParams(
-                json=structured_generation_schema
-            )
+            if self.dataset_config.task.structured_output_format is not None:
+                structured_outputs = StructuredOutputsParams(
+                    json=self.dataset_config.task.structured_output_format.model_json_schema(),
+                    disable_fallback=True,
+                )
+            else:
```
saattrupdan: We should change this check to explicitly checking for the NER task, and have an `else` clause that raises an error

src/euroeval/data_models.py
```diff
@@ -108,6 +108,12 @@ class Task:
         uses_structured_output (optional):
             Whether the task uses structured output. If True, the task will return
             structured output (e.g., BIO tags for NER). Defaults to False.
+        structured_output_format (optional):
+            Schema to use for structured output as a (deserializable) JSON string.
+            Only used if `uses_structured_output` is True.
```
saattrupdan, line 119: Maybe we could include a line in the `__post_init__` below which sets `structured_output_format` to None if `uses_structured_output` is False, and log a warning?

src/euroeval/dataset_configs/english.py
```diff
@@ -13,10 +13,19 @@
...
 # Official datasets ###
 
+BFCL_CONFIG = DatasetConfig(
+    name="bfcl",
+    pretty_name="BFCL-v4",
+    source="EuroEval/bfcl-en",
+    task=TOOL_CALLING,
+    languages=[ENGLISH],
+)
+
```
saattrupdan: [nit] Can you move this down after IFEval? Then it groups up nicely with the other dataset configs.

src/euroeval/metrics/tool_calling.py
```diff
@@ -0,0 +1,124 @@
...
+            # we only care about required arguments
+            if required_args and key not in required_args:
+                continue
+            # every predicted argument should be in the list of expected values
+            if key not in pred_args or pred_args[key] not in values:
+                return False
+    return True
+
+
+class ToolCallingMetric(Metric):
```
saattrupdan: I'd move this up to the top of the module, and the helper functions below. Helps with readability.

src/euroeval/metrics/tool_calling.py
```diff
@@ -0,0 +1,124 @@
...
+    def __call__(
+        self,
+        predictions: c.Sequence,
+        references: c.Sequence,
+        dataset: "Dataset",
+        dataset_config: "DatasetConfig",
+        benchmark_config: "BenchmarkConfig",
+    ) -> float | None:
+        """Calculate tool calling metric.
+
```
saattrupdan: Missing Args section in the docstring

src/euroeval/metrics/tool_calling.py
```diff
@@ -0,0 +1,124 @@
...
+        for x in zip(predictions, references, function_descriptions):
+            results.append(_evaluate_function_toolcall_response(*x))
+        if not results:
+            return None
+        else:
+            return sum(results) / len(results)
+
+
+tool_calling_metric = ToolCallingMetric(
+    name="tool_calling_accuracy", pretty_name="Accuracy (Tool Calling)"
```
saattrupdan: Consistency

```suggestion
    name="tool_calling_accuracy", pretty_name="Tool Calling Accuracy"
```

CHANGELOG.md
harderj: I didn't type them myself, I assumed they where made by the markdown precommit functionality. Will look into it

harderj: doesn't seem to my markdownlint (what could it be then?) - i'll try to revert then

docs/datasets/english.md
harderj: whoops sorry for that one

harderj:
Ready for review again @saattrupdan (now actually) :crossed_fingers:

docs/datasets/english.md
saattrupdan: @harderj Missing this?

docs/datasets/english.md
saattrupdan: @harderj Missing this?

docs/datasets/english.md
saattrupdan: @harderj Missing this?

docs/tasks/tool-calling.md
saattrupdan: @harderj Missing this?

src/euroeval/benchmark_modules/vllm.py
saattrupdan: @harderj Missing this?

src/euroeval/metrics/tool_calling.py
```diff
@@ -0,0 +1,138 @@
...
+if t.TYPE_CHECKING:
+    from datasets.arrow_dataset import Dataset
+
+    from ..data_models import BenchmarkConfig, DatasetConfig
+
+from ..metrics.base import Metric
+
+
+class ToolCallingMetric(Metric):
+    """Metric for tool calling."""
```
saattrupdan: Same comment as before (it seems to have disappeared): Can we change the name of this metric to its actual name, tool calling accuracy?

src/euroeval/metrics/tool_calling.py
```diff
@@ -0,0 +1,138 @@
...
+
+    def __call__(
+        self,
+        predictions: c.Sequence,
+        references: c.Sequence,
+        dataset: "Dataset",
+        dataset_config: "DatasetConfig",
+        benchmark_config: "BenchmarkConfig",
+    ) -> float | None:
+        """Calculate tool calling metric.
```
saattrupdan:
```suggestion
        """Calculate tool calling accuracy.
```

src/euroeval/metrics/tool_calling.py
saattrupdan: @harderj Missing this?

src/euroeval/prompt_templates/tool_calling.py
saattrupdan: @harderj Missing this?

src/euroeval/prompt_templates/tool_calling.py
saattrupdan: @harderj Missing this?

src/euroeval/prompt_templates/tool_calling.py
saattrupdan: @harderj Missing this?

saattrupdan: @harderj Missing this?

src/euroeval/prompt_templates/tool_calling.py
saattrupdan: @harderj Missing this?

src/euroeval/prompt_templates/tool_calling.py
saattrupdan: @harderj Missing this?

src/euroeval/prompt_templates/tool_calling.py
saattrupdan: @harderj Missing this?

src/euroeval/tasks.py
saattrupdan: @harderj Missing this?

src/scripts/create_bfcl.py
saattrupdan: @harderj Missing this?

src/scripts/create_bfcl.py
saattrupdan: @harderj Missing this?

docs/datasets/english.md
harderj: can't seem to find DOI, but added link :+1:

docs/datasets/english.md
harderj: no

harderj: full list is available here: https://github.com/ShishirPatil/gorilla/tree/main/berkeley-function-call-leaderboard/bfcl_eval/data

docs/datasets/english.md
harderj: added this extra info

docs/datasets/english.md
harderj: The rest of the data requires more complicated evaluation, so I decided to postpone it.

docs/tasks/tool-calling.md
harderj: "exactly matches one of the options given in the ground truth"? I'm not sure how to make it clearer. Also the ground truth example below contains such multiple correct answers

src/euroeval/tasks.py
harderj: As I pointed out the last thread about this - yes we can expect >300 token answers here in some cases. I think 500 is a reasonable upper bound

src/euroeval/prompt_templates/tool_calling.py
harderj: seems to work, yeah. just tested with `gpt-5-mini`

src/euroeval/prompt_templates/tool_calling.py
harderj: I changed it to use `create_model` and the constants

src/euroeval/metrics/tool_calling.py
harderj: it is used to indicate that it is "private" - not used outside the source file. to enforce it of course one would need to e.g. fiddle with the `__init__.py` file, but i think it is overkill in this case

src/scripts/create_bfcl.py
harderj: they are not used outside the source file. in this case it's a script so it does not matter either way

src/scripts/create_bfcl.py
harderj: added a comment

src/euroeval/benchmark_modules/litellm.py
harderj: I don't think this will make the code significantly more error-prone, and it fixes the problem with the tool-calling task without any likely downside for classification tasks as far as i can see. we could talk about relying on something more robust than substring/key checking, but I think that that would belong in another pull-request

src/euroeval/benchmark_modules/litellm.py
harderj: ah yes sorry

docs/tasks/tool-calling.md
saattrupdan: Is it that there are multiple correct tool calls to use for a given sample? Or is it that there's one correct tool but different arguments? Or is it that it's also allowed to not use a tool call for a sample? And a quick example of multiple ground truths (in short layman text) would be great.

src/euroeval/benchmark_modules/litellm.py
saattrupdan: I guess my question here is whether it's possible for this to _ever_ happen? If so, then I'm completely on board with it, but otherwise it's a code block that'll never be used?

@harderj replying: (I'm not sure why this doesn't have a comment section)
In that case the answer is _yes_ it can happen

src/euroeval/benchmark_modules/vllm.py
```diff
@@ -20,6 +20,8 @@
 from transformers.models.auto.tokenization_auto import AutoTokenizer
 from urllib3.exceptions import RequestError
 
+from euroeval.tasks import NER
```
saattrupdan: For consistency, we use relative imports.

```suggestion
from ..tasks import NER
```

src/euroeval/benchmark_modules/vllm.py
```diff
@@ -487,21 +493,37 @@ def generate(self, inputs: dict) -> "GenerativeModelOutput":
...
-                f"{json.dumps(structured_generation_schema, ensure_ascii=False)}",
-                level=logging.DEBUG,
-            )
-            structured_outputs = StructuredOutputsParams(
-                json=structured_generation_schema
-            )
+            if self.dataset_config.task.structured_output_format is not None:
+                structured_outputs = StructuredOutputsParams(
+                    json=self.dataset_config.task.structured_output_format.model_json_schema(),
+                    disable_fallback=True,
```
saattrupdan: What does this do?

src/euroeval/prompt_templates/tool_calling.py
```diff
@@ -0,0 +1,33 @@
+"""Templates for the tool calling task."""
+
+import typing as t
+
+from euroeval.constants import TOOL_CALLING_ARGUMENTS_KEY, TOOL_CALLING_FUNCTION_KEY
```
saattrupdan: Relative imports, for consistency:

```suggestion
from ..constants import TOOL_CALLING_ARGUMENTS_KEY, TOOL_CALLING_FUNCTION_KEY
```

src/euroeval/tasks.py
```diff
@@ -169,6 +176,37 @@
...
+        TOOL_CALLING_ARGUMENTS_KEY: (dict[str, str], ...),
+    },
+)
+
+
+class ToolCallingResponse(pydantic.RootModel[list[ToolCall]]):
+    """For structured generation purposes in tool-calling."""
+
+    # root: list[ToolCall]
+    pass
```
saattrupdan: I'm not sure what's going on here, does it do anything if `root` is not set - is this because it's included in the `RootModel`? And do we know if proprietary vendors can handle this, as they're normally restricted to structured generation with JSON dicts?

src/euroeval/tasks.py
saattrupdan: Gotcha, sorry for not remembering!

src/scripts/create_bfcl.py
saattrupdan: Great, thanks

saattrupdan (changes requested):
Tiny edits left now. I'll also hop onto your branch to double check that the evaluation works. Otherwise we're good to go!

src/euroeval/tasks.py
saattrupdan: I just tried it with `gpt-4o-mini`, and I _did_ get the error that I was fearing:

```bash
Failed to generate text. The error message was: litellm.BadRequestError: OpenAIException - Invalid schema for response_format 'ToolCallingResponse': schema must be a JSON Schema of 'type: "object"', got 'type: "array"'.. Received Model Group=gpt-4o-mini
Available Model Group Fallbacks=None
```

You could change it to a regular dict instead, with a single key `tool_calls`, being the list of tool calls.

saattrupdan:
@harderj I tried running it with an open-weight model locally, and that worked great, so it's just making it work for proprietary models now (the list vs dict problem as described above).

Also, looking at the different BFCL versions, our split here corresponds to BFCL-v2, so we should use that terminology here (in docs, variable names and dataset names), to avoid confusion.

src/euroeval/tasks.py
harderj: Yes included in the RootModel.

aha I see - I guess the only option is to wrap the list in a dict then. I'm on it

harderj: should be fixed now

src/euroeval/benchmark_modules/vllm.py
harderj: as I understand it raises an error on failure to use the format. Which we probably don't want actually - so I'll remove it - thanks for the catch

CHANGELOG.md
```diff
@@ -7,8 +7,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+## [v16.16.1] - 2026-02-25
+
 ### Added
 
+- A task _Tool Calling_ and a dataset under this task _bfcl_ a subset of the
+Berkeley Function Calling Leaderboard benchmark (v4).
+Currently only supported for English.
```
saattrupdan: Can you move this up to the Unreleased section?

docs/datasets/english.md
```diff
@@ -1246,3 +1246,139 @@ You can evaluate this dataset directly as follows:
 ```bash
 euroeval --model <model-id> --dataset valeu-en
 ```
+
+## Tool calling
+
+### BFCL
```
saattrupdan:
```suggestion
### BFCL-v2
```

docs/datasets/english.md
```diff
@@ -1246,3 +1246,139 @@ You can evaluate this dataset directly as follows:
...
+            }
+        }
+    ]
+}
+```
+
+You can evaluate this dataset directly as follows:
+
+```bash
+euroeval --model <model-id> --dataset bfcl
```
saattrupdan:
```suggestion
euroeval --model <model-id> --dataset bfcl-v2
```

src/euroeval/constants.py
```diff
@@ -139,3 +139,9 @@
 # the languages. This is for instance the case for the speed benchmark, which has all
 # the languages. The threshold of 5 is somewhat arbitrary.
 MAX_NUMBER_OF_LOGGING_LANGUAGES = 5
+
+# Constants for the tool-calling task
+TOOL_CALLING_CALLS_KEY = "tool_calls"
+TOOL_CALLING_FUNCTION_KEY = "function"
+TOOL_CALLING_ARGUMENTS_KEY = "arguments"
+TOOL_CALLING_KEYS = [TOOL_CALLING_FUNCTION_KEY, TOOL_CALLING_ARGUMENTS_KEY]
```
saattrupdan, line 183: Is this still relevant, when we have the 'tool_calls' key now?

src/euroeval/dataset_configs/english.py
```diff
@@ -83,6 +84,14 @@
     val_split=None,
 )
 
+BFCL_CONFIG = DatasetConfig(
+    name="bfcl",
+    pretty_name="BFCL-v2",
+    source="EuroEval/bfcl-en",
+    task=TOOL_CALLING,
+    languages=[ENGLISH],
+)
```
saattrupdan:
```suggestion
BFCL_V2_CONFIG = DatasetConfig(
    name="bfcl-v2",
    pretty_name="BFCL-v2",
    source="EuroEval/bfcl-v2",
    task=TOOL_CALLING,
    languages=[ENGLISH],
)
```

src/euroeval/prompt_templates/tool_calling.py
```diff
@@ -0,0 +1,57 @@
...
+    __base__=pydantic.BaseModel,
+    **{TOOL_CALLING_FUNCTION_KEY: str, TOOL_CALLING_ARGUMENTS_KEY: dict[str, str]},
+)
+
+
+ToolCallingResponse = pydantic.create_model(
+    "ToolCallingResponse",
+    __base__=pydantic.BaseModel,
+    **{TOOL_CALLING_CALLS_KEY: list[ToolCall]},
+)
```
saattrupdan, line 29: Should be moved to `tasks` module?

src/euroeval/tasks.py
```diff
@@ -169,6 +170,20 @@
 )
 
 
+TOOL_CALLING = Task(
```
saattrupdan, line 173: This task needs the following arguments set:

```
    default_allowed_generative_types=[
        GenerativeType.INSTRUCTION_TUNED,
        GenerativeType.REASONING,
    ],
```

src/scripts/create_bfcl.py
```diff
@@ -0,0 +1,121 @@
+"""Create the bfcl datasets and upload them to the HF Hub."""
```
saattrupdan:
```suggestion
"""Create the BFCL-v2 dataset and upload it to the HF Hub."""
```

src/scripts/create_bfcl.py
```diff
@@ -0,0 +1,121 @@
...
+import random
+from pathlib import Path
+from urllib.request import urlopen
+
+from datasets import Dataset, DatasetDict
+from dotenv import load_dotenv
+
+
+def main() -> None:
+    """Create BFCL dataset (the part used by EuroEval) and push to hugging face."""
```
saattrupdan:
```suggestion
    """Create BFCL-v2 dataset (the part used by EuroEval) and push to hugging face."""
```

src/scripts/create_bfcl.py
```diff
@@ -0,0 +1,121 @@
...
+        item: dict
+        item["text"] = f"Functions:\n{function_str}\nQuestion: {question_str}"
+        item["function"] = function_str
+        item["question"] = question_str
+        item["target_text"] = json.dumps(item.pop("ground_truth"))
+
+    load_dotenv()
+    dataset_dict = _split_dataset_to_dict(items)
+    dataset_dict.push_to_hub(
+        "EuroEval/bfcl-en", private=True, token=os.getenv("HF_TOKEN")
```
saattrupdan:
```suggestion
        "EuroEval/bfcl-v2", private=True, token=os.getenv("HF_TOKEN")
```

CHANGELOG.md
```diff
@@ -7,8 +7,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+## [v16.16.1] - 2026-02-25
+
 ### Added
 
+- A task _Tool Calling_ and a dataset under this task _bfcl_ a subset of the
```
saattrupdan:
```suggestion
- A task _Tool Calling_ and a dataset under this task _bfcl-v2_ a subset of the
```

src/euroeval/metrics/tool_calling.py
```diff
@@ -0,0 +1,140 @@
+"""Tool calling metric."""
+
+import collections.abc as c
+import json
+import typing as t
+
+from euroeval.logging_utils import log_once
```
saattrupdan:
```suggestion
from ..logging_utils import log_once
```

src/euroeval/prompt_templates/tool_calling.py
harderj: That would result in circular import as it is used below. Would have to be in it's own module - but couldn't tell where in that case, maybe you have a suggestion @saattrupdan

src/euroeval/constants.py
harderj: yes, as it is used in `src/euroeval/benchmark_modules/litellm.py:951` @saattrupdan

src/euroeval/prompt_templates/tool_calling.py
saattrupdan: Let's keep it for now

PR #1555, opened 2026-02-19
Feat/use trust remote code flag when loading custom datasets

### LINKED ISSUE #1529: [FEATURE REQUEST] Adding `trust_remote_code` flag

With the new feature to evaluate remote datasets with euroeval it would be great to have a `trust_remote_code` flag to avoid loading unsafe code

saattrupdan:
### Added

- When evaluating community evaluation datasets from the Hugging Face Hub, we now
  require that the user actively set `--trust-remote-code` (or `trust_remote_code=True`
  if running with the `Benchmarker` class). This is to prevent accidental execution of
  malicious code.

Fixes #1529

### DIFF
--- a/docs/python-package.md
+++ b/docs/python-package.md
@@ -306,10 +306,17 @@ CONFIG = DatasetConfig(
 )
 ```
 
+!!! note
+
+    To benchmark a dataset from the Hugging Face Hub, you always need to set the
+    `--trust-remote-code` flag (or `trust_remote_code=True` if using the `Benchmarker`),
+    as the dataset configuration is loaded from the remote code. We advise you to always
+    look at the code of the dataset configuration before running the benchmark.
+
 You can then benchmark your custom dataset by simply running
 
 ```bash
-euroeval --dataset <org-id>/<repo-id> --model <model-id>
+euroeval --dataset <org-id>/<repo-id> --model <model-id> --trust-remote-code
 ```
 
 You can also run the benchmark from a Python script, by simply providing your repo ID to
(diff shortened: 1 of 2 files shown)


src/euroeval/custom_dataset_configs.py
```diff
@@ -59,6 +64,11 @@ def try_get_dataset_config_from_repo(
             dataset configs.
         cache_dir:
             The directory to store the cache in.
+        trust_remote_code:
+            Whether to trust remote code. If this is not set to True, then we will not
+            load the dataset config.
+        run_with_cli:
+            Whether the code is being run with the CLI.
```
KennethEnevoldsen, line 71: Why do you need the run with the CLI?

src/euroeval/custom_dataset_configs.py
```diff
@@ -81,6 +91,24 @@ def try_get_dataset_config_from_repo(
...
+    # user has allowed running code from remote repositories, and abort if not. We abort
+    # the entire evaluation here to avoid a double error message, and since it requires
+    # the user to explicitly allow it before continuing.
+    if not trust_remote_code:
+        rerunning_msg = (
+            "the --trust-remote-code flag"
+            if run_with_cli
+            else "`trust_remote_code=True`"
+        )
+        log_once(
```
KennethEnevoldsen, line 104: is it all for this message? (might be worth just rephrasing to avoid passing that argument throughout?)

saattrupdan: It's all for the message yep. That's all that argument is ever used for in the whole codebase 😛 

I mean I could rephrase it, but maybe it's quite nice to have appropriate logs in your concrete use case - or do you think it's overdoing it?

src/euroeval/custom_dataset_configs.py
saattrupdan: See below :)

saattrupdan:
@KennethEnevoldsen Updated the docs with the trust remote code flag mention now.

src/euroeval/custom_dataset_configs.py
KennethEnevoldsen: I think it is up to you so probably just leave it

docs/python-package.md
KennethEnevoldsen, line 1: Looks good!

PR #1556, opened 2026-02-19
Feat/debug model cache

### LINKED ISSUE #1530: [FEATURE REQUEST] Include ground truth labels and metadata in debug logging

Currently, we only log model input and output in the JSON cache when running in debugging mode (via `--debug` or `debug=True`). It has been requested that we also log the ground truth labels (if available) and all metadata columns from the dataset as well, especially since we're logging them in the terminal anyway.

We should also add better documentation on how to use these.

saattrupdan:
### Added

- We now add all metadata (including ground truth labels, if applicable) to the model cache when debug mode is enabled (with `--debug` or `debug=True`). We have added a [section in the documentation](https://euroeval.com/python-package/#analysing-the-results-of-generative-models) on how to use this feature.

Fixes #1530

Docs result:

Screenshot 2026-02-23 at 09 51 53
Screenshot 2026-02-23 at 09 51 59

### DIFF
--- a/docs/python-package.md
+++ b/docs/python-package.md
@@ -134,7 +134,7 @@ docker build --pull -t euroeval .
 With the Docker image built, we can now evaluate any model as follows:
 
 ```bash
-docker run -e args="<euroeval-arguments>" --gpus 1 --name euroeval --rm euroeval
+docker run --rm --gpus 1 euroeval <euroeval-arguments>
 ```
 
 Here `<euroeval-arguments>` consists of the arguments added to the `euroeval` CLI


docs/python-package.md
```diff
@@ -410,3 +410,25 @@ Again, with this you can benchmark your custom dataset by simply running
...
+## Analysing the results
+
+If you're evaluating a generative model and want to be able to analyse the model results
+more in-depth, you can run your evaluation with the `--debug` flag (or `debug=True` if
+using the `Benchmarker`), which will output all the model outputs and all the dataset
+metadata (including the ground truth labels, if present) to both the terminal as well as
+to a JSON file in your current working directory, named
+`<model-id>-<dataset-name>-model-outputs.json`.
+
+It is a JSON dictionary with keys being hashes of the input, and values being
```
KennethEnevoldsen: How can I match these up with the original data?

We could e.g. do an example with dala since it is public - or alternative upload a mini dataset to euroeval to illustrate

docs/python-package.md
```diff
@@ -410,3 +410,25 @@ Again, with this you can benchmark your custom dataset by simply running
...
+dictionaries with the following keys:
```
KennethEnevoldsen: I would love to just see an example here (even if it is in a fold down menu)

docs/python-package.md
saattrupdan: If the dataset has an ID column then it's quite easy of course, but I guess you mean if not. Maybe we could always add a `index` column to the datasets when loading them? Then there will always be that as ID at least.

docs/python-package.md
saattrupdan: Yeah thought about it too but felt it got clunky because of its large size. But a dropdown is a great idea!

saattrupdan: I've added an example now in [this commit](https://github.com/EuroEval/EuroEval/pull/1556/commits/5250e09587d8aa57e9033d0e888298a81dd61994), updated slightly in [this commit](https://github.com/EuroEval/EuroEval/pull/1556/commits/6299d5c544321a16c15aad57d1861c520ea5bb5e).

docs/python-package.md
saattrupdan: Added the extra column now in [this commit](https://github.com/EuroEval/EuroEval/pull/1556/commits/72140b951cae55991315b41b4b0fc7f1e734ae20)

docs/python-package.md
```diff
@@ -410,3 +410,184 @@ Again, with this you can benchmark your custom dataset by simply running
...
+              -100.0
+            ]
+          ]
+        ],
+        "corruption_type": null,
+        "label": "correct",
+        "index": 181,
+        "messages": [
+          {
+            "content": "S\u00e6tning: Styrkeforholdet m\u00e5 v\u00e6re det afg\u00f8rene, siger de begge.\n\nBestem om s\u00e6tningen er grammatisk korrekt eller ej. Svar kun med 'ja' eller 'nej', og intet andet.",
```
KennethEnevoldsen: formatting here seems a bit odd I would probably fix it and reduce to down to e.g. 4 messages.

docs/python-package.md
```diff
@@ -410,3 +410,184 @@ Again, with this you can benchmark your custom dataset by simply running
...
+            [
+              "!",
+              -100.0
+            ]
+          ]
+        ],
+        "corruption_type": null,
+        "label": "correct",
+        "index": 181,
+        "messages": [
```
KennethEnevoldsen, line 531: ~~This kinda looks like one long conversation - I don't suspect that is true?~~

ahh these are the few show example! Maybe shorten those so we actually see the

docs/python-package.md
```diff
@@ -410,3 +410,184 @@ Again, with this you can benchmark your custom dataset by simply running
...
+- Any metadata for the sample that was present in the dataset, including the ground truth
+  label, if present.
+
+??? example
+
+    Here is a (truncated) example of a model output file:
+
+    ```json
+    {
+      "cb3f9ea749fec9d2f83ca6d3a8744cce": {
```
KennethEnevoldsen, line 489: is this the sample id?

docs/python-package.md
```diff
@@ -410,3 +410,184 @@ Again, with this you can benchmark your custom dataset by simply running
...
+              -100.0
+            ],
+            [
+              "#",
+              -100.0
+            ],
+            [
+              "!",
+              -100.0
+            ]
```
KennethEnevoldsen, line 526: why are these here, seems like they are invalid

docs/python-package.md
KennethEnevoldsen: (resolved, seems like I can't resolve it)

docs/python-package.md
KennethEnevoldsen: Hmm I am just not sure how I would go about mapping `cb3f9ea749fec9d2f83ca6d3a8744cce` from below to the actual row in my data. Seems like that would be a common usecase

saattrupdan: `cb3f9ea749fec9d2f83ca6d3a8744cce` is just a hash ID and is irrelevant to further analysis. The `index` key would be used to map to the actual row.

docs/python-package.md
saattrupdan: You mean just use (...) to truncate the few-shot examples?

Ah yeah, I think I forgot to include the `ensure_ascii=False`. Annoying that it defaults to that.

docs/python-package.md
saattrupdan: It's the hash, see [this comment](https://github.com/EuroEval/EuroEval/pull/1556/#discussion_r2833727056)

docs/python-package.md
saattrupdan: I think that''s the structured generation preventing these from being sampled - it's coming directly from LiteLLM.

docs/python-package.md
saattrupdan: I've added more info to the docs section now on the `index` key, as well as making it the first key of each sample in the model output, for easier readability.

docs/python-package.md
saattrupdan: Fixed now

docs/python-package.md
KennethEnevoldsen: That was my suspicion as well. Seems very much like a debug log.

broader consideration (that I don't think we should address here). Should we allow  export to the instance level as specified in every eval ever?

docs/python-package.md
KennethEnevoldsen: Right - I would still really love to see the simple merge - the best way to show the user that this is what you should merge on

saattrupdan: I'm a bit confused. Firstly, there would be no need to merge at all, since the entire dataset will now lie in the cache? But they can sort by index to get the dataset exactly as it was initially. Are you thinking of some other use case here?

docs/python-package.md
saattrupdan: Yeah I don't see why not. Could be a flag set when evaluating to store it to disk.

docs/python-package.md
saattrupdan: I've added a little Python snippet now on how to recreate the dataset 🙂 

Here's a screenshot:

Screenshot 2026-02-21 at 12 05 18

KennethEnevoldsen: I suppose the sample is on there insofar as it's in the prompt - but that is probably good enough. 

I like the dataset conversion. I would suggest keeping that

KennethEnevoldsen (approved):
Alright I think this is at a point where I would be happy to see it merged - I would ask Gianluca as well for a review, since he might know the pain points a bit better

saattrupdan:
> Alright I think this is at a point where I would be happy to see it merged - I would ask Gianluca as well for a review, since he might know the pain points a bit better

Great stuff, and good idea.

I added an extra bit to the documentation, specifying the columns `text`/`messages`, being the full input to the model (including few-shot examples), and `prompt`, which only includes the present query (so the same but without the few-shot examples).

saattrupdan:
@N-essuno Can you have a look and see if this fits your use case as well? I can't formally request your review, for some reason.

N-essuno:
I had a look at the new docs, looks nice! 
I like the ID linking to dataset examples but in some debugging I used to do I personally found useful (and I hardcoded that locally) to also get directly everything I needed to look at in one run, without any additional post-processing. So I also included the clean dataset example (meaning each column, cause there might be some useful metadata in there. e.g. the corruption_type I use in DaLA, which is not used for classification but nice for error analysis) directly into separate fields so I didn't need to do any explicit linking or extracting the example from the prompt.
I would consider adding that if possible, or, another way to do it could be providing the post-processing function to link dataset examples with outputs as maybe would be more clean. 
I think leaving the post-processing to the user could be a bit annoying cause it's unclear where the datasets to be linked are and would still be needed some custom code.

I also noticed that the true label from the dataset is not provided (unless I missed something), I guess that would be needed regardless of the previous suggestion (which would also include this column of course).

saattrupdan:
> I had a look at the new docs, looks nice! I like the ID linking to dataset examples but in some debugging I used to do I personally found useful (and I hardcoded that locally) to also get directly everything I needed to look at in one run, without any additional post-processing. So I also included the clean dataset example (meaning each column, cause there might be some useful metadata in there. e.g. the corruption_type I use in DaLA, which is not used for classification but nice for error analysis) directly into separate fields so I didn't need to do any explicit linking or extracting the example from the prompt. I would consider adding that if possible, or, another way to do it could be providing the post-processing function to link dataset examples with outputs as maybe would be more clean. I think leaving the post-processing to the user could be a bit annoying cause it's unclear where the datasets to be linked are and would still be needed some custom code.
> 
> I also noticed that the true label from the dataset is not provided (unless I missed something), I guess that would be needed regardless of the previous suggestion (which would also include this column of course).

I'm a bit confused here. All dataset columns are included, including the dataset labels. I'm attaching a screenshot here on how it looks:

Screenshot 2026-02-23 at 09 51 53
Screenshot 2026-02-23 at 09 51 59

N-essuno:
> > I had a look at the new docs, looks nice! I like the ID linking to dataset examples but in some debugging I used to do I personally found useful (and I hardcoded that locally) to also get directly everything I needed to look at in one run, without any additional post-processing. So I also included the clean dataset example (meaning each column, cause there might be some useful metadata in there. e.g. the corruption_type I use in DaLA, which is not used for classification but nice for error analysis) directly into separate fields so I didn't need to do any explicit linking or extracting the example from the prompt. I would consider adding that if possible, or, another way to do it could be providing the post-processing function to link dataset examples with outputs as maybe would be more clean. I think leaving the post-processing to the user could be a bit annoying cause it's unclear where the datasets to be linked are and would still be needed some custom code.
> > I also noticed that the true label from the dataset is not provided (unless I missed something), I guess that would be needed regardless of the previous suggestion (which would also include this column of course).
> 
> I'm a bit confused here. All dataset columns are included, including the dataset labels. I'm attaching a screenshot here on how it looks:
> 
> Screenshot 2026-02-23 at 09 51 53 Screenshot 2026-02-23 at 09 51 59

Ok, I'm not sure what I looked at but it was just a partial docs probably, now I see that everything is included so that's perfect!

Issue #1586, opened 2026-02-28
[BUG] Old OpenAI models crash on structured generation


Rijgersberg:
### 🐛 Describe the bug

Running the benchmark on older OpenAI models such as `gpt-3.5-turbo-1106` triggers a crash. This is caused by the fact that these models don't support the structured response `response_format`, but the fallback code doesn't catch this case.

On `main`:
```bash
$ euroeval --verbose --model gpt-3.5-turbo-1106 --dataset dbrd --api-key <api-key>
Dataset dbrd does not have a euroeval_config.py file, so we cannot load it. Skipping.
Started EuroEval run. Run with `--verbose` for more information.
Fetching model configurations:   0%|—————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————————| 0/1 [00:00<?, ?it/s][DEBUG] 2026-02-28 22:10:53 · The model 'gpt-3.5-turbo-1106' was identified by the LiteLLMModel benchmark module.
[DEBUG] 2026-02-28 22:10:53 · Using the model cache directory '.euroeval_cache/model_cache/gpt-3.5-turbo-1106' for the model 'gpt-3.5-turbo-1106'.
                                                                                                                                                                                                                           
Loading the model gpt-3.5-turbo-1106...
    ↳ The model gpt-3.5-turbo-1106 has an unknown number of parameters, an unknown vocabulary size, and an unknown maximum sequence length.
[DEBUG] 2026-02-28 22:10:53 · Detected generative type 'INSTRUCTION_TUNED' for model 'gpt-3.5-turbo-1106'
[DEBUG] 2026-02-28 22:10:53 · We will use logprobs with the model 'gpt-3.5-turbo-1106' since the dataset supports it and no tokeniser is available.

[2026-02-28 22:10:53]
Few-shot benchmarking gpt-3.5-turbo-1106 on the validation split of the truncated version of the Dutch sentiment classification dataset DBRD (1/1 benchmarks)...
Failed to generate text. The error message was: litellm.BadRequestError: OpenAIException - Invalid parameter: 'response_format' of type 'json_schema' is not supported with this model. Learn more about supported models at the Structured Outputs guide: https://platform.openai.com/docs/guides/structured-outputs. Received Model Group=gpt-3.5-turbo-1106                                                                                        
Available Model Group Fallbacks=None

Errored 1 benchmarks
```



### Operating System

MacOS

### Device

Other

### Python version

3.12.x

[...]

PR #1587, opened 2026-02-28
Fix graceful fallback from structured generation for older OpenAI models

Rijgersberg:
Fixed the bug described in https://github.com/EuroEval/EuroEval/issues/1586.
> Running the benchmark on older OpenAI models such as `gpt-3.5-turbo-1106` triggers a crash. This is caused by the fact that these models don't support the structured response `response_format`, but the fallback code doesn't catch this case.

During development I also found a small bug that sometimes crashes the loading of cached results of NER. That needed fixing too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Added
 
+- Added the new Danish linguistic acceptability dataset DaLA. It's marked as
+  unofficial for now. This was added by @N-essuno ✨
+- Added the new Natural Language Inference task type and the Danish Entailment Dataset
+  [`danish-entailment`](https://github.com/kuhumcst/danish-semantic-reasoning-benchmark/tree/main/entailment).
+  The split is given by 32 / 286 samples for train / test, respectively (no validation
+  split).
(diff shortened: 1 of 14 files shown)


Rijgersberg:
Mmh I see the CI code-check fails on a file I didn't touch...

saattrupdan:
> Mmh I see the CI code-check fails on a file I didn't touch...

Yeah don't mind that, it's being fixed in another PR 🙂

src/euroeval/generation.py
```diff
@@ -240,10 +240,19 @@ def generate_single_iteration(
             input_batch=cached_dataset[:], model_output=model_output
         )
         if model_output.predicted_labels is None:
-            model_output.predicted_labels = [
-                dataset_config.prompt_label_mapping.get(label, label).lower()
-                for label in extracted_labels
-            ]
+            if pred2extracted := dataset_config.prompt_label_mapping:
```
Rijgersberg, line 247: This fix mirrors the codepath already present in https://github.com/EuroEval/EuroEval/blob/99f41b7868c4bef30a20b9036c28fa8cb9ee04b7/src/euroeval/generation.py#L198-L210

saattrupdan:
@Rijgersberg Can you merge the upstream changes to `main` into this branch please?

PR #1600, opened 2026-03-03
Fix documentation: F1-score is the primary metric for reading comprehension, not EM

Rijgersberg:
The [documentation](https://euroeval.com/tasks/reading-comprehension/#metrics) states:

> The primary metric we use when evaluating the performance of a model on the reading comprehension task is the exact match (EM) score, which is the percentage of questions for which the model provides the exact answer.
> 
> We also report the [F1-score](https://en.wikipedia.org/wiki/F1_score) on a character-basis, which is more lenient than the EM score, as it allows for small differences in the answer.

Looking at the [leaderboard config](https://github.com/EuroEval/leaderboards/blob/5eea3206d6c17fa7d5bf70a1da598f0e86951b71/task_config.yaml#L16-L19) and the scores on the leaderboards, this is incorrect. In fact the F1-score is the primary metric.

This PR corrects the documentation.


saattrupdan:
@Rijgersberg  Looks great, nice catch! Can you pull the newest main branch upstream? Then we can get it merged in 🙂

Rijgersberg:
Sorry, I'm stumbling a bit with git rebase. I'll try this again.

PR #1601, opened 2026-03-03
Fix documentation: F1-score is the primary metric for reading comprehension, not EM

Rijgersberg:
The [documentation](https://euroeval.com/tasks/reading-comprehension/#metrics) states:

> The primary metric we use when evaluating the performance of a model on the reading comprehension task is the exact match (EM) score, which is the percentage of questions for which the model provides the exact answer.
> 
> We also report the [F1-score](https://en.wikipedia.org/wiki/F1_score) on a character-basis, which is more lenient than the EM score, as it allows for small differences in the answer.

Looking at the [leaderboard config](https://github.com/EuroEval/leaderboards/blob/5eea3206d6c17fa7d5bf70a1da598f0e86951b71/task_config.yaml#L16-L19) and the scores on the leaderboards, this is incorrect. In fact the F1-score is the primary metric.

This PR corrects the documentation. 

Sorry for the duplicate PR, I had some trouble with git rebase on the previous (closed) PR.

### DIFF
--- a/docs/datasets/albanian.md
+++ b/docs/datasets/albanian.md
@@ -384,12 +384,15 @@ euroeval --model <model-id> --dataset global-mmlu-lite-sq
 
 ### Unofficial: INCLUDE-sq
 
-This dataset is part of [INCLUDE](https://doi.org/10.48550/arXiv.2411.19799), a comprehensive
-knowledge- and reasoning-centric benchmark that evaluates multilingual LLMs across 44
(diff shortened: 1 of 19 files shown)


saattrupdan:
@Rijgersberg Can you pull the newest main (don't rebase, just merge) from upstream? Then we should be good to go. The markdown linting issues is fixed there as well.

saattrupdan (approved):
LGTM!

PR #1460, opened 2026-01-21
Add hallucination detection with LettuceDetect for Danish

FrejaThoresen:
### Added

- Evaluation benchmark for hallucination detection, reporting a hallucination rate
  (hallucinated_tokens/total_tokens).

Fixes #735

PR #1620, opened 2026-03-11
Feat/add hallucination detection

saattrupdan:
### Added

- Evaluation benchmark for hallucination detection, reporting a hallucination rate
  (`hallucinated_tokens/total_tokens`).

This fixes #1402 and is a 'fork' of the PR https://github.com/EuroEval/EuroEval/pull/1460.


saattrupdan:
@copilot deal with reviews and also implement the following:
- There is currently only a Danish dataset config and docs, but this should be extended to all the other languages in the library. For the examples in the docs, just use the same examples as in MultiWikiQA for that language
- Add tests for the new metric module
- In all tests using mocking, use the `pytest-mock` functionalities instead of unittest, if possible

saattrupdan:
@copilot Implement tests again, code has changed now

src/euroeval/metrics/token_hallucination_classifier.py
```diff
@@ -0,0 +1,185 @@
...
+
+    transformer_detector = detector.detector
+    tokenizer = transformer_detector.tokenizer
+    max_length = transformer_detector.max_length
+
+    predicted_texts = [p["prediction_text"] for p in predictions]
+
+    hallucinated_tokens = 0
+    total_tokens = 0
+    skipped_samples = 0
```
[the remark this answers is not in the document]
FrejaThoresen: if one prediction fails due to e.g. the context being longer than what the model admits, then the sample is skipped. Hence, it's not a must that all samples in the dataset must be evaluated.

src/euroeval/generation.py
```diff
@@ -297,9 +297,23 @@ def generate_single_iteration(
         )
         ground_truth = []
 
+    # `all_preds` and `ground_truth` are ordered as the non-cached examples followed
```
FrejaThoresen: @saattrupdan there was a problem with the order of the dataset. Can you check if this will affect other evaluations? Will clean up the comment afterwards.

FrejaThoresen:
Changed since March:
- Implementation is only for Danish
- The test dataset is now translated RAGTruth, instead of Multi-Wiki-Synthetic-QA
- Model and dataset have been moved to EuroEval's page on huggingface

src/euroeval/generation.py
FrejaThoresen: Fixed in update. The dataset is not ordered, therefore using ids instead of zip.

FrejaThoresen:
Changes since review in June:
- Add src/scripts/dataset_creation/create_ragtruth.py
- - with val split of 256 samples, and test of 2048 samples
- Remove option of default device
- Check the need for check of max input length is needed (conclusion: It is, and will skip the sample)
- Check if model exist before loadning
- Download model in download() method for offline running
- - Had to add some logic for finding the models ids, as the dataset_config (which is needed for model name) is not passed to the download method.
- Revise readme file
- Use EMPTY_TEMPLATE instead of RC
- Change default allowed types (only reasoning and instruction)
- Remove postprocessing function for ragtruth, as it is now in the create_ragtruth script instead

src/euroeval/dataset_configs/danish.py
```diff
@@ -105,6 +106,16 @@
     bootstrap_samples=False,
 )
 
+RAGTRUTH_DA_CONFIG = DatasetConfig(
+    name="ragtruth-da",
+    pretty_name="RAGTruth-da",
+    source="EuroEval/ragtruth-translated-hallucinations-da-mini",
+    task=HALLU,
+    languages=[DANISH],
+    instruction_prompt="{text}",
```
saattrupdan: This is not required, as it's the default with the empty prompt template.

```suggestion
```

src/euroeval/metrics/token_hallucination_classifier.py
```diff
@@ -0,0 +1,281 @@
...
+    models up front (e.g. for offline benchmarking), since the per-dataset language
+    is not known inside the metric's ``download`` method.
+
+    Returns:
+        The set of Hugging Face Hub repository IDs of all hallucination detection
+        models referenced by built-in dataset configurations.
+    """
+    # Imported here rather than at module level to avoid a circular import, since
+    # the dataset configurations import this metric module via the task registry.
+    from .. import dataset_configs  # noqa: PLC0415
```
saattrupdan: I would use the `get_all_dataset_configs` function here, improved from `..dataset_configs`

src/euroeval/metrics/token_hallucination_classifier.py
```diff
@@ -0,0 +1,281 @@
...
+    model: str,
+    device: Device,
+    cache_dir: str,
+    max_input_length: int = DEFAULT_MAX_INPUT_LENGTH,
+) -> float:
+    """Load model and detect hallucinations.
+
+    Args:
+        dataset:
+            Hallucination dataset, generated with e.g. lettuce. Each example must
```
saattrupdan: Nit: lettuce -> lettucedetect

src/euroeval/metrics/token_hallucination_classifier.py
```diff
@@ -0,0 +1,281 @@
...
+            provide an ``"id"`` field and a ``"context"`` field containing the full
+            RAG prompt.
+        predictions: Iterable of prediction objects, each containing an ``"id"``
```
saattrupdan: Nit: Missing newline after colon.

src/euroeval/metrics/token_hallucination_classifier.py
```diff
@@ -0,0 +1,281 @@
...
+    return model_ids
+
+
+def detect_hallucinations(
+    dataset: Dataset,
+    predictions: c.Iterable[dict[str, t.Any]],
+    model: str,
+    device: Device,
+    cache_dir: str,
+    max_input_length: int = DEFAULT_MAX_INPUT_LENGTH,
```
saattrupdan: No need to have this as an argument, since you're not changing it anywhere. Just use the constant as the value inside the function.

src/euroeval/metrics/token_hallucination_classifier.py
```diff
@@ -0,0 +1,281 @@
...
+from ..enums import Device
+from ..exceptions import InvalidBenchmark
+from .base import Metric
+
+logger = logging.getLogger(__name__)
+
+# The underlying mmBERT-small detector models support up to 8192 positions, but
+# lettucedetect's ``TransformerDetector`` defaults to a 4096-token budget. Use the
+# model's full capacity so fewer samples are skipped for exceeding the budget.
+DEFAULT_MAX_INPUT_LENGTH = 8192
```
saattrupdan: We probably want to reuse the already existing `MAX_CONTEXT_LENGTH` from the `constant` module here.

src/euroeval/metrics/token_hallucination_classifier.py
```diff
@@ -0,0 +1,281 @@
...
+    detector = HallucinationDetector(
+        method="transformer", model_path=model, device=device, cache_dir=cache_dir
+    )
+
+    transformer_detector = detector.detector
+    # ``HallucinationDetector`` does not forward ``max_length`` to the underlying
+    # transformer detector, so override it directly to use the configured budget.
+    transformer_detector.max_length = max_input_length
+    tokenizer = transformer_detector.tokenizer
+    max_length = transformer_detector.max_length
```
saattrupdan: A bit unsure what the `max_length` variable is. Isn't it always just the same as `max_input_length`?

src/euroeval/metrics/token_hallucination_classifier.py
```diff
@@ -0,0 +1,281 @@
...
+    Returns:
+        Whether the answer is too long to be evaluated alongside a prompt.
+    """
+    answer_token_count = len(tokenizer(answer, add_special_tokens=False)["input_ids"])
+    # Reserve room for special tokens ([CLS], two [SEP]) and at least one prompt
+    # token, matching the detector's ``truncation="only_first"`` requirement.
+    return answer_token_count >= max_length - 4
+
+
+class TokenHallucinationMetric(Metric):
```
saattrupdan: Nit: Can we move this up to the top of the module? Makes it a lot easier to follow what's going on, as it's the main starting point.

src/euroeval/metrics/token_hallucination_classifier.py
```diff
@@ -0,0 +1,281 @@
...
+    total_tokens = 0
+    skipped_samples = 0
+
+    for prediction in predictions:
+        prompt = id_to_context[prediction["id"]]
+        predicted_text = prediction["prediction_text"]
+
+        if _answer_too_long(
+            answer=predicted_text, tokenizer=tokenizer, max_length=max_length
+        ):
```
saattrupdan, line 255: Unsure why this is relevant, since we're enforcing the maximum length set in the task configuration. I think this is the block that you mentioned maybe weren't triggered ever. We should either remove it or add a "TODO: (...)" comment indicating that we need to figure out if it's needed at all.

src/euroeval/metrics/token_hallucination_classifier.py
```diff
@@ -0,0 +1,281 @@
...
+            The metric object itself.
+        """
+        api = HfApi()
+        for model_id in _hallucination_model_ids():
+            if not api.repo_exists(repo_id=model_id):
+                logger.warning(
+                    f"The hallucination detection model {model_id!r} does not exist "
+                    "on the Hugging Face Hub, skipping download."
+                )
+                continue
```
saattrupdan: Isn't this a warning that the user should report? In which case we can urge the user to open an issue on Github (with a link).

src/euroeval/tasks.py
```diff
@@ -294,6 +294,21 @@ def get_all_tasks() -> dict[str, "Task"]:
     default_labels=[],
 )
 
+HALLU = Task(
+    name="hallucination",
+    task_group=TaskGroup.QUESTION_ANSWERING,
+    template_dict=EMPTY_TEMPLATES,
+    metrics=[m.hallucination_metric],
+    default_num_few_shot_examples=0,
```
saattrupdan, line 314: Nit: We don't need to set this if `requires_zero_shot` is set.

src/frontend/md/datasets/danish.md
```diff
@@ -1927,3 +1927,64 @@ You can evaluate this dataset directly as follows:
...
+contains retrieval-augmented generation (RAG) prompts together with model-generated
+answers annotated for hallucinations. Rather than evaluating the correctness of the
+generated answer, this task evaluates the degree to which the model hallucinates,
+i.e., generates tokens that are not grounded in the provided context.
+
+The hallucination detection is performed using the
+[LettuceDetect](https://github.com/KRLabsOrg/LettuceDetect) library, which uses a
+transformer-based classifier to predict hallucination at the token level. The metric
+reported is the hallucination rate, computed as the ratio of hallucinated tokens to total
+tokens in the generated answers.
```
saattrupdan, line 2128: I would include a link to your own paper as well here 🙂

saattrupdan, line 2128: I would include a link to your own paper as well here 🙂

saattrupdan (commented):
Some more comments and suggestions

src/euroeval/metrics/token_hallucination_classifier.py
FrejaThoresen: Added TODO.

FrejaThoresen:
Thanks for the review @saattrupdan! The comments should all be resolved. Here are the main changes accordingly to your comments:
- Remove warning when model is not found on huggingface
- Add link to paper
- Add TODO for checking max length of inputs for the hallucination detector
- Using the MAX_CONTEXT_LENGTH from ..constants

PR #1633, opened 2026-03-16
fix(deps): improve vllm support on macOS

viggo-gascou:
**Problem**

Generation seems to hang indefinitely on vLLM v0.11.0 on macOS arm64 for me, requiring upgrading vLLM.
However, since v0.12.0 `flashinfer-python` became a default dependency (previously an optional extra), which causes uv to complain about missing CUDA/nvidia libs on macOS arm64 as `flashinfer-python`'s PyPI metadata unconditionally requires `nvidia-cudnn-frontend` which has no macOS wheels (which makes sense).
Additionally, `vllm-metal` has had several updates without any official PyPI releases, so the PyPI version lags significantly behind, which includes several improvements.

**Solution**

- Upgrade vllm to v0.14.1 (chosen to align with the existing Linux version, though newer versions also seem to work),  and building from source via `tool.uv.sources` to avoid the `flashinfer-python` issues.
- Switch `vllm-metal` to use git source directly for the same reason.
- Add `environments` to `tool.uv` to work around vllm's dependency markers causing uv's cross-platform resolver to look for `torch+cpu` for impossible platform combinations (`s390x+darwin`, `x86_64+darwin`), which only exists on PyTorch's own index and not PyPI.

**Breaking changes**

> [!WARNING]
> This drops support for macOS x86_64 (Intel Macs) — vllm-metal only targets Apple Silicon anyways but the `environments` restriction now explicitly limits macOS resolution to arm64.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Fixed
+
+- vLLM has been upgraded from v0.11.0 to v0.14.1 on macOS arm64, fixing generation
+  hanging indefinitely. `vLLM` and `vllm-metal` are now both installed from their git
+  sources rather than PyPI, as there are no wheels for macOS available for `vLLM` and
+  `vllm-metal` PyPI releases lag behind.
+  Note that this means that we no longer support macOS x86_64 (Intel Macs).
+
 ## [v17.0.0] - 2026-03-16
 
 ### Added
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -70,6 +70,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Removed the `save_safetensors` argument when initialising `TrainingArguments` in
   `finetuning.py`, as this has been removed from `transformers` and thus causes an error
   if set.
+- vLLM has been upgraded from v0.11.0 to v0.14.1 on macOS arm64, fixing generation
+  hanging indefinitely. `vLLM` and `vllm-metal` are now both installed from their git
+  sources rather than PyPI, as there are no wheels for macOS available for `vLLM` and
+  `vllm-metal` PyPI releases lag behind.
+  Note that this means that we no longer support macOS x86_64 (Intel Macs).
```
saattrupdan: Can you move this up to Unreleased?

viggo-gascou: Ah yes, didn't notice!

PR #1634, opened 2026-03-16
Feat: Wrong Language Penalization

### LINKED ISSUE #614: [BUG] BERTScore should use monolingual models

viggo-gascou:
This PR closes #614 by adding language penalization to SacreBLEU ChrF scoring by applying a binary language correctness penalty per prediction before averaging sentence-level ChrF scores.

If a prediction is detected as being in the wrong language, its ChrF contribution is set to 0. 
Without language penalization, models can receive good ChrF scores, and thus leaderboard ranks, even when generating in the wrong language. This makes rankings less faithful to the benchmark objective. Penalizing wrong-language outputs addresses this and clearly exposes models that fail to generate in the required language.

**What changed**

- Added a language detector for metric-time language validation, using [`lingua-py`](https://github.com/pemistahl/lingua-py)
- Updated ChrF computation to support language penalization.
- Applied per-prediction language penalties (1.0 / 0.0) before score aggregation.
- Since Danish and Norwegian are so similar in writing (sometimes identical), the confidence score of predictions in datasets in either Danish or Norwegian is a combination of the individual confidence scores for Danish and Norwegian.
- The minimum confidence level for a specific sample is set via the constant `MIN_LANG_CONFIDENCE_SCORE` and is currently set to `0.75`.

**Tests**

I have tested the addition of the language penalization on three models on the Danish summarization dataset `nordjylland-news`. The first two should receive little to no penalty and the `AI-Sweden` model should receive a much larger penalty due to it mainly responding in Swedish regardless of the input text, see the results below. 

| Model | Metric | Previous | New | Diff |
|---|---|---:|---:|---:|
| `deepcogito/cogito-v1-preview-llama-8B` | `chr_f3pp` | 37.29 ± 0.14 | 37.19 ± 0.79 (CI) / ± 0.20 (SE) | -0.10 |
| `deepcogito/cogito-v1-preview-llama-8B` | `chr_f4pp` | 40.02 ± 0.17 | 39.73 ± 0.68 (CI) / ± 0.17 (SE) | -0.29 |
| `allenai/Llama-3.1-Tulu-3-8B-SFT` | `chr_f3pp` | 30.09 ± 1.24 | 29.72 ± 2.33 (CI) / ± 0.59 (SE) | -0.37 |
| `allenai/Llama-3.1-Tulu-3-8B-SFT` | `chr_f4pp` | 30.54 ± 1.39 | 30.15 ± 2.61 (CI) / ± 0.67 (SE) | -0.39 |
| `AI-Sweden-Models/Llama-3-8B-instruct` | `chr_f3pp` | 27.35 ± 0.32 | 0.37 ± 0.43 (CI) / ± 0.11 (SE) | -26.98 |
| `AI-Sweden-Models/Llama-3-8B-instruct` | `chr_f4pp` | 29.70 ± 0.37 | 0.39 ± 0.45 (CI) / ± 0.12 (SE) | -29.31 |

Overall, these results look as expected: the `AI-Sweden-Models/Llama-3-8B-instruct` model is heavily penalized for wrong-language outputs, while the other two models show only minor changes. I have not yet been able to validate this on a broader set of models, since I do not have the enough compute available, so additional validation on more models is very welcome! 😄

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Fixed
 
+- Fixed a logging bug when evaluating models on the tool calling task.
 - vLLM has been upgraded from v0.11.0 to v0.14.1 on macOS arm64, fixing generation
   hanging indefinitely. `vLLM` and `vllm-metal` are now both installed from their git
   sources rather than PyPI, as there are no wheels for macOS available for `vLLM` and
--- a/src/euroeval/metrics/tool_calling.py
+++ b/src/euroeval/metrics/tool_calling.py
@@ -2,6 +2,7 @@
 
 import collections.abc as c
 import json
+import logging
 import typing as t
 
 from ..logging_utils import log_once
@@ -95,7 +96,8 @@ def _evaluate_function_toolcall_response(
         # get predicted function name
         if "function" not in pred_call:
             log_once(
-                "Tool call prediction did not contain required keyword 'function'."
+                "Tool call prediction did not contain required keyword 'function'.",
+                level=logging.DEBUG,
             )
             return False
         else:
@@ -104,7 +106,8 @@ def _evaluate_function_toolcall_response(
         # get predicted arguments
         if "arguments" not in pred_call:
             log_once(
-                "Tool call prediction did not contain required keyword 'arguments'."
+                "Tool call prediction did not contain required keyword 'arguments'.",
+                level=logging.DEBUG,
             )
             return False
         else:


src/euroeval/metrics/language_detection.py
```diff
@@ -0,0 +1,131 @@
...
+        # Danish and Norwegian variants are mutually included since they are
+        # difficult to distinguish, even for humans, and some sentences can be
+        # identical across them.
+        danish_and_norwegian = {DANISH, NORWEGIAN_BOKMÅL, NORWEGIAN_NYNORSK}
+        dataset_languages = set(dataset_config.languages)
+
+        if danish_and_norwegian & dataset_languages:
+            language_codes = [lang.code for lang in danish_and_norwegian]
+        else:
+            language_codes = [lang.code for lang in dataset_config.languages]
```
saattrupdan: Don't we want to _add_ all the languages in `danish_and_norwegian` if the intersection is non-empty? Right now we're replacing all the existing dataset languages with the Danish and Norwegian ones, so if `dataset_languages` consists of Danish and Swedish, then we end up losing Swedish, don't we?

src/euroeval/metrics/language_detection.py
```diff
@@ -0,0 +1,131 @@
...
+            if lang != NORWEGIAN.code
+        ]
+
+        return self._detect_language(
+            predictions=tuple(predictions),
+            detector_languages=detector_languages,
+            target_language_codes=tuple(target_language_codes),
+        )
+
+    @cache_arguments("predictions", "target_language_codes")
```
saattrupdan: Is `LinguaLanguage` not cacheable? Is that why `detector_languages` is not included here?

src/euroeval/metrics/sacrebleu.py
```diff
@@ -60,20 +87,28 @@ def __call__(
...
             ).score
             for prediction, reference in zip(predictions, references)
         ]
-        return 1.0 if not scores else sum(scores) / len(scores)
+
+        if not scores:
+            return 1.0
+
+        if self.language_detector is not None:
+            penalties = self.language_detector(predictions, dataset_config)
```
saattrupdan: [nit] Would prefer keyword arguments, as it can potentially prevent future issues.

src/euroeval/metrics/language_detection.py
```diff
@@ -0,0 +1,131 @@
...
+        target_language_codes = get_correct_language_codes(
+            language_codes=language_codes
+        )
+
+        # Skipping "no" norwegian as lingua does not have a code for it
+        detector_languages = [
+            LinguaLanguage.from_iso_code_639_1(IsoCode639_1.from_str(lang))
+            for lang in target_language_codes
+            if lang != NORWEGIAN.code
+        ]
```
saattrupdan, line 85: The language detection is being applied to all the ChrF metrics, which are also used for translation now. Translation dataset configs have two languages in their `languages` attribute: the source language and the target language. In this case we really only want to test for the target language (i.e., the second entry in `languages`).

I know it seems a bit hacky, but could we just drop the first language when loading `dataset_languages` above, when we're dealing with a translation task?

src/euroeval/metrics/language_detection.py
viggo-gascou: Ah yeah, didn't think that could be a possibility!

src/euroeval/metrics/language_detection.py
viggo-gascou: I don't think so, since lingua is converted from rust, but I don't think I've tried it, will investigate

src/euroeval/metrics/language_detection.py
viggo-gascou: Yeah i guess that could work, we could also add an optional `target_language` attributes in `DatasetConfig` instead? Although we do already have quite a lot of attributes in `DatasetConfig`

src/euroeval/metrics/language_detection.py
viggo-gascou: Seems like it is hashable at least, so I think it should be cacheable? Should I just remove all of the args since `cache_arguments` uses all args by default?

PR #1690, opened 2026-04-13
Fix(deps): enable structured output support on macOS

### LINKED ISSUE #1477: Figure out support for Apple Silicon
[vLLM does seem to have some support ](https://docs.vllm.ai/en/latest/getting_started/installation/cpu/#apple-silicon)either through vllm-metal. Though I couldn't get the installation to work, related issue is here:

https://github.com/vllm-project/vllm-metal/issues/83

Another direction is to build vLLM from source (from docs):

> vLLM has experimental support for macOS with Apple Silicon. For now, users must build from source to natively run on macOS.

The intention of this is mostly to remove a sharp edge mentioned here:

 _Originally mentioned by @KennethEnevoldsen in [#1473](https://github.com/EuroEval/EuroEval/issues/1473#issuecomment-3806087485)_

I see it as likely that we decide to wait for support for Apple Silicon (using vLLM). In that case, this issue can act as a pointer to others that it has been examined.

viggo-gascou:
This PR updates `vllm` from `v0.14.1` to `v0.18.1` on macOS. Previously, benchmarking generative models on tasks on was not possible, due to an `xgrammar` incompatibility. The updated version resolves this, so the restriction has been removed. This should close #1477

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed a bug when evaluating a generative task that uses logprobs with partial
   completions. The issue was that the first label token mapping wasn't updated
   immediately after the dataset config changed, which has been fixed now.
+- Updated vLLM from v0.14.1 to v0.18.1 on macOS Apple Silicon, enabling structured
+  output support for tasks such as sequence classification and multiple choice.
 
 ## [v17.1.0] - 2026-03-24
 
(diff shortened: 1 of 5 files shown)


viggo-gascou:
Hmm, not sure why that test is failing, it shouldn't based on what is updated here. It is also failing from main branch. I'll investigate 
[edit] from what I can tell the `01-ai/Yi-6B` model already adds a prefix space so we don't need to add it ourselves, so the expected value in the test is wrong. I'll make a PR with a fix

saattrupdan (approved):
LGTM, thanks!

PR #1691, opened 2026-04-14
Fix: Broken Tests

viggo-gascou:
This PR fixes three test issues:

- The expected value for `01-ai/Yi-6B` in `test_should_prefix_space_be_added_to_labels` was incorrect, the tokenizer automatically adds a prefix space, so the expected value should be `False`, not `True`.
- The model ID for `anthropic-model` used `claude-3-5-haiku-20251001` in `test_get_model_config` which is deprecated, causing the test to be skipped. Updated to `claude-haiku-4-5-20251001`.
- The no-internet encoder test was failing because disable_socket blocked all socket connections including `localhost`, which is required for the test to run. Changed to `allow_hosts(["127.0.0.1"])` to block external connections while still allowing `localhost`.

All tests pass locally, except for the ones that require API keys.

### DIFF
--- a/tests/test_benchmarker.py
+++ b/tests/test_benchmarker.py
@@ -148,7 +148,7 @@ def test_benchmark_ollama(
     assert all(isinstance(result, BenchmarkResult) for result in benchmark_result)
 
 
-@pytest.mark.disable_socket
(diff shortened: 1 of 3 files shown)


saattrupdan (approved):
LGTM!

Issue #1698, opened 2026-05-04
[FEATURE REQUEST] Cloze Formulation support


tvosch:
## What

Add **Cloze Formulation (CF)** as an alternative to the existing Multiple Choice
Formulation (MCF) for MCQ tasks.

In MCF the model sees the question with enumerated choices (`a.`/`b.`/`c.`/`d.`)
and we pick the label whose **first token** has the highest logprob. In CF we score
each full answer text as a scored continuation:

```text
score(choice_i) = sum(log P(answer_tokens | bare_question_prompt))
```

The winning choice is the one with the highest score after optional length
normalization.

## Why

LLMs struggle with MCF early in training so for ablations and smaller models.
Cloze formulation generally provides a better proxy for smaller-scale models and ablations.
[SmolLM playbook](https://huggingface.co/spaces/HuggingFaceTB/smol-training-playbook)

## Reference implementations

| Project | Approach |
| --- | --- |
| [OLMES (Allen AI, 2024)](https://arxiv.org/abs/2406.08446) | Defines CF as the principled alternative to MCF; recommends character-length normalization |
| [EleutherAI LM Evaluation Harness](https://github.com/EleutherAI/lm-evaluation-harness) | Uses `loglikelihood` per choice; character normalization by default |

## Normalization options

| Method | Formula |
| --- | --- |
| `none` | `Σ log P(token)` |
| `token` | `Σ log P(token) / num_tokens` |
| `character` | `Σ log P(token) / num_chars` ← default (matches OLMES / Llama) |

## Usage preview

```bash
euroeval \
  --model <hf-model-id> \
  --dataset belebele-nl \
  --scoring-method cf \
  --cf-normalization character
```

Drop `--scoring-method` (or set it to `mcf`) to keep the existing behaviour
unchanged.

## Scope

- HuggingFace and vLLM backends can support CF as both expose per-token logprobs
  via a standard forward pass.
- LiteLLM does not reliably expose logprobs, so CF is not supported there.
- CF is an unofficial, non-default scoring method:  `mcf` remains the default.
- Using CF with a non-MCQ task raises `InvalidBenchmark` regardless of backend.

### Concrete example — `belebele-nl`, sample 329

**MCF** — model sees all choices, must produce one letter token:

```text
Hieronder staan meerkeuzevragen (met antwoorden).

[few-shot examples …]

Vraag: Tekst: Fluke schreef het volgende: sommige mensen probeerden vrouwen
[...]

PR #1699, opened 2026-05-04
Feat/Cloze Formulation support

tvosch:
Implements #1698 
### Changes

- **New `--scoring-method` CLI flag** (`mcf` default, `cf` opt-in) and
  **`--cf-normalization`** (`none` / `token` / `character`, default `character`)
- **New `EvaluationType` and `CFNormalization` enums**
- **New `src/euroeval/task_group_utils/cloze.py`** — prompt building, few-shot
  rendering, score normalization, label extraction
- **vLLM backend**: `_generate_cf` + `score_completions` using `prompt_logprobs`;
  `prepare_dataset` builds bare-question prompts with full answer text in few-shot
- **HF encoder backend**: raises `InvalidBenchmark` for CF (encoder models cannot
  do cloze scoring)
- **LiteLLM backend**: raises `InvalidBenchmark` for CF (logprob exposure is
  unreliable)
- Tests for all new utilities, vLLM CF path, and backend guards

### Scope / non-goals

- CF is opt-in and unofficial; `mcf` remains the default
- Only vLLM is supported
- Non-MCQ tasks raise `InvalidBenchmark` regardless of backend

### Results

```bash
euroeval --model HPLT/hplt2c_nld_checkpoints --dataset belebele-nl --evaluation-type cf --cf-normalization character --evaluate-test-split
- Matthew's Correlation Coefficient: 15.89% ± 1.72%
- Accuracy: 36.86% ± 1.31%

euroeval --model HPLT/hplt2c_nld_checkpoints --dataset belebele-nl --evaluation-type cf --cf-normalization token --evaluate-test-split
- Matthew's Correlation Coefficient: 15.33% ± 1.29%
- Accuracy: 36.43% ± 0.97%

euroeval --model HPLT/hplt2c_nld_checkpoints --dataset belebele-nl --evaluation-type cf --cf-normalization none --evaluate-test-split
Finished benchmarking HPLT/hplt2c_nld_checkpoints on the truncated version of the Dutch multiple choice reading comprehension dataset Belebele-nl.
- Matthew's Correlation Coefficient: 16.32% ± 1.08%
- Accuracy: 37.16% ± 0.81%
```

Lighteval==0.10.0 reference:

```bash
lighteval accelerate "model_name=HPLT/hplt2c_nld_checkpoints,batch_size=1" "lighteval|belebele_nld_Latn_cf|5|0" --custom-tasks load_multilingual.py
| Task | Version | Metric | Value |
|------|--------|--------|-------|
| all |  | acc_norm_token | 0.3744 |
|  |  | acc_norm | 0.3911 |
| lighteval:belebele_nld_Latn_cf:5 | 0 | acc_norm_token | 0.3744 |
|  |  | acc_norm | 0.3911 |
```


saattrupdan (commented):
Rather than having extra configuration, why don't we just add cloze as a separate metric to multiple choice tasks? If it's not supported, we can just set it to NaN. Then there's no extra complexity. We'd have to see if it increases the evaluation time too much, however.

tvosch:
@saattrupdan I thought about this too, though the added evaluation time could be significant, hence why the CF is decoupled via arguments.
To be fair, I think the metric is relatively niche. Personally I only use it for ablation studies on smaller undertrained models and once the model is sufficiently pretrained MCQ usually suffices anyway

src/euroeval/enums.py
```diff
@@ -148,3 +148,40 @@ class TaskGroup(AutoStrEnum):
     QUESTION_ANSWERING = auto()
     TEXT_TO_TEXT = auto()
     SPEED = auto()
+
+
+class EvaluationType(AutoStrEnum):
```
saattrupdan: **Rename `EvaluationType` → `ScoringMethod`** and `evaluation_type` → `scoring_method` throughout (CLI handling, `Benchmarker` kwargs, `BenchmarkConfig`, `BenchmarkConfigParams`, `BenchmarkResult`, gating in backends, tests).

Two reasons:
1. The CLI flag is already `--scoring-method` — the Python field name should match the user-facing surface.
2. "Evaluation type" reads like it belongs next to `task_group` (MCQ vs QA vs sequence classification). "Scoring method" is what this actually is: *how* did we score an MCQ.

`CFNormalization` and the per-call `cf_normalization` kwarg stay as-is.

src/euroeval/data_models.py
```diff
@@ -770,6 +784,8 @@ class BenchmarkConfig:
     run_with_cli: bool
     max_context_length: int | None
     vocabulary_size: int | None
+    evaluation_type: EvaluationType = EvaluationType.MCF
+    cf_normalization: CFNormalization = CFNormalization.CHARACTER
```
saattrupdan: **`BenchmarkResult` needs the same discriminator fields** (using the renamed `scoring_method`, see comment on `enums.py`):

```python
class BenchmarkResult(pydantic.BaseModel):
    ...
    scoring_method: str = "mcf"
    cf_normalization: str | None = None
```

Defaults are safe so existing jsonl entries continue to round-trip.

And backfill them in `BenchmarkResult.from_dict` (~line 895 area, where `few_shot`/`validation_split` are already backfilled):

```python
if "scoring_method" not in config:
    config["scoring_method"] = "mcf"
if "cf_normalization" not in config:
    config["cf_normalization"] = None
```

Without this, CF and MCF runs are indistinguishable in `euroeval_benchmark_results.jsonl`, and the leaderboard collector has nothing to filter on.

src/euroeval/benchmarker.py
```diff
@@ -222,6 +239,8 @@ def __init__(
             gpu_memory_utilization=gpu_memory_utilization,
             attention_backend=attention_backend,
             generative_type=generative_type,
+            evaluation_type=evaluation_type,
+            cf_normalization=cf_normalization,
```
saattrupdan: **Populate the new `BenchmarkResult` fields from `benchmark_config`** in the `BenchmarkResult(...)` construction at line 1069 (not in this diff, but needs updating):

```python
scoring_method=str(benchmark_config.scoring_method),
cf_normalization=(
    str(benchmark_config.cf_normalization)
    if benchmark_config.scoring_method == ScoringMethod.CF
    else None
),
```

(Using the renamed `scoring_method` / `ScoringMethod` — see comment on `enums.py`.)

src/euroeval/benchmark_modules/vllm.py
```diff
@@ -947,6 +1023,136 @@ def generate(self, inputs: dict) -> "GenerativeModelOutput":
...
+            for candidate in candidates:
+                full = prompt + candidate
+                full_ids = list(tok(full, add_special_tokens=False).input_ids)
+                # The prompt's final token often merges with the first candidate
+                # token at tokeniser seams (e.g. SentencePiece "_Foo" vs "Foo").
+                # Find the longest prefix of the prompt's token sequence that
+                # also prefixes the full sequence, then treat the remaining
+                # tokens as the continuation. At least one continuation token
+                # is required.
+                prefix_len = len(prompt_ids)
```
saattrupdan: **The empty-candidate guard misses the case where `prompt_ids == full_ids`.** The loop's first clause (`prefix_len >= len(full_ids)`) shrinks `prefix_len` even when the prompt and full sequences are equal-length, so the post-loop guard never fires and we read the *last prompt token's* logprob as a phantom continuation score. `test_empty_candidate_documented_boundary` documents this as a known wrong-result path.

Suggested fix — add an upfront check before the trim loop:

```suggestion
                # An empty candidate (or one whose tokens are entirely
                # absorbed at the prompt seam) has nothing to score.
                if len(full_ids) <= len(prompt_ids) and (
                    full_ids == prompt_ids[: len(full_ids)]
                ):
                    raise InvalidBenchmark(
                        "Candidate is empty after tokenisation: cannot "
                        f"score an empty continuation. Prompt: {prompt!r}, "
                        f"candidate: {candidate!r}."
                    )
                # The prompt's final token often merges with the first candidate
                # token at tokeniser seams (e.g. SentencePiece "_Foo" vs "Foo").
                # Find the longest prefix of the prompt's token sequence that
                # also prefixes the full sequence, then treat the remaining
                # tokens as the continuation. At least one continuation token
                # is required.
                prefix_len = len(prompt_ids)
```

tests/test_benchmark_modules/test_vllm.py
```diff
@@ -317,3 +317,250 @@ def test_load_model_passes_correct_max_model_len_to_llm(
...
+        prompt_logprobs = [None, {2: self._lp(-0.1)}, {3: self._lp(-0.2)}]
+        model = self._build_model(
+            tokenize_map=tokenize_map, prompt_logprobs_per_call=[prompt_logprobs]
+        )
+        with patch("euroeval.benchmark_modules.vllm.SamplingParams", create=True):
+            scores = model.score_completions(
+                prompts=[prompt], completions=[[candidate]]
+            )
+        # Continuation reads the *last* prompt token's logprob — see note above.
+        assert scores == [[[-0.2]]]
```
saattrupdan: **Flip this test to assert the raise** once the empty-candidate bug is fixed (see comment on `vllm.py`). "Production datasets never have empty choice strings" is true today but it's a future foot-gun — if anyone later adds a dataset with `""` as a placeholder choice, scoring will silently rank it.

```suggestion
    def test_empty_candidate_raises(self) -> None:
        """An empty candidate raises `InvalidBenchmark` rather than scoring."""
        prompt = "Q"
        candidate = ""
        tokenize_map = {
            prompt: [1, 2, 3],
            prompt + candidate: [1, 2, 3],  # full_ids == prompt_ids.
        }
        prompt_logprobs = [None, {2: self._lp(-0.1)}, {3: self._lp(-0.2)}]
        model = self._build_model(
            tokenize_map=tokenize_map, prompt_logprobs_per_call=[prompt_logprobs]
        )
        with patch("euroeval.benchmark_modules.vllm.SamplingParams", create=True):
            with pytest.raises(InvalidBenchmark, match="empty after tokenisation"):
                model.score_completions(
                    prompts=[prompt], completions=[[candidate]]
                )
```

src/euroeval/task_group_utils/cloze.py
```diff
@@ -0,0 +1,242 @@
...
+    Returns:
+        The rendered few-shot section.
+    """
+    return prompt_template.format(
+        text=bare_input.replace("\n", " ").strip(),
+        label=answer_text.replace("\n", " ").strip(),
+    )
+
+
+def parse_mcq_text(text: str) -> tuple[str, list[str]]:
```
saattrupdan: Nit: a one-line note in the docstring clarifying that this function is now a **fallback** for pre-CF cached datasets would help future readers. The preprocessing step now always writes `bare_input`/`raw_choices` directly (per the changes in `preprocessing.py`), so this regex-based parsing only fires when an old cached dataset is loaded.

src/euroeval/task_group_utils/cloze.py
```diff
@@ -0,0 +1,242 @@
...
+        prompt_prefix:
+            The dataset's prompt prefix (may be empty). ``{labels_str}``, if
+            present, is substituted with an empty string because CF does not
+            enumerate the label letters.
+
+    Returns:
+        The full prompt string.
+    """
+    tail = prompt_template.format(text=bare_input.replace("\n", " ").strip(), label="")
+    rendered_prefix = prompt_prefix.format(labels_str="") if prompt_prefix else ""
```
saattrupdan: Nit: a `# Only {labels_str} is currently used in prompt_prefix templates.` comment would prevent a silent `KeyError` if someone adds a new format slot to a prefix template down the line. (I checked all current `dataset_configs/*.py` and only `{labels_str}` appears, so this is safe today.)

saattrupdan (changes requested):
*Heads up: this review was drafted by Claude (Anthropic's AI assistant), acting on @saattrupdan's behalf. He's reviewed it before posting. Happy to clarify or revise any point.*

Thanks for the careful work on this — the tokeniser-seam handling in `score_completions`, the cache namespacing with the explicit MCF bit-identity test, and the gating in the non-vLLM backends are all really well done. The reference comparison against lighteval in the PR description is also great.

A few changes needed before merge, summarised here with detailed comments inline.

**Blocking**:

1. **`BenchmarkResult` needs a discriminator field.** The CF/MCF distinction is currently invisible in `euroeval_benchmark_results.jsonl` — two runs of the same model+dataset with different `--scoring-method` are indistinguishable on disk, and the leaderboard collector has nothing to filter on. We want the leaderboard to only include MCF runs, so this needs to be persisted, backfilled in `from_dict`, populated when constructing the result, and round-tripped through the EEE conversion. See inline comments on `data_models.py` and `benchmarker.py`.
2. **Rename `evaluation_type` → `scoring_method`.** The CLI flag is already `--scoring-method`, and "evaluation type" reads like it belongs alongside `task_group` (MCQ vs QA vs ...) rather than "how did we score the MCQ". See inline comment on `enums.py`.
3. **Fix the empty-candidate bug in `score_completions`.** Currently silently scores the last prompt-token's logprob as a phantom continuation; the existing test even documents this as a known wrong-result path. See inline comments + suggestions on `vllm.py` and `test_vllm.py`.
4. **Add a `CHANGELOG.md` entry** under `## [Unreleased] > ### Added`. Suggested wording:

   ```markdown
   - Added opt-in Cloze Formulation (CF) scoring for multiple-choice tasks, as an
     alternative to the default Multiple-Choice Formulation (MCF). Selected via
     `--scoring-method cf`, with length normalization controlled by
     `--cf-normalization` (`none` / `token` / `character`, default `character`).
     Currently supported only by the vLLM backend; the HF encoder and LiteLLM
     backends raise `InvalidBenchmark` if CF is requested. The selected scoring
     method is now persisted in `euroeval_benchmark_results.jsonl` so MCF and CF
     runs can be distinguished. Thanks to [@tvosch](https://github.com/tvosch)
     for the contribution!
   ```

**Also needed** (not modified by this PR so no inline anchor): **update `src/euroeval/eee_utils.py`** to carry the new `scoring_method` and `cf_normalization` fields through the EEE conversion, otherwise the EEE round-trip drops them silently:

- In `benchmark_result_to_eee_dict`, add the two fields to `eval_lib_additional_details` (alongside `few_shot`/`validation_split`).
- In `benchmark_result_from_eee_dict`, read them back via `eval_lib_additional.get(...)`, defaulting `scoring_method` to `"mcf"` and `cf_normalization` to `None`.

On the broader "should the discriminator be a generic params dict for future modes?" question — I'd say no. Flat sibling fields (NULL when not applicable) stay greppable in the jsonl, give the leaderboard collector a single column to filter on, and keep type safety. Future scoring modes can add their own sibling fields. The cost is one column per mode-family, which is appropriate friction — it forces a leaderboard-semantics conversation each time someone adds a scoring variant.

**Nice-to-haves** (non-blocking, can land later):
- User-facing log when CF mode is active — currently nothing distinguishes a CF run from MCF in the logs until results come out.
- Inline comments on a couple of fragile bits of `cloze.py` (see inline).

Happy to push the blocker fixes to your branch directly if it's easier — `maintainerCanModify` is on. Otherwise let me know and I'll wait for an update.

saattrupdan:
Hey @tvosch, is this ready for review? Saw you made some changes.

saattrupdan:
@tvosch I'm also thinking that this could be a more generally useful evaluation method, outside only covering multiple-choice (doesn't have to be in this PR). In general it would be useful to evaluate base decoders on all tasks using BPB, where we also use the cloze formulation on the multiple-choice tasks. This follows the methodology in [the Olmix paper](https://arxiv.org/abs/2602.12237).

Only a few changes missing I think (again, not necessary in this PR):
- Byte normalisation
- Using your scoring helper function on the prompt+label, instead of generating. This will work for all tasks, as it's already set up like that for base decoders.
- Having a single CLI flag to enable this, I think. Something like `--report-bits-per-byte`.

saattrupdan:

**Blocking items fixed:**
1. ✓ `BenchmarkResult` already had `scoring_method` and `cf_normalization` fields with backfill in `from_dict`
2. ✓ `evaluation_type` → `scoring_method` rename already done (enum was already named `ScoringMethod`)
3. ✓ Fixed empty-candidate bug in `score_completions` - added upfront guard for when candidate tokens are entirely absorbed at prompt seam
4. ✓ CHANGELOG entry already present from merge
5. ✓ EEE conversion already handles both fields in `eee_utils.py`

**Additional fixes:**
- Added test `test_candidate_absorbed_at_seam_raises` for the boundary case
- Replaced all `pyrefly:` ignore comments with `ty:`
- Removed unused `ty: ignore` comments from `question_answering.py`

All checks pass (`make check`).

saattrupdan:
Addressed Copilot review comments:

✅ **Simplified backward compatibility** — Since we don't care about pre-CF cached datasets, removed `parse_mcq_text()` entirely and its tests. The `_ensure_cf_columns()` function now just returns the expected columns (preprocessing always provides them).

✅ **Docstring nits** — Added comment about `{labels_str}` being the only currently-used slot in prompt_prefix templates.

All other review items were already addressed in earlier commits:
- ✅ `EvaluationType` → `ScoringMethod` rename
- ✅ `BenchmarkResult` discriminator fields added
- ✅ Empty-candidate guard with upfront check
- ✅ Test for empty candidate raising

@copilot Ready for re-review.

rakkit:
> Having a single CLI flag to enable this, I think. Something like --report-bits-per-byte.

in practice we could compute CF & BPB together as in formula they are same (rather than run infer twice)

saattrupdan:
Fixed all test failures from CF scoring method changes:

- **HF/LiteLLM gating tests**: Updated to expect `InvalidModel` (correct exception for backend support issues)
- **BenchmarkResult tests**: Added `scoring_method='mcf'` to expected fixture values (backfill sets default)
- **Cache namespace tests**: Fixed mock to include `task.task_group`, and updated cache suffix logic to only add `-cf` for CF (MCF uses legacy unsuffixed path for backward compatibility)

All 11 previously-failing tests now pass, `make check` passes.

saattrupdan:
> > Having a single CLI flag to enable this, I think. Something like --report-bits-per-byte.
> 
> in practice we could compute CF & BPB together as in formula they are same (rather than run infer twice)

Yeah good point. Also, if we're using CF and BPB together, then we really only need to do inference for the correct answer, saving a lot of compute as well. This is also the method used in [Olmix](https://arxiv.org/abs/2602.12237).

Is there a use case where we'd like CF without BPB? Because if not then we might as well merge the two.

rakkit:
> is there a use case where we'd like CF without BPB? Because if not then we might as well merge the two.

its more like another way around.

For some evals, if we explicitly want MCF (e.g for big model train instead of proxy), then we need to define both MCF, and CF&BPB.

For genQA there is no "Choice", so we need to define generative formula (f1/em), and BPB-only (no cf).

PR #1695, opened 2026-04-22
Adding Dutch proverbs

lswiers:
This PR adds a benchmark for the purpose of testing the knowledge of Dutch proverbs. The accompanying dataset was created in the context of the [GPT-NL project](https://gpt-nl.nl/).

The dataset consists of brief scenarios and two possible proverbs for the Large Language Model to select from. The dataset was created manually and reviewed by native Dutch speakers.

PR #1700, opened 2026-05-07
feat: add Gemma4TextForCausalLM architecture alias for vLLM compatibility

lardinator:
## Summary

- Adds an `_ARCHITECTURE_ALIASES` dict to remap HuggingFace architecture names that vLLM does not recognise to their vLLM-compatible equivalents
- Maps `Gemma4TextForCausalLM` → `Gemma4ForCausalLM`, using vLLM's `hf_overrides` parameter so the model-loader picks up the supported class name
- Also updates the in-memory `hf_model_config` so EuroEval's own code sees the remapped architecture

## Context

[Transformers 4.57](https://github.com/huggingface/transformers/releases/tag/v4.51.0) split Gemma 4 into two architecture classes:
- `Gemma4ForConditionalGeneration` — multimodal (text+vision+audio)
- `Gemma4TextForCausalLM` — text-only backbone

vLLM (as of 0.19.0) only registers `Gemma4ForCausalLM`, so loading any model whose `config.json` declares `Gemma4TextForCausalLM` fails with:
```
Model architectures ['Gemma4TextForCausalLM'] are not supported for now.
```

This affects text-only Gemma 4 fine-tunes (e.g. continual pre-training runs) that use the text-only class.

## Design

The fix uses an extensible alias dict (`_ARCHITECTURE_ALIASES`) so future architecture mismatches can be resolved by adding a single line. The remapping is passed to vLLM via the [`hf_overrides` parameter](https://docs.vllm.ai/en/latest/configuration/model_resolution.html), which is the documented way to override model config at load time.

## Test plan

- [ ] Load a Gemma 4 text-only model (architecture `Gemma4TextForCausalLM`) via EuroEval — should succeed without error
- [ ] Verify the INFO log message "Remapping model architectures..." appears
- [ ] Load a standard model (e.g. Llama) — no remapping should occur, no change in behaviour


lardinator:
The `code-check` failure is an import sorting issue in `src/scripts/create_dutch_proverbs.py` on the main branch (introduced in #1695), not in this PR. Running `ruff check --fix` on main reproduces it:

```
Found 1 error (1 fixed, 0 remaining).
```

This PR only modifies `src/euroeval/benchmark_modules/vllm.py`. The pytest checks pass.

saattrupdan (approved):
Looks great, thanks for the fix 🙂

PR #1705, opened 2026-05-17
feat(evaluation-queue): skip oversized models via safetensors header

saattrupdan:
## Summary
- Reads each candidate model's safetensors header and computes dtype-aware weight bytes (handles int8/fp8/int4 correctly).
- Compares the estimate (× 1.2 overhead for activations/KV cache) against the local GPU memory reported by `nvidia-smi`.
- If the model is too large, the issue is left **unassigned** so a larger machine can pick it up — no status change, no comment.
- Pre-check is skipped silently when GPU memory cannot be determined or the repo is not safetensors, preserving today's behavior.
- `EUROEVAL_GPU_MEMORY_BYTES` env var overrides the detected value (useful in tests).

## Test plan
- [ ] Run on a machine without `nvidia-smi` → log says "Could not determine local GPU memory" and queue proceeds as before.
- [ ] Set `EUROEVAL_GPU_MEMORY_BYTES=1000000000` and run against a queue containing a 7B model → that issue is logged as skipped and left unassigned.
- [ ] Set `EUROEVAL_GPU_MEMORY_BYTES=100000000000` → queue proceeds to evaluate the model as before.
- [ ] Issue with a model that has no safetensors weights → log notes the metadata is unavailable and the issue is processed (no false skip).

### DIFF
--- a/src/scripts/process_evaluation_queue.py
+++ b/src/scripts/process_evaluation_queue.py
@@ -43,6 +43,8 @@ from huggingface_hub.errors import (
     SafetensorsParsingError,
 )
 
+from euroeval.string_utils import split_model_id
+
 logging.basicConfig(
     level=logging.INFO, format="%(asctime)s | %(levelname)s | %(message)s"
 )
@@ -359,8 +361,9 @@ def estimated_model_bytes(model_id: str) -> int | None:
         The total weight bytes across all tensors, or None when the repo is
         not a safetensors repo or the metadata cannot be parsed.
     """
-    repo, _, revision = model_id.partition("@")
-    revision = revision or None
+    components = split_model_id(model_id=model_id)
+    repo = components.model_id
+    revision = components.revision
     token = os.environ.get("HF_TOKEN") or os.environ.get("HUGGINGFACE_API_KEY")
     try:
         meta = get_safetensors_metadata(repo_id=repo, revision=revision, token=token)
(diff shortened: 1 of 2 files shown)


src/scripts/process_evaluation_queue.py
```diff
@@ -63,6 +70,31 @@
 # "Waiting for bug fix" statuses.
 ERROR_MARKER_RE = re.compile(r"<!--\s*errored-on:\s*v([^\s>-]+)\s*-->")
 
+# Bytes-per-parameter for each safetensors dtype string. Used to compute the
+# weight footprint of a model from its safetensors header before deciding
+# whether it can fit on the local GPU.
+DTYPE_BYTES: dict[str, int] = {
```
saattrupdan, line 82: @copilot What about if a dtype doesn't match anything here? We should have a fallback (where we look for occurrences of "4", "8", "16", "32" and so on to guess the dtype, and if all fails then we assume BF16.

src/scripts/process_evaluation_queue.py
```diff
@@ -256,6 +307,86 @@ def huggingface_param_count(info: object) -> int:
...
+
+    Args:
+        model_id:
+            The Hugging Face repo id, optionally suffixed with ``@<revision>``.
+
+    Returns:
+        The total weight bytes across all tensors, or None when the repo is
+        not a safetensors repo or the metadata cannot be parsed.
+    """
+    repo, _, revision = model_id.partition("@")
```
saattrupdan: Remember that model ID can also have parameters like "<model-id>#low", and @ and # can appear in any order. There is already a `split_model_id` in the `string_utils` module that you can use to split it up.

PR #1708, opened 2026-05-17
feat(evaluation-queue): validate eval results before reporting success

saattrupdan:
## Summary
Most euroeval failures are caught internally and the subprocess exits 0, so the previous \`ok = returncode == 0\` check happily posted a comment even when no results were produced. This PR:

- Streams euroeval's combined output through stderr while also capturing it, preserving live progress.
- Treats a run as failed if the subprocess crashed, produced no new JSONL lines, or reported \`errored N benchmarks\` with N > 0.
- On failure, scans captured output for OOM patterns. If matched → return the issue to the queue unmarked (like #1705's pre-check).
- Otherwise → mark errored + post an error comment with the reason and the tail of captured output.

## Test plan
- [ ] Run against an issue where the eval succeeds → results comment posted as today.
- [ ] Force an OOM (e.g. set tiny GPU memory or run a huge model) → no comment, issue unassigned, log says "OOM detected … returning issue to queue".
- [ ] Force a non-OOM error (e.g. broken model) → error comment posted, \`errored-on:\` marker set, issue unassigned.
- [ ] Force a partial failure where some datasets error but euroeval exits 0 → treated as failure, error comment posted (not OOM).

### DIFF
--- a/src/scripts/process_evaluation_queue.py
+++ b/src/scripts/process_evaluation_queue.py
@@ -63,6 +63,24 @@ _WGERMANIC = "West Germanic languages (Dutch, English, German)"
 # "Waiting for bug fix" statuses.
 ERROR_MARKER_RE = re.compile(r"<!--\s*errored-on:\s*v([^\s>-]+)\s*-->")
 
+# Patterns we treat as OOM in the captured euroeval output. When any of these
+# match we leave the issue unassigned and unmarked so a larger machine can pick
+# it up, rather than recording a hard error on the issue.
+OOM_PATTERN = re.compile(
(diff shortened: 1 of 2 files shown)


saattrupdan (commented):
@copilot Don't check for OOM, as we're already checking that the model can fit before we start evaluating. But you should check that all the official datasets (i.e., the ones used in the leaderboards) in the language groups have evaluation results in the JSONL, before proceeding.

saattrupdan:
@copilot resolve the merge conflicts in this pull request

saattrupdan (commented):
LGTM

PR #1709, opened 2026-05-17
feat: cache HF Hub lookups to avoid rate limiting

saattrupdan:
## Summary
The queue processor hit `HfApi.model_info()` once per candidate per run, so a chatty cron tripped HF's API limits. This PR adds caching at two levels:

**Queue processor (persistent):**
- New `cached_model_summary()` backed by `~/.cache/euroeval/queue_hf_cache.json` (path overridable via `EUROEVAL_QUEUE_HF_CACHE`).
- 6-hour TTL for positive results; negatives are not cached so typo corrections take effect immediately.
- Replaces `huggingface_model_info()` + `huggingface_param_count()` in the main loop.

**Package (in-process):**
- `cache_arguments(\"dataset_id\")` on `get_repo_split_names`.
- Cached wrappers `_list_repo_files` and `_repo_exists` in `custom_dataset_configs.py`, eliminating duplicate Hub calls when loading custom datasets across many (dataset, model) combinations in one run.

## Test plan
- [ ] Run the queue twice in quick succession against the same backlog → second run has no HF traffic for already-seen models (verify via `pip install responses` or by checking the cache file).
- [ ] `rm ~/.cache/euroeval/queue_hf_cache.json` → next run repopulates it.
- [ ] Edit a queue issue's model id from a typo to a real one → still resolves on the next run (negatives not cached).
- [ ] Run a benchmark touching a single custom HF dataset across multiple models → HF dataset lookups happen exactly once for that dataset.

### DIFF
--- a/src/euroeval/custom_dataset_configs.py
+++ b/src/euroeval/custom_dataset_configs.py
@@ -13,11 +13,50 @@ from types import ModuleType
 
 from huggingface_hub import HfApi
 
+from .caching_utils import cache_arguments
 from .data_models import DatasetConfig
 from .logging_utils import log_once
 from .split_utils import get_repo_splits
 from .utils import get_hf_token
 from .yaml_config import load_yaml_config
+
+
+@cache_arguments("dataset_id", "revision")
+def _list_repo_files(hf_api: HfApi, dataset_id: str, revision: str) -> list[str]:
+    """Cached wrapper around ``HfApi.list_repo_files`` for dataset repos.
+
+    Args:
+        hf_api:
+            The Hugging Face API client.
+        dataset_id:
+            The dataset repo id.
+        revision:
+            The git revision to list files from.
+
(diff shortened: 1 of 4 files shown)


src/euroeval/custom_dataset_configs.py
```diff
@@ -13,13 +13,52 @@
...
+    Args:
+        hf_api:
+            The Hugging Face API client.
+        dataset_id:
+            The dataset repo id.
+
+    Returns:
+        True if the dataset repo exists on the Hub, otherwise False.
+    """
+    return hf_api.repo_exists(repo_id=dataset_id, repo_type="dataset")
```
saattrupdan: @copilot Can you move these to a new `hf_hub_utils` module?

saattrupdan (commented):
LGTM

PR #1846, opened 2026-05-28
feat: add ordinal ranks and confidence intervals to leaderboard

saattrupdan:
## Summary

Adds true ordinal ranks alongside mean rank scores with confidence intervals to the EuroEval leaderboard.

### Backend (`src/leaderboards/`)

- **score_computation.py**
  - Rewrote `compute_ranks` to use mean-of-normalised-differences instead of Welch's t-test gated cumulative ranking
  - Return type changed from `float` to `dict` with `score`, `ci_lower`, `ci_upper` keys
  - Added `compute_standard_ranks` for ordinal ranks (1, 2, 3…) with ties using Welch's t-test

- **leaderboard_generation.py**
  - Updated `generate_dataframe` to use new rank structure
  - Added `standard_ranks` parameter
  - New columns: `rank` (ordinal, integer) and `mean_rank_score` (formatted as "1.23 ± 0.08")
  - Updated column ordering and simplified leaderboard

- **core_models.py**
  - Updated type annotations and rank access pattern for `_pareto_languages_per_model`

### Frontend (`src/frontend/components/`)

- **LeaderboardTable.vue**
  - Added `NO_HEATMAP_COLS` set to exclude `rank` and `mean_rank_score` columns from heatmap styling
  - Both columns now display as plain text (no heatmaps)

## Key design decisions

- **No intermediate CIs** — only overall level (and dataset level, which already exists)
- **Missing data** — models missing datasets are excluded entirely, so no partial-aggregation edge case
- **95% CI** throughout, matching the dataset-level convention
- **`significantly_better`** is reused for tie-breaking in standard ranks; it stays in `utils.py` unchanged

## Testing

- All pre-commit checks pass (ruff, pyrefly, etc.)

### DIFF
--- a/src/frontend/components/LeaderboardTable.vue
+++ b/src/frontend/components/LeaderboardTable.vue
@@ -22,16 +22,23 @@ const sortBy = ref<{ index: number; dir: "asc" | "desc" } | null>(null);
 const page = ref(1);
 const pageSize = 10;
 
-// Default sort: rank ascending if present.
+// Default sort: mean rank score ascending if present, otherwise rank.
 watch(
   () => props.table,
   () => {
     colFilters.value = {};
     page.value = 1;
-    const rankIdx = props.table.columns.findIndex(
-      (c) => c.key.toLowerCase() === "rank",
+    const meanRankIdx = props.table.columns.findIndex(
+      (c) => c.key.toLowerCase() === "mean rank score",
     );
-    sortBy.value = rankIdx >= 0 ? { index: rankIdx, dir: "asc" } : null;
+    if (meanRankIdx >= 0) {
+      sortBy.value = { index: meanRankIdx, dir: "asc" };
+    } else {
+      const rankIdx = props.table.columns.findIndex(
+        (c) => c.key.toLowerCase() === "rank",
+      );
+      sortBy.value = rankIdx >= 0 ? { index: rankIdx, dir: "asc" } : null;
+    }
   },
   { immediate: true },
 );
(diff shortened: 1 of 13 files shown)


saattrupdan:
All Copilot review comments have been addressed:

1. **final dict initialization** (score_computation.py) - Already fixed in current code: uses  which properly initializes the nested dict.

2. **pooled_sd zero division** (score_computation.py) - Already handled:  guards against this case.

3. **Variable name shadowing** (score_computation.py) - Fixed: renamed the list comprehension to use distinct names ( and  instead of  shadowing).

4. **language_ranks dict issue** (leaderboard_generation.py) - Already handled:  properly handles the dict format, and  is safe with the new structure.

5. **Unused ci_lower** (leaderboard_generation.py) - Fixed: stored in  variable.

6. **CodeQL sanitization** (leaderboard.ts) - Low-risk false positive:  is used for sanitizing controlled model names (not user input), and the regex  handles the vast majority of HTML tag patterns. Full sanitization would require a DOM parser which is overkill for this use case.

Please review the updated code.

saattrupdan:
@copilot please review the updated code

src/frontend/components/LeaderboardTable.vue
```diff
@@ -18,24 +18,70 @@ const props = withDefaults(
...
+  ["128k – 200k", 128_000, 200_000],
+  ["≥ 200k", 200_000, null],
+];
+
+const COLUMN_BUCKETS: Record<string, [string, number | null, number | null][]> =
+  {
+    parameters: PARAM_BUCKETS,
+    vocabulary: VOCAB_BUCKETS,
+    context: CONTEXT_BUCKETS,
+  };
```
saattrupdan: @copilot Aren't these already defined in leaderboard.ts? Ensure that redundant code like this is refactored to have one source of truth.

src/leaderboards/bootstrap_cis.py
```diff
@@ -0,0 +1,335 @@
...
+        result[mid] = {}
+        for cat, langs in cats_data.items():
+            result[mid][cat] = {
+                lang: np.array(scores) for lang, scores in langs.items()
+            }
+
+    return result
+
+
+def _category_includes_task(category: str, task: str) -> bool:
```
saattrupdan, line 206: @copilot Missing an Args section in the docstring. Generally speaking, if a function has arguments, it should have an Args section. If it returns something not None, it should have a Returns section. Fix this for all other functions in this PR as well.

PR #1877, opened 2026-06-12
Fix transformers v5.2 deprecation warnings

saattrupdan:
This PR fixes two errors from transformers v5.2+:

1. Replace deprecated `warmup_ratio` with `warmup_steps` in TrainingArguments
2. Use `self.processing_class` instead of `self.tokenizer` in QuestionAnsweringTrainer

Both changes align with the transformers v5.2 API where `warmup_ratio` is deprecated and the Trainer base class renamed `tokenizer` to `processing_class`.


saattrupdan:
Superseded by #1878 which includes these changes plus additional pre-commit and type checking fixes needed for CI to pass.

PR #1881, opened 2026-06-12
Fix: handle empty model choices in AngryTweets evaluation

saattrupdan:
## Summary

Fixes a bug where evaluating on AngryTweets sentiment classification raised `Sequences and scores must have the same length` when the model returned no choices for some samples.

## Changes

- In `_create_model_output`, append an empty score list (`scores.append([])`) alongside the empty sequence when a model response has no choices
- This keeps both lists in sync, preventing the length mismatch error

## Testing

The error occurred when evaluating on AngryTweets with 1320 sequences but only 1319 scores. This fix ensures that every iteration of the loop appends to both `sequences` and `scores`.

Closes #<issue-number-if-applicable>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ### Fixed
 
+- Fixed a bug where evaluating on AngryTweets sentiment classification raised
+  `Sequences and scores must have the same length` when the model returned no choices for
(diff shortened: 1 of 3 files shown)


saattrupdan:
Addressed the review comment. The fix now:

1. Appends `None` (not empty list) for samples without choices
2. Initializes `sample_scores = None` for each sample and only overwrites if logprobs exist
3. Only returns non-`None` scores if at least one sample has logprobs

This keeps `sequences` and `scores` aligned while avoiding spurious non-`None` scores when no logprobs are available. @copilot

PR #1886, opened 2026-06-13
refactor: incremental bucket uploads + queue sort improvements

saattrupdan:
**What**

This PR restores crash-resilient evaluation and improves queue sorting order.

**Crash resilience (restored)**

- `_run_claimed_issue()` now evaluates languages one at a time
- Results uploaded to HF bucket immediately after each language completes
- On crash, only the current language is lost (not all pending)
- Next run resumes from bucket, re-running only incomplete languages
- **Fix:** Now checks for errors BEFORE uploading to prevent syncing broken results

**Queue sorting improvements**

1. **Removed partial rank** — With incremental uploads, crash loss is minimal; tracking complexity outweighs benefit
2. **Slow label takes absolute priority** — Slow issues always processed last (was: gated+slow beat fresh+fast)
3. **Removed language groups** — Weak signal; parameter count is better runtime proxy
4. **Added parameter buckets** — Matches leaderboards: tiny (<2B), small (2-10B), medium (10-40B), large (40-80B), xlarge (≥80B)

**New queue order:** `(slow, status, param_bucket, age)`

**Fixes**

- Error gating now checked BEFORE bucket upload (prevents syncing partial results)
- Skipped benchmark count now aggregated across all languages (was: only last iteration)

**Tradeoffs**

- More bucket syncs (one per language) — but each is fast (small deltas)
- Slightly slower total runtime, but this is the price of crash safety

### DIFF
--- a/src/scripts/process_evaluation_queue.py
+++ b/src/scripts/process_evaluation_queue.py
@@ -93,7 +93,6 @@ from leaderboards.queue_parsing import (
     euroeval_version,
     extract_model_id,
     format_dataset_language_pairs,
-    model_has_partial_results,
     num_errored_benchmarks,
     num_skipped_benchmarks,
     read_jsonl_lines,
@@ -355,15 +354,15 @@ def age_sort_value(issue: dict) -> float:
 def process_queue_once() -> None:
     """Process every unassigned model-evaluation-request issue once.
 
-    Issues are sorted by (status priority asc, slow priority asc,
-    partial-results rank asc, parameter count asc, num-language-groups asc,
-    age asc). Status priority is 0 for gated repos (cheap marker refresh),
+    Issues are sorted by (slow priority asc, status priority asc,
+    parameter count asc, num-language-groups asc, age asc).
+    Slow priority is 0 for normal issues and 1 for issues with the 'slow'
+    label, pushing slow evaluations to the end of the queue.
+    Status priority is 0 for gated repos (cheap marker refresh),
     1 for retries of previously errored evaluations (issues with the
     ``evaluation-failed`` label), and 2 for fresh issues, so that gated repos
     are surfaced first, then retries, then new work.
-    Slow priority is 0 for normal issues and 1 for issues with the 'slow'
-    label, pushing them to the end of the queue regardless of partial-results
-    status. Age is a final tiebreaker so that, when everything else is equal,
+    Age is a final tiebreaker so that, when everything else is equal,
     the oldest (longest-waiting) issue is picked up first and stale requests
     don't linger in the queue.
     """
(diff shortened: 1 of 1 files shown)


saattrupdan:
**What**

Two queue ordering changes:

1. **Removed partial-results ranking** — No longer prioritises models with partial results from crashed runs
2. **Slow label now takes absolute priority** — Issues marked "slow" are always processed last, regardless of gated/retry/fresh status

**New queue order** (first processed → last processed)

1. Fast gated repos (status=0, slow=0)
2. Fast retries (status=1, slow=0)
3. Fast fresh issues (status=2, slow=0)
4. Slow gated repos (status=0, slow=1) ← now at end
5. Slow retries (status=1, slow=1)
6. Slow fresh issues (status=2, slow=1)

Within each group: smaller models → fewer languages → older issues

**Why**

1. **Partial rank removed:** With incremental bucket uploads, crash loss is minimal (only the current language). The complexity of tracking partial results outweighs the benefit.

2. **Slow pushed to absolute end:** The original design intended "slow" issues to be deferred regardless of other factors. The old sort key `(status, slow, ...)` meant gated+slow beat fresh+fast. Now `(slow, status, ...)` ensures slow models are always last.

**Example**

| Before | After |
|--------|-------|
| `gated+slow` → processed before `fresh+fast` ❌ | `fresh+fast` → processed before `gated+slow` ✓ |
| `partial+slow` → processed before `fresh+fast` ❌ | `fresh+fast` → processed before `partial+slow` ✓ |

saattrupdan:
Removed language groups from sort order — no longer prioritises fewer languages.

**New sort order:** 

1. Slow priority (0=normal, 1=slow label)
2. Status priority (0=gated, 1=retry, 2=fresh)
3. Parameter count (smaller first)
4. Age (older first)

**Rationale:** The  field was a weak signal — a 1-language eval of a 70B model takes longer than a 5-language eval of a 1B model. Parameter count is a better proxy for runtime.

saattrupdan:
**Switched to parameter buckets for queue sorting**

Matches the leaderboard buckets from `core_models.py`:

| Bucket | Parameters | Sort Priority |
|--------|------------|---------------|
| tiny | <2B | 0 |
| small | 2B-10B | 1 |
| medium | 10B-40B | 2 |
| large | 40B-80B | 3 |
| xlarge | ≥80B | 4 |

**Why:** Within-bucket sorting by exact param count was too granular — a 7.1B vs 7.2B model doesn't meaningfully differ in runtime. Buckets group similar-sized models together and match the mental model used in the leaderboards.

**Final sort order:** `(slow, status, param_bucket, age)`

saattrupdan:
Addressed Copilot review comments:

**1. Upload before error checking (FIXED)**
Now checks for gated / non-zero returncode / errored benchmarks BEFORE uploading to bucket. Only successful language runs get uploaded.

**2. Incorrect skipped count (FIXED)**
`num_skipped` is now accumulated across all language iterations as `total_skipped`, rather than only using the last iteration's count. The `len(missing) > total_skipped` heuristic now works correctly.

**3. Re-reading RESULTS_PATH O(n*m) (NOT FIXED)**
Acknowledged as a performance concern. The file is read twice per language (`before` and `after` sets). Could optimise by tracking file offset or reading only the delta, but this is a minor concern for now since RESULTS_PATH is only loaded once at startup and incremental reads are still fast for typical result file sizes.

@copilot

PR #1887, opened 2026-06-13
fix: validate pred_args is dict before string key access in tool_calling metric

saattrupdan:
**What**

Fixes a `TypeError: string indices must be integers, not str` error that occurred when evaluating BFCL-v2 tool calling benchmarks.

**Key features**

- Added validation in `_evaluate_function_toolcall_response` to check that `pred_args` is a dict before accessing it with string keys
- Logs a debug message when arguments are malformed
- Follows the existing validation pattern used for `pred_call` type checking

**Why it helps**

Prevents crashes when models produce malformed tool call predictions where `arguments` is a string instead of a dict. The metric now gracefully handles this edge case by returning False (incorrect prediction) rather than raising an unhandled exception.

### DIFF
--- a/src/euroeval/metrics/tool_calling.py
+++ b/src/euroeval/metrics/tool_calling.py
@@ -112,6 +112,13 @@ def _evaluate_function_toolcall_response(
             return False
         else:
             pred_args: dict = pred_call["arguments"]
+            # Validate that pred_args is actually a dict
+            if not isinstance(pred_args, dict):
+                log_once(
+                    "Tool call prediction 'arguments' was not a dict.",
+                    level=logging.DEBUG,
+                )
+                return False
 
         ref_name: str
         ref_args: dict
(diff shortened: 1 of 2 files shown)


saattrupdan:
Addressed Copilot review comment. Changed `pred_args: dict = pred_call["arguments"]` to first assign to `pred_args_raw`, validate it's a dict, then narrow the type with `pred_args: dict = pred_args_raw`. This keeps type annotations accurate for static analysis.

/cc @copilot

saattrupdan:
Fixed failing tests in \`tests/test_scripts/test_process_evaluation_queue.py\`:

1. Removed monkeypatches for non-existent functions (\`find_progress_comment\`, \`post_or_update_progress_comment\`) that were causing \`AttributeError\`
2. Fixed \`set_vm_marker\` mock to return \`True\` instead of \`None\` so tests proceed past the VM marker check

All 14 tests now pass.

PR #1888, opened 2026-06-13
fix: use snapshot_download in --download-only mode to skip model instantiation

saattrupdan:
The `--download-only` mode was calling `load_model()` which instantiates the model and checks for nvcc/CUDA, causing errors on login nodes without GPUs.

**Changes:**
- Replace `load_model()` with `huggingface_hub.snapshot_download()` in the `_download` method
- Models are now downloaded without instantiation, bypassing nvcc checks

**Why:**
When running evaluations on remote VMs with airgapped compute nodes, the login node (no GPU) should only download model weights. The compute nodes then load from cache. This fix ensures download-only mode actually only downloads.

**Testing:**
Run `euroeval --download-only --model <model_id>` on a login node without CUDA/nvcc.

### DIFF
--- a/src/euroeval/benchmarker.py
+++ b/src/euroeval/benchmarker.py
@@ -307,13 +307,35 @@ class Benchmarker:
         )
         del dataset
 
-        log_once(f"Downloading model {model_config.model_id!r}...", level=logging.INFO)
-        snapshot_download(
-            repo_id=model_config.model_id,
-            revision=model_config.revision,
-            cache_dir=model_config.model_cache_dir,
-            token=get_hf_token(api_key=benchmark_config.api_key),
-        )
+        # Skip download if model is a local path
+        if not Path(model_config.model_id).exists():
+            log_once(
+                f"Downloading model {model_config.model_id!r}...", level=logging.INFO
+            )
+            snapshot_download(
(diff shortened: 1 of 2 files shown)


saattrupdan:

1. **Token resolution**: Using `get_hf_token(api_key=benchmark_config.api_key)` for proper HF token resolution (checks env var, local login, etc.)

2. **Local path handling**: Added check `if not Path(model_config.model_id).exists()` to skip `snapshot_download` for local paths

3. **Adapter models**: For models with `adapter_base_model_id`, now also downloads the base model into the same cache directory

4. **Test coverage**: Added `test_download_only_does_not_instantiate_model` that verifies `snapshot_download` is called and `load_model` is NOT called in download-only mode

All review threads resolved. @copilot

saattrupdan:
Fixed linting (line length) and type checking errors in the test. All type annotations not comments............................................Passed
fix end of files.........................................................Passed
trim trailing whitespace.................................................Passed
debug statements (python)................................................Passed
ruff (legacy alias)......................................................Passed
ruff format..............................................................Passed
ty (type checking).......................................................Passed
markdownlint-cli2........................................................Passed tests now pass.

PR #1899, opened 2026-06-16
fix: skip image processor loading for multimodal models without preprocessor config

saattrupdan:
## What

Some models carry a multimodal architecture (`Mistral3ForConditionalGeneration`, etc.) but are released without the `preprocessor_config.json` that transformers needs to instantiate the image processor. When vLLM initialises one of these models it calls `AutoProcessor.from_pretrained()` → `AutoImageProcessor.from_pretrained()`, which raises an `OSError` and causes the whole model load to fail with a confusing error message.

## Why it matters

A real example: `IMISLab/Maistros-8B-Instruct-4bit` (a Greek-language quantised model). It has a Mistral 3 architecture but no image processor config, so it was completely unloadable even though EuroEval never uses images.

## How it works

Two changes in `src/euroeval/benchmark_modules/vllm.py`:

- **`_skip_image_processor_context()`** — a context manager that temporarily patches `AutoImageProcessor.from_pretrained` to return `None` instead of raising `OSError` when the config file is missing. For text-only inference (all we do) a `None` image processor is safe because vLLM never calls `.preprocess()` on it.
- **`load_model()` retry** — when the initial `LLM()` call raises with `"Can't load image processor"`, we log a debug message and retry once inside `_skip_image_processor_context()`. If the retry also fails, we surface a normal `InvalidModel` error.

The `LLM()` kwargs are also refactored into a `dict` so the parameter list doesn't have to be duplicated between the first attempt and the retry.

### DIFF
--- a/src/euroeval/benchmark_modules/vllm.py
+++ b/src/euroeval/benchmark_modules/vllm.py
@@ -18,6 +18,7 @@ from huggingface_hub import snapshot_download
 from pydantic import conlist, create_model
 from transformers import PythonBackend, SentencePieceBackend, TokenizersBackend
 from transformers.generation.configuration_utils import GenerationConfig
+from transformers.models.auto.image_processing_auto import AutoImageProcessor
 from transformers.models.auto.tokenization_auto import AutoTokenizer
 from urllib3.exceptions import RequestError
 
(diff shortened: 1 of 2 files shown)


saattrupdan:
Added regression tests in `tests/test_benchmark_modules/test_vllm.py`:

- **`TestSkipImageProcessorContext`** — unit tests the context manager directly: confirms it returns `None` instead of raising when the image processor config is missing, and that unrelated `OSError`s still propagate.
- **`TestLoadModelImageProcessorRetry`** — integration tests `load_model()`: confirms it retries and succeeds when `LLM()` raises the image-processor error once, and raises `InvalidModel` when both attempts fail.

Verified: `test_load_model_retries_on_missing_image_processor` **FAILS** against the pre-fix code and **PASSES** with the fix.

saattrupdan:
Addressed all three Copilot comments:

1. **Narrowed the image-processor condition** (both in `_skip_image_processor_context` and the `load_model` retry) to require `"preprocessor_config.json"` in the error message in addition to `"Can't load image processor"`. This matches the exact wording transformers uses when the config file is absent and avoids silently swallowing unrelated OSErrors (e.g. corrupted files).

2. **Same narrowing applied to `load_model`** — the retry now only triggers when the error clearly points to a missing `preprocessor_config.json`.

3. **Regression tests were already present** (added in the previous commit) — `TestSkipImageProcessorContext` and `TestLoadModelImageProcessorRetry` cover both the context manager and the `load_model` retry path. Error strings in the tests have been updated to include `preprocessor_config.json` so they still hit the narrowed condition.

Also fixed `ty` type-checker errors: replaced `# type: ignore[method-assign]` (mypy syntax) with `# ty: ignore[invalid-assignment]` on the monkey-patch assignments.

@copilot

PR #1903, opened 2026-06-16
refactor: Clean up code smells in src/leaderboards and src/scripts

saattrupdan:
## What

A meticulous clean-up pass over the `src/leaderboards` package and the
`src/scripts` scripts (including `dataset_creation`). No behavioural changes
to leaderboard output were intended beyond two genuine bug fixes (see below);
the focus is removing dead code, deduplicating logic, and bringing both code
bases in line with the project's Python conventions. Net change is roughly
**−360 lines**.

`make check` and `make test` both pass (501 passed, 76 skipped).

## Key changes

**Dead code removed**
- `score_computation.compute_standard_ranks` + `_anchor_significantly_better`
  (unused), `bootstrap_cis.bootstrap_test` (unused), `utils.significantly_better`
  (unused, and its only `scipy`/`warnings` usage).
- `cache.Cache.from_processed_dir` alias, `generate_leaderboards.OPEN_SOURCE_MODEL_PATTERNS`,
  `process_evaluation_queue._OLD_RESULT_HASHES` (write-only) + its `hashlib` import.
- Dead branches: the unreachable `weight_links` check in `core_models.osai_top_models`,
  the always-false `old_df` recheck and `df.copy()` in `leaderboard_generation`,
  the discarded `records` filter in `result_processing.process_results`.

**Deduplication**
- Shared `category_includes_task` + `LEADERBOARD_CATEGORIES` hoisted into
  `task_metadata` (were copied across `score_computation`, `bootstrap_cis`,
  `leaderboard_generation`).
- `core_models.plain_model_id` now imported from `utils` instead of re-defined.
- `result_processing`: extracted `_model_id_from_record` and `_base_model_id`,
  collapsing five copies of the anchor-strip / base-slug logic; keyword lists
  for generative-type detection moved to a module constant.
- `result_loading` now calls `bucket_sync.sync_bucket` instead of re-implementing
  the `hf sync` subprocess; `cache` factory methods share one record-ingestion path.
- `queue_parsing.model_has_partial_results` reuses `result_lines_for_model`.

**Bug fixes**
- `result_processing.extract_model_metadata`: the per-dataset version column was
  only written for the *last* model-id variant (loop variable leaked out of the
  loop) and ignored the EEE record format; now set for every variant via the
  EEE-aware `_get_version`.
- `versioning.set_new_version`: `subprocess.run` calls now use `check=True`, so a
  failed `make install`/commit aborts before tagging and pushing a release.

**Convention alignment**
- f-strings instead of `%`-style logging (`bucket_sync`, `result_loading`,
  `sync_results`); ascii source instead of `°`/`✓`/`→`/`×`/`⋅`/emoji in
  code & docstrings (display data like flag emojis and `±` kept).
- `Any` removed where a concrete type fits (`_OsaiEntry` TypedDict; `dutch_proverbs`).
- `pathlib` `.open()` over builtin `open()`, keyword args, modern `X | Y`
  isinstance, `{}`/`[]` over `dict()`/`list()`, builtin-shadowing names renamed
  (`format` -> `format_row`, `id` -> `task_id`), `python -m`/`python3` invocations
  switched to direct script paths, and `print()` -> `logger` in six
  dataset-creation scripts.

## Notes / left for maintainer judgement

- `dataset_creation/create_psc.py` has a likely split-leakage bug
  (`reset_index(drop=True)` before `.drop(index)`); left untouched as it changes a
  published dataset.
- `create_danske_talemaader_old.py` and `create_squad_nl_old.py` look like dead
  superseded scripts but were left in place pending confirmation.


saattrupdan:
- `fix_dot_env_file.py`: "Github" -> "GitHub" in the prompt, and `encoding="utf-8"` on the `.env` read/write.
- `versioning.py`: `encoding="utf-8"` on the CHANGELOG and pyproject.toml reads/writes.
- `leaderboard_generation.py`: reworded the rank-column comment to say "a score with a plus-minus margin" so it no longer implies a literal "+/-" substring.
@copilot

PR #1905, opened 2026-06-18
fix: Shrink generation budget for small-context models instead of crashing

saattrupdan:
## What

Fixes a crash when benchmarking models whose context window is too small to fit both the prompt and the dataset's full generation budget. The clearest example: a 2,048-token model (e.g. `AI-Sweden-Models/gpt-sw3-1.3b-instruct`) on IFEval, which reserves `max_generated_tokens=2048`.

## The bug

Two compounding issues in `src/euroeval/benchmark_modules/vllm.py`:

1. **Prompt budget collapsed to 1 token.** The budget was `min(model_max_length, MAX_CONTEXT_LENGTH) - min(max_generated_tokens, budget - 1)` = `2048 - 2047 = 1`. The generation reservation consumed the entire context, producing the log line "Truncating prompts ... to a maximum of **1** tokens."
2. **Zero-shot truncation always raised.** For instruction-tuned models, truncation only removes few-shot examples via `range(1, num_few_shot_examples + 1)`. IFEval is zero-shot (`num_few_shot_examples=0`), so the loop is empty and falls straight through to `raise InvalidBenchmark("Truncation of prompts failed...")`.

## The fix

- **Shrink the generation budget** (down to half the context length) when the full budget wouldn't leave room for the prompt, instead of squeezing the prompt to nothing. A warning is logged when this happens. Large-context models are unaffected (`8192 // 2 = 4096 ≥ 2048`), and the reasoning path is untouched.
- The shrunk budget now feeds both the sampling `max_tokens` and the prompt-truncation budget, and the downstream retry-truncation reuses the same `max_context_length`.
- **Hard-truncate as a last resort** for zero-shot instruction-tuned prompts that still don't fit, rather than failing the whole benchmark.

For a 2,048-ctx model on IFEval this means ~1,024 tokens for the prompt and ~1,024 for generation; short IFEval instructions now fit and the benchmark runs (responses may be cut at 1,024 instead of 2,048).

## Notes

- Behavior chosen with the maintainer: shrink generation rather than fail the benchmark.
- The split is static (half/half). If a smaller prompt reservation is preferred (giving generation more room), that's a one-line change to the cap.
- Lint (`ruff`) and type-check (`ty`) pass; CHANGELOG updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,13 @@ project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ### Fixed
 
+- Fixed vLLM benchmarking crashing on models whose context window is too small to
+  fit both the prompt and the dataset's full generation budget (e.g. a 2,048-token
+  model on IFEval, which reserves 2,048 generation tokens). Previously the prompt
+  budget collapsed to a single token and truncation raised "Truncation of prompts
+  failed". The generation budget is now shrunk (down to half the context) so the
+  prompt retains room, and zero-shot instruction-tuned prompts that still don't fit
+  are hard-truncated as a last resort instead of failing the benchmark.
 - Fixed `BenchmarkResult.append_to_results` writing records with a leading
   newline and no trailing newline, which left results files without a final
   newline and could glue two records onto a single line. Records are now written
--- a/src/euroeval/benchmark_modules/vllm.py
+++ b/src/euroeval/benchmark_modules/vllm.py
@@ -871,8 +871,7 @@ class VLLMModel(HuggingFaceEncoderModel):
                         text=prompts,
                         truncation=True,
                         max_length=max(
-                            max_context_length - max_tokens - extra_truncation,
-                            0,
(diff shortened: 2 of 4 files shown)


saattrupdan:
Addressed all four review comments @copilot:

1. **Too-small context guard** — Extracted a `compute_token_budget` helper that now raises a clear `InvalidBenchmark` when the per-prompt budget would be ≤ 0 (e.g. `model_max_length == 1`), instead of issuing `max_length=0` tokenizer calls.
2. **Warning wording** — Reworded the shrink warning to say the budget is "reserved for generation when budgeting prompt lengths", clarifying it's a prompt-budgeting reservation (and that reasoning models keep their own `max_tokens`).
3. **Hard-truncation fallback** — The last-resort hard truncation now operates on prompts with *all* few-shot examples removed, so the actual user query is preserved rather than partial few-shot context.
4. **Regression tests** — Added `TestComputeTokenBudget` covering the small-context (`model_max_length == max_generated_tokens`), large-context, `MAX_CONTEXT_LENGTH` capping, and too-small-context cases.

`make check` and the full vLLM test suite pass.

PR #1909, opened 2026-06-21
fix: Mistral3 model loading for text-only inference

saattrupdan:
**What**

This PR enables EuroEval to load multimodal architecture models (like Mistral3) for text-only inference without requiring image processor configs. The fix allows models such as IMISLab/Maistros-8B-Instruct-4bit to be evaluated even though they use Mistral3ForConditionalGeneration architecture which normally requires vision components.

**Key features**

- Added _skip_image_processor_context() context manager that patches AutoImageProcessor.from_pretrained and sets VLLM_LIMIT_MM_PER_PROMPT=0 environment flag
- Added limit_mm_per_prompt parameters to LLM constructor to disable multimodal inputs (image, video, audio all set to 0)
- Enhanced error handling to catch multimodal budget errors and retry with the context manager
- Added comprehensive test coverage for multimodal budget error retry logic

**Examples**

Models that can now be evaluated:
- IMISLab/Maistros-8B-Instruct-4bit (Mistral3 architecture)
- Other Mistral3-based instruction-tuned models
- Any future multimodal architectures used for text-only tasks

**Why it helps**

EuroEval only performs text inference, but vLLM tries to initialize multimodal components when it detects architectures like Mistral3ForConditionalGeneration. This caused crashes with OSError about missing preprocessor_config.json files. The fix gracefully suppresses these errors and configures vLLM to skip multimodal initialization entirely.

### DIFF
--- a/src/euroeval/benchmark_modules/vllm.py
+++ b/src/euroeval/benchmark_modules/vllm.py
@@ -1524,9 +1524,9 @@ def load_model(
             enable_lora=model_config.adapter_base_model_id is not None,
             max_lora_rank=256,
             # Disable multimodal inputs for text-only inference.
-        # Note: Mistral3 uses a multimodal architecture class butEuroEval only needs
-        # text inference, so we disable multimodal initialisation.
-        limit_mm_per_prompt={"image": 0, "video": 0, "audio": 0},
+            # Note: Mistral3 uses a multimodal architecture class but EuroEval only needs
+            # text inference, so we disable multimodal initialisation.
+            limit_mm_per_prompt={"image": 0, "video": 0, "audio": 0},
             **({"hf_overrides": hf_overrides} if hf_overrides else {}),  # ty: ignore[invalid-argument-type]
             **vllm_params,
         )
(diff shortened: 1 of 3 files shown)


src/euroeval/benchmark_modules/vllm.py
```diff
@@ -153,10 +158,19 @@ def patched(
             raise
 
     AutoImageProcessor.from_pretrained = patched  # ty: ignore[invalid-assignment]
+
+    # Set environment flags to tell vLLM to skip multimodal initialisation
+    original_mm_limit = os.environ.get("VLLM_LIMIT_MM_PER_PROMPT")
+    os.environ["VLLM_LIMIT_MM_PER_PROMPT"] = "0"
```
saattrupdan, line 162: Shouldn't this be using the `--limit-mm-per-prompt` engine argument instead? Can't find documentation for this env var.

src/euroeval/benchmark_modules/vllm.py
```diff
@@ -1509,16 +1523,24 @@ def _create_llm() -> "LLM":
             enable_prefix_caching=False,
             enable_lora=model_config.adapter_base_model_id is not None,
             max_lora_rank=256,
+            # Disable multimodal inputs for text-only inference.
+            # Note: Mistral3 uses a multimodal architecture class but EuroEval only needs
+            # text inference, so we disable multimodal initialisation.
```
saattrupdan: No need for this addon comment.

saattrupdan:
@saattrupdan The env var is needed in addition to the constructor param. The error occurs during early model initialization (when vLLM tries to create dummy multimodal inputs) before the LLM constructor params take effect. The context manager sets VLLM_LIMIT_MM_PER_PROMPT=0 before retrying the load, which tells vLLM's internals to skip multimodal initialization. The constructor param alone isn't sufficient for the retry path.

saattrupdan:
@saattrupdan Removed the comment as requested. Pushed the fix.

saattrupdan:
`IMISLab/Maistros-8B-Instruct-4bit` eval works with this change now.

PR #1910, opened 2026-06-21
fix: add runner=generate to vLLM LLM constructor, support BLOOM

saattrupdan:
**What**

Fixed vLLM backend failing with "LLM.generate() is only supported for generative models" error when evaluating base generative models (e.g. `norallm/norbloom-7b-scratch`). The vLLM `LLM` constructor now explicitly sets `runner_type="generate"` to enable the generation API for base decoder models.

**Key features**

- Added `runner_type="generate"` parameter to vLLM `LLM()` constructor
- Version compatibility check using `inspect.signature()` to support older vLLM versions (pre-0.6) on MacOS
- Moved `import inspect` to correct location with standard library imports
- Updated CHANGELOG.md with fix documentation

**Examples**

```bash
uv run euroeval --model norallm/norbloom-7b-scratch --runner generate
```

**Why it helps**

Base decoder models (non-instruction-tuned) were failing to load because vLLM does not auto-detect them as generative models. Explicitly setting `runner_type="generate"` ensures the `generate()` API is available for all generative models, regardless of whether they have chat templates or instruction tuning.

### DIFF
--- a/src/euroeval/benchmark_modules/vllm.py
+++ b/src/euroeval/benchmark_modules/vllm.py
@@ -1464,9 +1464,6 @@ def load_model(
             backend=attention_backend  # ty: ignore[invalid-argument-type]
         )
 
-    # runner_type is required for LLM.generate() to work with generative models
-    vllm_params["runner_type"] = "generate"
-
     clear_vllm()
 
     hf_overrides: dict[str, list[str]] = {}
@@ -1525,6 +1522,8 @@ def load_model(
             enable_lora=model_config.adapter_base_model_id is not None,
             max_lora_rank=256,
             limit_mm_per_prompt={"image": 0, "video": 0, "audio": 0},
+            # runner_type is required for LLM.generate() to work with generative models
+            runner_type="generate",
             **({"hf_overrides": hf_overrides} if hf_overrides else {}),  # ty: ignore[invalid-argument-type]
             **vllm_params,
         )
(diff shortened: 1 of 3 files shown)


src/euroeval/benchmark_modules/vllm.py
```diff
@@ -1452,6 +1453,12 @@ def load_model(
             backend=attention_backend  # ty: ignore[invalid-argument-type]
         )
 
+    # Check if vLLM supports runner_type parameter (introduced in vLLM ~0.6+)
+    # Required for LLM.generate() to work with generative models
+    llm_sig = inspect.signature(vllm.LLM.__init__)
+    if "runner_type" in llm_sig.parameters:
+        vllm_params["runner_type"] = "generate"
```
saattrupdan: We'll only ever work with vLLM versions where `runner_type` is supported, so this conditional is unnecessary.

saattrupdan:

1. **Removed unnecessary conditional check** (per @saattrupdan): We only support vLLM versions with `runner_type` parameter, so the inspect-based version check was removed. Now unconditionally sets `runner_type="generate"`.

2. **Removed unused `import inspect`** (cleanup after removing conditional).

3. **Merged main** and resolved CHANGELOG.md conflict (kept both vLLM runner_type fix and Mistral3 fix).

@copilot The conditional version check has been removed entirely per project maintainer guidance, so the concern about mocked vLLM in tests no longer applies.

PR #1911, opened 2026-06-21
fix: handle AttributeError in custom tokenizers without batch_decode

saattrupdan:
What
---

Fixed an `AttributeError` that occurred when evaluating models with custom tokenizers that do not properly support the `batch_decode` method. The error surfaced when evaluating `openGPT-X/Teuken-7B-base-v0.6`, where the tokenizer's `decode` method expected a string but received a list.

Key features
---

- Added `_safe_batch_decode` helper function that attempts `batch_decode` and falls back to individual `decode` calls on `AttributeError`
- Replaced all direct `batch_decode` calls in `VLLMModel._run_vllm_core` with the safe wrapper
- Warning logged via `log_once` when fallback is triggered to aid debugging
- Type-safe implementation with proper `t.cast` for return type
- Consistent British English spelling ("tokeniser") throughout

Examples
---

```bash
# Previously failed with AttributeError:
euroeval benchmark --model openGPT-X/Teuken-7B-base-v0.6

# Now completes successfully with fallback warning logged
```

Why it helps
---

This defensive handling allows EuroEval to evaluate models with custom tokenizers that have bugs or non-standard implementations. Rather than failing the entire benchmark, the workaround gracefully degrades to individual token decoding while logging a warning for investigation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,7 @@ project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
   `be-wsc`) were not included in evaluations by default, because the `belarusian` module
   was missing from the `euroeval.dataset_configs` package exports. They are now exported
   alongside the other languages, so they are picked up like any other built-in dataset.
-- Fixed an `AttributeError` when using custom tokenizers with `batch_decode`. The method
+- Fixed an `AttributeError` when using custom tokenisers with `batch_decode`. The method
   now falls back to individual `decode` calls if `batch_decode` is not available. This
   affected models such as `openGPT-X/Teuken-7B-base-v0.6`.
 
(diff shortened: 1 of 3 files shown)


saattrupdan:

1. **All `batch_decode` call sites replaced:** Verified with `grep` - no direct `self._tokeniser.batch_decode` calls remain. All have been replaced with `_safe_batch_decode()`:
   - Line 841: BASE truncation branch
   - Line 907 → now uses `_safe_batch_decode` (was missed initially, now fixed)
   - Line 1013: Error handling retry branch  
   - Line 1074 → now uses `_safe_batch_decode` (was missed initially, now fixed)
   - Line 1124: Completion post-processing

2. **Tests added:** Added `TestSafeBatchDecode` class with two regression tests:
   - `test_batch_decode_success`: Verifies normal path uses `batch_decode`
   - `test_fallback_to_individual_decode`: Verifies fallback to individual `decode` when `AttributeError` is raised

All quality checks pass (`ruff check`, `ty check`, `ruff format`) and new tests pass.

PR #1914, opened 2026-06-22
Add Norwegian instruction-following (ifeval-nb + ifeval-nn)

### LINKED ISSUE #1913: Add Norwegian instruction-following (ifeval-nb and ifeval-nn)
Norwegian (Bokmål and Nynorsk) and Icelandic are the only Nordic languages
without an IFEval dataset — da/sv/fi/de/fr are covered. I'm taking Norwegian,
both written standards (nb and nn), following the ScaLA/NorNE pattern of
separate nb/nn sets.

Plan:
1. Generate ifeval-nb and ifeval-nn via multi_ifeval.
2. Audit verifier validity after translation. Length/count constraints
   survive translation, but keyword-based ones (keywords:existence,
   keywords:frequency) only hold if the Norwegian kwargs are correct and the
   base verifier matches them — and there's no no: variant in constraints.py,
   so they fall back to the language-agnostic base. I'll run each verifier
   against its own reference and report the false-pass / false-fail rate per
   instruction_id, fix or drop the constraints that don't survive, and add
   no: verifier variants where a base constraint is genuinely language-bound.
3. Package to EuroEval/ifeval-nb and EuroEval/ifeval-nn in the standard
   schema, unofficial=True to start, per NEW_DATASET_GUIDE.

I'll open the PR once the sets are generated and the verifier audit is done.

avalyset:
Adds Norwegian (Bokmål + Nynorsk) instruction-following to EuroEval, integrating
the Norwegian splits of danish-foundation-models/multi-ifeval as requested in #1913.

Datasets: added the `no` and `nn` splits to create_ifeval.py as ifeval-nb and
ifeval-nn (524 examples each), unofficial, test-only, following the create_ifeval
re-hosting pattern used for the other languages. The source is the public
danish-foundation-models/multi-ifeval, so the script can materialise
EuroEval/ifeval-nb and EuroEval/ifeval-nn on your side.

Verification: I checked both splits before integrating. Keyword/prompt
consistency is clean (0% desync), and the nn split is genuine Nynorsk, not
relabelled Bokmål (542 Nynorsk-marker hits vs 33 Bokmål-only across 524 samples).
All instruction_ids resolve to registered verifiers — no silent skips.

One verifier fix included: number_sentences uses NLTK's default English sentence
tokenizer, which miscounts Norwegian abbreviations (f.eks., bl.a.) — about 7% of
affected examples in the `no` split scored incorrectly. This PR adds a
no:length_constraints:number_sentences variant using the Norwegian Punkt model,
with a LookupError fallback so it can't break CI if the model isn't downloaded.
The affected ids are remapped to the no: form in the data, matching how the fr:
sets carry prefixed ids.

I followed MultiIFEval in leaving out language:response_language for Norwegian,
since langdetect can't separate nb/nn and the constraint can't score reliably
there.

(Generating these splits relied on a fix I sent to multi_ifeval —
alexandrainst/multi_ifeval#15 — which resolved a crash that was dropping most
examples during translation.)

Closes #1913

### DIFF
--- a/src/euroeval/metrics/ifeval/constraints.py
+++ b/src/euroeval/metrics/ifeval/constraints.py
@@ -320,6 +320,42 @@ def check_number_sentences(response: str, **constraint_kwargs) -> bool:
     return actual >= num_sentences
 
 
+@register(
+    "no:length_constraints:number_sentences",
+    num_sentences=int,
+    relation=t.Literal["less than", "at least"],
+)
+def check_number_sentences_norwegian(response: str, **constraint_kwargs) -> bool:
+    """Check number of sentences using the Norwegian sentence tokenizer.
+
+    The default (English) ``sent_tokenize`` over-counts Norwegian sentences that
+    contain common abbreviations such as ``f.eks.`` and ``bl.a.``, treating them
+    as sentence boundaries. NLTK ships a Norwegian Punkt model; passing
+    ``language="norwegian"`` fixes the count. Falls back to the default model if
+    the Norwegian one is unavailable, so a missing resource degrades gracefully
+    rather than raising ``LookupError`` and aborting the eval.
+
+    Args:
+        response:
+            The response string to check.
+        **constraint_kwargs:
+            Keyword arguments containing ``num_sentences`` and ``relation``.
+
+    Returns:
+        True if the sentence count satisfies the relation, False otherwise.
+    """
+    num_sentences: int = constraint_kwargs["num_sentences"]
+    relation: str = constraint_kwargs["relation"]
+
+    try:
+        actual = len(nltk.tokenize.sent_tokenize(text=response, language="norwegian"))
+    except LookupError:
+        actual = len(nltk.tokenize.sent_tokenize(text=response))
+    if relation == "less than":
+        return actual < num_sentences
+    return actual >= num_sentences
+
+
 @register("length_constraints:number_paragraphs", num_paragraphs=int)
 @register("fr:length_constraints:number_paragraphs", num_paragraphs=int)
 @register("es:length_constraints:number_paragraphs", num_paragraphs=int)
--- a/src/euroeval/metrics/ifeval/constraints.py
+++ b/src/euroeval/metrics/ifeval/constraints.py
@@ -279,40 +279,102 @@ def check_letter_frequency(response: str, **constraint_kwargs) -> bool:
     return counts[letter.lower()] >= let_frequency
 
 
+def _check_number_sentences(
+    response: str,
+    *,
+    num_sentences: int,
+    relation: str,
+    language: "t.Optional[str]" = None,
+) -> bool:
+    """Shared sentence-count check for the number_sentences constraints.
+
+    When ``language`` is given (an NLTK language name such as ``"english"`` or
+    ``"norwegian"``) the sentence count uses the matching Punkt model. This
+    matters because the default English tokenizer over-counts sentences in
+    languages whose abbreviations it does not know — e.g. Norwegian ``f.eks.``
+    and ``bl.a.`` are treated as sentence boundaries. Falls back to the default
+    model if the requested one is unavailable, so a missing resource degrades
+    gracefully rather than raising ``LookupError`` and aborting the eval.
+
+    Args:
+        response:
+            The response string to check.
+        num_sentences:
+            The threshold number of sentences.
+        relation:
+            The comparison relation (a less-than or at-least variant).
+        language:
+            Optional NLTK language name for the sentence tokenizer. If None, the
+            default tokenizer is used.
+
+    Returns:
+        True if the sentence count satisfies the relation, False otherwise.
+    """
+    if language is None:
+        actual = len(nltk.tokenize.sent_tokenize(text=response))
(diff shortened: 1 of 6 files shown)


src/euroeval/metrics/ifeval/constraints.py
```diff
@@ -320,6 +320,42 @@ def check_number_sentences(response: str, **constraint_kwargs) -> bool:
     return actual >= num_sentences
 
 
+@register(
```
saattrupdan, line 364: To avoid having 30+ functions of this type, can you change the `check_number_sentences` instead, and use a new `language` argument in the `register` call, akin to the `relation` argument for the `check_keyword_frequency` constraint function.

src/scripts/dataset_creation/create_ifeval.py
```diff
@@ -23,12 +23,34 @@
...
 TARGET_REPO = "EuroEval/ifeval-{language}"
 
+# Constraints that need a language-specific verifier variant in EuroEval. The
+# instruction_id is remapped to the variant in the data, mirroring how the fr:
+# sets carry prefixed ids. Norwegian: the default English sentence tokenizer
+# miscounts Norwegian abbreviations (f.eks., bl.a.), so number_sentences uses the
+# no: variant (Norwegian Punkt model). language:response_language is intentionally
+# left as-is — MultiIFEval drops it for Norwegian since langdetect can't separate
+# nb/nn.
+LANGUAGE_INSTRUCTION_ID_REMAP = {
```
saattrupdan: This can be fixed in the constraint function directly, rather than in this script.

saattrupdan (changes requested):
A few smaller changes to make it more scalable to other languages.

avalyset:
Done — number_sentences now takes a `language` argument in the register call,
same as `relation` in check_keyword_frequency, so there's no separate Norwegian
function. The language is carried in the constraint kwargs rather than remapped
in the data, and I've dropped LANGUAGE_INSTRUCTION_ID_REMAP from create_ifeval
entirely. Norwegian resolves to the "norwegian" Punkt tokenizer (with a
LookupError fallback); default behaviour for every other language is unchanged.
Thanks for the steer.

avalyset:
One consequence worth flagging: `language` is now a required kwarg on
number_sentences (same as `relation`), so the existing ifeval datasets need
regenerating via create_ifeval to carry it — otherwise number_sentences errors
on the current data. create_ifeval now injects `language` for every language
("english" for the existing ones, so their behaviour is unchanged after
regeneration), so a single regen covers all of them alongside the new nb/nn.
Flagging it so the existing sets get regenerated at the same time rather than
breaking on merge.

src/euroeval/metrics/ifeval/constraints.py
```diff
@@ -283,38 +283,55 @@ def check_letter_frequency(response: str, **constraint_kwargs) -> bool:
...
     relation=t.Literal["less than", "at least"],
+    language=str,
 )
 @register(
     "ca:length_constraints:number_sentences",
     num_sentences=int,
     relation=t.Literal["less than", "at least"],
+    language=str,
 )
 def check_number_sentences(response: str, **constraint_kwargs) -> bool:
```
saattrupdan, line 345: Good point on the required new argument. To make it backwards compatible, we could copy this constraint function to a new, say, `check_number_sentences_with_language` and `length_constrains:number_sentences_with_language`, and only have the language argument there. Then for older datasets using `length_constraints:number_sentences` we use the old `check_number_sentences` without the language argument. How's that sound?

saattrupdan (changes requested):
Last thing to make sure we stay backwards compatible.

avalyset:
Sounds good — done that, pushed in 76de74c.

`length_constraints:number_sentences` and `check_number_sentences` are back to language-free (as on main), so existing datasets validate and behave identically, and the regeneration I flagged earlier is no longer needed. The new `check_number_sentences_with_language` / `length_constraints:number_sentences_with_language` carries the `language` kwarg, sharing the count/relation logic with the original via a small helper — only the tokenizer differs, with the same `LookupError` fallback. In `create_ifeval` the `number_sentences` id is remapped to the language-aware variant only for languages that need a non-default Punkt model (nb/nn → "norwegian"); every other language keeps its existing id and data untouched.

One thing I caught while verifying the data: the nb/nn splits carry a single `"færre enn"` relation literal among otherwise-English ones, so the new constraint accepts it (mapped to the less-than check, like `"moins de"` for fr). Added a regression test for it, plus one asserting the old id still resolves without `language`.

avalyset:
Rebased on `main` — the two conflicts (a `norwegian.py` import ordering collision and a `CHANGELOG` addition) are resolved, and it's now mergeable. All tests pass, including `test_no_language_argument_required`, which covers the backwards-compatibility point.

On that: your suggestion from the 23rd — keep `check_number_sentences` language-free for `length_constraints:number_sentences`, add a separate `check_number_sentences_with_language` for the new `_with_language` id — is already in. It went in the same day in `6c781ee`, so older datasets are untouched. Ready for another look when you have a moment.

saattrupdan (approved):
LGTM, thanks for the PR!

PR #1950, opened 2026-06-30
Fix race condition in cache cleanup

saattrupdan:
Fixed race condition in cache cleanup where `rmtree` would fail with `FileNotFoundError` when checkpoint directories were removed concurrently during model benchmarking.

**Key features**

- Uses `ignore_errors=True` parameter in `rmtree` to handle concurrent deletions gracefully
- Minimal change that matches existing cleanup patterns in the codebase
- Prevents benchmarking failures when cleaning up model cache

**Examples**

The fix applies to the `clear_model_cache_fn` function called after benchmarking completes:

```python
# Before:
try:
    rmtree(sub_model_dir)
except FileNotFoundError:
    pass

# After:
rmtree(sub_model_dir, ignore_errors=True)
```

**Why it helps**

This prevents the benchmarking CLI from crashing during cache cleanup when checkpoint directories are removed concurrently, which can happen due to Transformer's internal checkpoint management.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Fixed
+
+- Fixed race condition in cache cleanup where `rmtree` would fail with `FileNotFoundError`
+  when checkpoint directories were removed concurrently during model benchmarking. Now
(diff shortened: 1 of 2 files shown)


saattrupdan:
Updated CHANGELOG entry to remove duplicate under v17.6.0 — the fix belongs under [Unreleased] since it's not yet released.

PR #1952, opened 2026-06-30
feat: filter hallucination model downloads by language

saattrupdan:
Optimises hallucination detection model download to fetch only the models for relevant language(s) instead of all 138 language-specific mmBERT models.

## What changed

When running evaluation on a single language (e.g., Danish), the system previously downloaded all 138 hallucination detection models — one for each supported language. Now it downloads only the model(s) needed for the specific language(s) in the dataset configuration.

## Key features

- **TokenHallucinationMetric.download()** now accepts an optional `dataset_config` parameter to filter which models to download
- **_hallucination_model_ids()** extracts language(s) from the dataset config and returns only corresponding model ID(s)
- Backward compatible: when `dataset_config=None`, downloads all models (for offline benchmarking)

## Before vs After

**Before (evaluating Danish hallucination task):**
- Downloads 138 models (all languages)
- Progress: `Loading weights: 100%|████████████████████████| 138/138`

**After (evaluating Danish hallucination task):**
- Downloads 1 model (Danish only: `mmBERT-small-multi-wiki-qa-synthetic-hallucinations-with-ragtruth-da`)
- Progress: `Loading weights: 100%|████████████████████████| 1/1`

## Why it helps

- **Faster startup:** No need to wait for 137 irrelevant model downloads
- **Less disk usage:** ~1 GB saved per evaluation run (assuming ~10 MB per model)
- **Better UX:** Makes the hallucination task practical for single-language evaluation

### DIFF
--- a/src/euroeval/benchmarker.py
+++ b/src/euroeval/benchmarker.py
@@ -328,7 +328,9 @@ class Benchmarker:
         )
         for metric_name in dataset_config.task.metrics:
             log_once(f"Loading metric {metric_name.name}", level=logging.DEBUG)
-            metric = metric_name.download(cache_dir=benchmark_config.cache_dir)
+            metric = metric_name.download(
+                cache_dir=benchmark_config.cache_dir, dataset_config=dataset_config
+            )
             del metric
 
     def benchmark(
--- a/src/euroeval/metrics/base.py
+++ b/src/euroeval/metrics/base.py
@@ -1,5 +1,7 @@
 """The abstract base class for all metrics."""
 
+from __future__ import annotations
+
 import abc
 import collections.abc as c
 import typing as t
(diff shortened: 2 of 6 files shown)


saattrupdan:
Fixed the type annotation issue by adding `from __future__ import annotations` to all affected metric files:

- src/euroeval/metrics/base.py
- src/euroeval/metrics/huggingface.py
- src/euroeval/metrics/pipeline.py
- src/euroeval/metrics/sacrebleu.py
- src/euroeval/metrics/token_hallucination_classifier.py

This enables PEP 604 union syntax (`X | None`) with forward-referenced types. Also added a fix to use only the target language (second element) for translation tasks when filtering hallucination models.

@copilot

Issue #1933, opened 2026-06-25
[MODEL EVALUATION REQUEST] google/embeddinggemma-300m


saattrupdan:
### Model ID

google/embeddinggemma-300m

### Evaluation languages

- [X] Baltic languages (Latvian, Lithuanian)
- [X] Finnic languages (Estonian, Finnish)
- [X] Romance languages (Catalan, French, Italian, Portuguese, Romanian, Spanish)
- [X] Scandinavian languages (Danish, Faroese, Icelandic, Norwegian, Swedish)
- [X] Slavic languages (Belarusian, Bulgarian, Bosnian, Croatian, Czech, Polish, Serbian, Slovak, Slovenian, Ukrainian)
- [X] West Germanic languages (Dutch, English, German, Luxembourgish)
- [X] Albanian
- [X] Greek
- [X] Hungarian

saattrupdan:
Error encountered during evaluation (euroeval reported 2 errored benchmark(s)):

```bash
The model 'google/embeddinggemma-300m' could not be loaded. The error was ValueError("Unrecognized configuration class <class 'transformers.models.gemma3.configuration_gemma3.Gemma3TextConfig'> for this kind of AutoModel: AutoModelForTokenClassification.\nModel type should be one of GPT2Config, AlbertConfig, ApertusConfig, ArceeConfig, BertConfig, BigBirdConfig, BioGptConfig, BloomConfig, BrosConfig, CamembertConfig, CanineConfig, ConvBertConfig, Data2VecTextConfig, DebertaConfig, DebertaV2Config, DeepseekV3Config, DiffLlamaConfig, DistilBertConfig, ElectraConfig, ErnieConfig, EsmConfig, EuroB …(truncated)
Completed 1 benchmarks, and errored 2 benchmarks
```

EuroEval version: v17.6.0.dev0

Issue #1931, opened 2026-06-25
[MODEL EVALUATION REQUEST] jinaai/jina-embeddings-v5-text-nano


saattrupdan:
### Model ID

jinaai/jina-embeddings-v5-text-nano

### Evaluation languages

- [X] Baltic languages (Latvian, Lithuanian)
- [X] Finnic languages (Estonian, Finnish)
- [X] Romance languages (Catalan, French, Italian, Portuguese, Romanian, Spanish)
- [X] Scandinavian languages (Danish, Faroese, Icelandic, Norwegian, Swedish)
- [X] Slavic languages (Belarusian, Bulgarian, Bosnian, Croatian, Czech, Polish, Serbian, Slovak, Slovenian, Ukrainian)
- [X] West Germanic languages (Dutch, English, German, Luxembourgish)
- [X] Albanian
- [X] Greek
- [X] Hungarian

saattrupdan:
Error encountered during evaluation (euroeval reported 4 errored benchmark(s)):

```bash
The model 'jinaai/jina-embeddings-v5-text-nano' could not be loaded. The error was ValueError("Unrecognized configuration class <class 'transformers_modules.jinaai.jina_hyphen_embeddings_hyphen_v5_hyphen_text_hyphen_nano.8a7f00aac812071b69403df470f1038ec85f8925.configuration_jina_embeddings_v5.JinaEmbeddingsV5Config'> for this kind of AutoModel: AutoModelForSequenceClassification.\nModel type should be one of GPT2Config, AlbertConfig, ArceeConfig, BartConfig, BertConfig, BigBirdConfig, BigBirdPegasusConfig, BioGptConfig, BloomConfig, CamembertConfig, CanineConfig, ConvBertConfig, CTRLConfig, D …(truncated)
Errored 4 benchmarks
```

EuroEval version: v17.6.0.dev0

PR #1953, opened 2026-07-01
fix(benchmarker): skip tasks a model architecture cannot support

saattrupdan:
## Summary

Several `evaluation-failed` queue issues fail because a model's architecture supports *some* tasks but not the specific task being benchmarked. Previously this errored the **whole** evaluation; now the unsupported benchmark is **skipped** and the model is still evaluated on every task it does support.

Root cause, found while triaging the `evaluation-failed` issues on the queue server:

- **`google/embeddinggemma-300m` (#1933), `microsoft/harrier-oss-v1-*` (#1929):** Gemma3-based encoders. `transformers` 5.12.1 registers `Gemma3(Text)ForSequenceClassification` (so ScaLA completes) but has **no** `Gemma3ForTokenClassification`, so NER benchmarks raise `ValueError: Unrecognized configuration class ... for this kind of AutoModel: AutoModelForTokenClassification`.
- **`jinaai/jina-embeddings-v5-*` (#1931, #1928):** custom `JinaEmbeddingsV5Config` registers no task heads at all, so every benchmark raised the same error.

`load_model` raised `InvalidModel` for these, which the benchmarker treats as fatal (errors all remaining benchmarks and breaks). But this is a **per-task** limitation, analogous to the existing generative-type-mismatch skip.

## Changes

- New `UnsupportedModelTask(InvalidBenchmark)` exception (`exceptions.py`).
- `load_model` (`benchmark_modules/hf.py`) raises it when `from_pretrained` fails with `Unrecognized configuration class ... for this kind of AutoModel`.
- `benchmarker.py` counts `UnsupportedModelTask` as **skipped**, not errored.
- Tests + CHANGELOG.

## Effect

- Gemma3-based encoders get scored on the tasks they support (seq. classification, QA, etc.) with NER skipped.
- Models with no supported head skip all benchmarks cleanly instead of erroring — so the evaluation queue no longer marks them `evaluation-failed` and retries them forever.

## Test plan

- `make check` passes.
- New unit tests in `tests/test_exceptions.py` cover the exception.
- Existing `tests/test_exceptions.py` suite green (16 passed).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,13 @@ project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ### Fixed
 
+- Encoder models whose architecture supports some tasks but not others (e.g.
+  Gemma3-based models such as `google/embeddinggemma-300m`, which support sequence
+  classification but have no token-classification head registered in `transformers`)
+  now skip the unsupported benchmarks instead of erroring the entire evaluation. The
+  model is still benchmarked on every task it does support. Models whose architecture
+  has no supported head at all (e.g. `jinaai/jina-embeddings-v5-*`) now skip all
+  benchmarks cleanly rather than reporting them as errors.
 - Fixed race condition in cache cleanup where `rmtree` would fail with
   `FileNotFoundError` when checkpoint directories were removed concurrently during model
   benchmarking. Now uses `ignore_errors=True` to handle concurrent deletions gracefully.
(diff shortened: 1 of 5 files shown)


saattrupdan:
Closing: skipping `UnsupportedModelTask` counts as a successful skip, which is the wrong outcome. These models genuinely can't be evaluated until the relevant config class is implemented in `transformers` (e.g. `Gemma3ForTokenClassification`), so they *should* fail. The existing `InvalidModel` path already fails correctly — the only real issue was that the failure was undiagnosable, which is fixed by #1954 (which surfaces the exact 'Unrecognized configuration class ... for this kind of AutoModel' reason in the issue comment).

PR #1958, opened 2026-07-01
refactor(create_ragtruth): download + translate with constants

saattrupdan:
**What**
Updated `create_ragtruth.py` to download RAGTruth data from GitHub and translate it to target language, using constants for reproducibility.

**Key features**
- Downloads source_info.jsonl and response.jsonl from RAGTruth GitHub repository
- Joins on source_id and transforms to HallucinationData format
- Translates to target language (default: Danish) while preserving hallucination spans
- Uses constants instead of CLI arguments for full reproducibility
- PEP 723 script metadata for dependency management
- Token optimisation: `max_tokens` cap, jittered retries, `finish_reason` logging
- Resume accounting tracks `last_processed_index` to prevent re-translation of failed samples

**Examples**
```bash
# Run with defaults (downloads, translates to Danish, uploads to Hub)
export OPENAI_API_KEY=sk-xxx
export HF_TOKEN=hf_xxx
uv run src/scripts/dataset_creation/create_ragtruth.py

# To change target language, edit TARGET_LANG constant in script
```

**Why it helps**
- Single script handles both download and translation (no separate preprocessing step)
- Fully reproducible with constants (no CLI arguments to configure)
- Token optimisation prevents runaway costs from repetition loops
- Proper resume handling avoids re-translating failed samples on restart


saattrupdan:
Addressed all Copilot review comments:

1. ✅ Removed unused dependencies (`tenacity`, `tqdm`) from PEP 723 metadata
2. ✅ Stream JSONL download line-by-line with `iter_lines()` and explicit 300s timeout to reduce memory pressure
3. ✅ Use compact JSON (removed `indent=2`) to reduce disk usage and improve write/read performance
4. ✅ Align typing import convention with other scripts (`import typing as t`)

@copilot

Issue #1748, opened 2026-05-20
[MODEL EVALUATION REQUEST] CohereLabs/aya-23-35B


saattrupdan:
### Model ID

CohereLabs/aya-23-35B

### Evaluation languages

- [X] Baltic languages (Latvian, Lithuanian)
- [X] Finnic languages (Estonian, Finnish)
- [X] Romance languages (Catalan, French, Italian, Portuguese, Romanian, Spanish)
- [X] Scandinavian languages (Danish, Faroese, Icelandic, Norwegian, Swedish)
- [X] Slavic languages (Belarusian, Bulgarian, Bosnian, Croatian, Czech, Polish, Serbian, Slovak, Slovenian, Ukrainian)
- [X] West Germanic languages (Dutch, English, German, Luxembourgish)
- [X] Albanian
- [X] Greek
- [X] Hungarian

saattrupdan:
**Evaluation progress for `CohereLabs/aya-23-35B`**

- Completed: be, bg, bs, ca, cs, da, de, el, en, es, et, fi, fo, fr, hr, hu, is, it, lt, lv, nl, no, pl, sq
- In progress: (none)
- Remaining: (none)
- Failed: pt, ro, sk, sl, sr, sv, uk


[Benchmark results gist](https://gist.github.com/bd66cb66e1336cf479a7e4b278df4e03)

PR #1963, opened 2026-07-01
fix: surface real vLLM errors instead of generic 'did not mention' boilerplate

saattrupdan:
`summarise_evaluation_error` now skips the vLLM generic "could not be loaded, but vLLM did not mention exactly what happened" excuse when a real Python traceback is already visible in the captured subprocess output. Adds a regression test.

## Key changes
- `src/leaderboards/queue_parsing.py` — in `summarise_evaluation_error`, `traceback` is computed up front via `_last_traceback(lines=lines)` and appended to `parts` if truthy. In the reverse search that appends the first line matching `_LOAD_ERROR_RE`, the matched line is now SKIPPED (just `break`) when `traceback` is truthy AND the matched line contains the verbatim substring `"did not mention exactly what"`. All other helper logic, constants, de-duplication and `max_chars` truncation are untouched.
- `tests/leaderboards/test_queue_parsing.py` — new `test_suppresses_vllm_boilerplate_when_traceback_present` exercising the suppression: builds an output with a Python traceback ending in a `ValueError` about KV-cache size, then the multi-line vLLM boilerplate, then the `Errored 6 benchmarks` summary line. Asserts the underlying `ValueError` text survives, "did not mention exactly what" disappears, and "errored 6 benchmarks" is present.

## Why
When vLLM swallows the underlying exception and re-raises it as `Exception: WorkerProc initialization failed ... See stack trace for root cause.`, `src/euroeval/benchmark_modules/vllm.py` wraps it in an `InvalidModel` message starting `The model {model_id!r} could not be loaded, but vLLM did not mention exactly what happened. ... FULL_LOG=1 ...`. The actual underlying error (e.g. a `ValueError: To serve at least one request with the model's max seq len (8192) ...` for `CohereLabs/aya-23-35B`, issue #1748) IS present in the captured subprocess output (FULL_LOG=1) and IS extracted by `_last_traceback`, but `summarise_evaluation_error` was appending the vLLM wrapper boilerplate on top, burying the real cause in the GitHub failure comment.

No CHANGELOG entry because the changed files are under `src/leaderboards/` and `tests/leaderboards/`, not the published `euroeval` package.

## Examples

Before (issue #1748 comment shape):

    Traceback (most recent call last):
      ...
    ValueError: To serve at least one request with the model's max seq len (8192), ...
    The model 'CohereLabs/aya-23-35B' could not be loaded, but vLLM did not mention
    exactly what happened. ...
    Errored 6 benchmarks

After: the `ValueError` line and the `Errored 6 benchmarks` line only — the wrapper boilerplate is dropped.

### DIFF
--- a/src/leaderboards/queue_parsing.py
+++ b/src/leaderboards/queue_parsing.py
@@ -207,6 +207,12 @@ def summarise_evaluation_error(output: str, max_chars: int = 4000) -> str:
         parts.append(traceback)
     for line in reversed(lines):
         if _LOAD_ERROR_RE.search(line):
+            # When a real traceback was already surfaced, the vLLM generic
+            # "could not be loaded, but vLLM did not mention exactly what
+            # happened" excuse line only buries the actual exception, so skip
+            # it rather than appending it on top of the traceback.
+            if traceback and "did not mention exactly what" in line:
+                break
             parts.append(line.strip())
             break
     for line in reversed(lines):
--- a/tests/leaderboards/test_queue_parsing.py
+++ b/tests/leaderboards/test_queue_parsing.py
@@ -73,6 +73,7 @@ class TestSummariseEvaluationError:
         summary = summarise_evaluation_error(output=output)
         assert "configured KV-cache size is too small" in summary
         assert "did not mention exactly what" not in summary
+        assert "could not be loaded, but vLLM did" not in summary
         assert "errored 6 benchmarks" in summary
 
     def test_includes_errored_summary_line(self) -> None:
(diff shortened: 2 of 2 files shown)


saattrupdan:
- The test fixture uses Python adjacent string literals which concatenate at compile time into a single runtime line, so `"did not mention exactly what"` does appear on one line and the suppression branch IS exercised. No change needed for that comment.
- Added a stronger assertion: `assert "could not be loaded, but vLLM did" not in summary` to guard against the boilerplate prefix surviving even if the exact substring is split differently.
@copilot

PR #1964, opened 2026-07-02
fix: drop verbose=False and narrow mistral tokeniser fallback

### LINKED ISSUE #1775: [MODEL EVALUATION REQUEST] gordicaleksa/SlovenianGPT
### Model ID

gordicaleksa/SlovenianGPT

### Evaluation languages

- [ ] Baltic languages (Latvian, Lithuanian)
- [ ] Finnic languages (Estonian, Finnish)
- [ ] Romance languages (Catalan, French, Italian, Portuguese, Romanian, Spanish)
- [ ] Scandinavian languages (Danish, Faroese, Icelandic, Norwegian, Swedish)
- [X] Slavic languages (Belarusian, Bulgarian, Bosnian, Croatian, Czech, Polish, Serbian, Slovak, Slovenian, Ukrainian)
- [ ] West Germanic languages (Dutch, English, German)
- [ ] Albanian
- [ ] Greek
- [ ] Hungarian

saattrupdan:
Removes the `verbose=False` kwarg from both tokeniser loaders and narrows the
Mistral tokeniser fallback to only trigger for actual Mistral models (not any
error mentioning "mistral"). Fixes #1775.

## Key changes

- `src/euroeval/benchmark_modules/hf.py` — removed `verbose=False` from the
  `loading_kwargs` dict in `load_tokeniser`.
- `src/euroeval/benchmark_modules/vllm.py` — removed `verbose=False` from the
  `AutoTokenizer.from_pretrained` call. Replaced the
  `"mistral" in str(e).lower()` error-string match with a new pure helper
  `_is_mistral_tokeniser_model` that checks model identity only.
- `CHANGELOG.md` — added a Fixed entry under `[Unreleased]`.
- `tests/test_benchmark_modules/test_vllm.py` — added
  `TestIsMistralTokeniserModel` with 5 test cases.

## Why

`MistralCommonBackend.from_pretrained` rejects the `verbose` kwarg that
`AutoTokenizer` passes to all backends, causing a `ValueError` for any model
that transformers routes through the Mistral backend (issue #1775). The
`verbose=False` was unnecessary — euroeval has its own logging.

Worse, the fallback `"mistral" in str(e).lower()` check matched the error
message itself (which contains "MistralCommonBackend") even for non-Mistral
models like `gordicaleksa/SlovenianGPT`, triggering the wrong recovery path.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,13 +29,11 @@ project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 - Fixed race condition in cache cleanup where `rmtree` would fail with
   `FileNotFoundError` when checkpoint directories were removed concurrently during model
   benchmarking. Now uses `ignore_errors=True` to handle concurrent deletions gracefully.
-- The tokeniser loader no longer passes `verbose=False` to
+- The HF, vLLM and fresh model tokeniser loaders no longer pass `verbose=False` to
   `AutoTokenizer.from_pretrained`, since the `MistralCommonBackend` rejects this
-  argument, which raised an error for non-Mistral models such as
-  `gordicaleksa/SlovenianGPT`. The vLLM mistral tokeniser fallback is now gated on
-  model identity (model ID and/or model config) rather than the error string, which
-  previously matched the `MistralCommonBackend` message for unrelated models. Fixes
-  #1775.
+  argument. The vLLM mistral tokeniser fallback is now gated on model identity (model
+  ID and/or model config) rather than the error string, which previously matched the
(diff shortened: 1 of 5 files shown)


saattrupdan:
- Base model detection now also handles versioned base IDs like `mistralai/Mistral-7B-v0.1` (no "base" in ID), while exempting instruct models.
- Removed `verbose=False` from `fresh.py` tokeniser loader as well (was missed in the first pass).
- Added test for versioned base model (`mistralai/Mistral-7B-v0.1`).
- CHANGELOG entry updated to mention all three tokeniser loaders (HF, vLLM, fresh).
@copilot

Issue #1868, opened 2026-06-03
[MODEL EVALUATION REQUEST] google/gemma-4-12B-it-assistant


saattrupdan:
### Model ID

google/gemma-4-12B-it-assistant

### Evaluation languages

- [X] Baltic languages (Latvian, Lithuanian)
- [X] Finnic languages (Estonian, Finnish)
- [X] Romance languages (Catalan, French, Italian, Portuguese, Romanian, Spanish)
- [X] Scandinavian languages (Danish, Faroese, Icelandic, Norwegian, Swedish)
- [X] Slavic languages (Belarusian, Bulgarian, Bosnian, Croatian, Czech, Polish, Serbian, Slovak, Slovenian, Ukrainian)
- [X] West Germanic languages (Dutch, English, German)
- [X] Albanian
- [X] Greek
- [X] Hungarian

saattrupdan:
**Evaluation progress for `google/gemma-4-12B-it-assistant`**

- Completed: be
- In progress: (none)
- Remaining: (none)
- Failed: bg, bs, ca, cs, da, de, el, en, es, et, fi, fo, fr, hr, hu, is, it, lt, lv, nl, no, pl, pt, ro, sk, sl, sq, sr, sv, uk

saattrupdan:
Error encountered during evaluation (euroeval exited with code 1 for language 'bg'):

```bash
ache/model_cache/google--gemma-4-12B-it-assistant', 'seed': 4242, 'max_model_len': 8192, 'distributed_executor_backend': 'mp', 'enable_prefix_caching': False, 'gpu_memory_utilization': 0.8, 'max_num_batched_tokens': 8192, 'disable_log_stats': True, 'revision': 'main', 'enforce_eager': True, 'disable_custom_all_reduce': True, 'max_lora_rank': 256, 'model': 'google/gemma-4-12B-it-assistant'}[0m
[93mResolved architecture: TransformersMultiModalForCausalLM[0m
[93mUsing max model len 8192[0m
[93mChunked prefill is enabled with max_num_batched_tokens=8192.[0m
[93mAsynchronous scheduling is enabled.[0m
[93mEnforce eager set, disabling torch.compile and CUDAGraphs. This is equivalent to setting -cc.mode=none -cc.cudagraph_mode=none[0m
[93mInductor compilation was disabled by user settings, optimizations settings that are only active during inductor compilation will be ignored.[0m
[...]

PR #1965, opened 2026-07-02
fix: surface worker-prefixed vLLM errors in failure summaries

### LINKED ISSUE #1943: [MODEL EVALUATION REQUEST] swiss-ai/Apertus-v1.1-1.5B
### Model ID

swiss-ai/Apertus-v1.1-1.5B

### Evaluation languages

- [X] Baltic languages (Latvian, Lithuanian)
- [X] Finnic languages (Estonian, Finnish)
- [X] Romance languages (Catalan, French, Italian, Portuguese, Romanian, Spanish)
- [X] Scandinavian languages (Danish, Faroese, Icelandic, Norwegian, Swedish)
- [X] Slavic languages (Belarusian, Bulgarian, Bosnian, Croatian, Czech, Polish, Serbian, Slovak, Slovenian, Ukrainian)
- [X] West Germanic languages (Dutch, English, German)
- [X] Albanian
- [X] Greek
- [X] Hungarian

saattrupdan:
Improves `summarise_evaluation_error` to extract the real underlying error
from vLLM worker-process log lines, instead of showing only the generic
"WorkerProc initialization failed ... See stack trace for root cause."
message. Fixes #1943, #1749, #1868.

## Key changes

- `src/leaderboards/queue_parsing.py` — added three module-level regexes for
  worker-prefixed log lines, a `_worker_errors` helper that extracts the
  worker's traceback, and updated `summarise_evaluation_error` to prefer the
  worker block when the main-process traceback is the generic WorkerProc
  exception.
- `tests/leaderboards/test_queue_parsing.py` — three new tests covering:
  worker-prefixed error extraction, worker errors preferred over the generic
  WorkerProc traceback, and genuine tracebacks still taking priority.

## Why

When vLLM's `WorkerProc` fails to initialize, the actual error is logged by
the worker subprocess to stderr (prefixed with `(Worker_TP* pid=*) ERROR`),
but the main process wraps it in `Exception: WorkerProc initialization
failed ... See stack trace for root cause.` (with `raise e from None`, which
strips the cause). The real error is in the captured `FULL_LOG` output but
was not extracted by `summarise_evaluation_error`, so the GitHub failure
comment showed only the uninformative generic message.

### DIFF
--- a/src/leaderboards/queue_parsing.py
+++ b/src/leaderboards/queue_parsing.py
@@ -50,6 +50,14 @@ _TRACEBACK_START = "Traceback (most recent call last):"
 _EXCEPTION_LINE_RE = re.compile(r"^[\w.]*(Error|Exception|Timeout|Interrupt)\b.*")
 _SUMMARY_LINE_RE = re.compile(r"errored\s+\d+\s+benchmarks?", re.IGNORECASE)
 
+# vLLM's ``WorkerProc`` runs in a subprocess that logs to stderr with a
+# ``(Worker_TP* pid=*)`` prefix. When initialisation fails the main process only
+# re-raises a generic ``WorkerProc initialization failed`` exception, while the
+# real cause is in the worker-prefixed ``ERROR`` lines that precede it.
+_WORKER_PREFIX_RE = re.compile(r"^\(Worker_TP\d+\s+pid=\d+\)")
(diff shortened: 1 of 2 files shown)


saattrupdan:
Addressed Copilot comment:
- Checked the regex formatting — ruff's formatter is the authority here and prefers single-line for these short patterns. No change needed; the existing `_LOAD_ERROR_RE` is multi-line because it's longer and wrapped manually, not by the formatter.
@copilot

PR #1971, opened 2026-07-03
fix: collapse duplicate leaderboard scores per model+benchmark (#1970)

### LINKED ISSUE #1970: [BUG] multiple scores per benchmark on the leaderboard
### 🐛 Describe the bug

I think something is wrong with multiple scores appearing on the leaderboard for single model+benchmark combinations.

I first noticed it on the new Apertus 1.1 results, which shows many scores for single benchmarks where we expect only one or two:

Image

### Operating System

MacOS

### Device

Mac GPU

### Python version

3.13.x

### EuroEval version

17.6.0

### Transformers version

n/a

### vLLM version

n/a

saattrupdan:
## Problem

Issue #1970: single model+benchmark combinations show many scores on the leaderboard. Concretely, the Apertus v1.1 rows rendered a metric **repeated**, e.g. in `finnish_all_models.csv`:

```
83.79 ± 0.01 / 91.17 ± 0.01 / 83.79 ± 0.01 / 91.17 ± 0.01
```

That's the primary/secondary pair (`83.79`/`91.17`) duplicated — two records contributing the same two metrics. One or two values per cell is expected (primary, optional secondary); more means duplicate records.

## Root cause

`get_record_hash` (used to deduplicate records) hashed more than what identifies a leaderboard *row*:

- It hashed the **raw model name**, which is stored anchored (`<a ...>org/repo</a>`) for some records and plain for others.
- It hashed the **`generative` flag**, which the row label (`extract_model_ids_from_record`) never uses.

So two records that render on the same row — e.g. one tagged `generative: true` and one missing the flag — got **different hashes**, survived deduplication, and both landed on the same row.

There was also a **caching gotcha**: `load_raw_results` is `@cache`d and is populated by `process_results` *before* it rewrites the per-model files, so `generate_leaderboard` re-used the pre-dedup records within the same run. The dedup done by `process_results` only helped the *next* run.

## Fix

- **Align the hash with the row identity** — `get_record_hash` now strips the anchor, always includes `few_shot`, and drops `generative`. Records that render on the same row hash identically.
- **Deduplicate at the point of consumption** — `group_results_by_model` deduplicates by hash too, so the fix takes effect immediately regardless of the `load_raw_results` cache timing.
- **Guard against regressions** — `generate_dataframe` raises a `ValueError` if any dataset cell would contain more than the primary/secondary pair (more than one slash).
- **Shared helper** — extracted an O(n) `deduplicate_records` helper reused by both `process_results` and `score_extraction`, replacing the previous O(n²) implementation.

## Tests

`tests/leaderboards/test_records.py` and `tests/leaderboards/test_record_fields.py` cover: anchored/plain collapse, generative-flag collapse, newest-version retention, and that genuinely distinct rows (zero-shot / val / different dataset) stay separate.

### DIFF
--- a/src/leaderboards/leaderboard_generation.py
+++ b/src/leaderboards/leaderboard_generation.py
@@ -333,6 +333,11 @@ def generate_dataframe(
     Returns:
         A list of pairs (df, df_simplified), where df is the full leaderboard DataFrame
         and df_simplified is the simplified version.
+
+    Raises:
+        ValueError:
+            If a model has more than two scores for a single dataset, which
+            indicates that duplicate records survived deduplication.
     """
     if model_results == {}:
         logger.error("No model results found, skipping leaderboard generation.")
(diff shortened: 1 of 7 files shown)


saattrupdan:
Addressed all review comments in e21730db:
- Parameterized the `_record` helper return type as `dict[str, object]`.
- Switched the helper calls to keyword arguments to match the `tests/leaderboards/*` convention.
@copilot

PR #1974, opened 2026-07-04
feat: leaderboard-dataset-swap script (eval + config/doc swap + PR)

saattrupdan:
## What

Adds `src/scripts/swap_leaderboard_dataset.py`, which replaces the official dataset for a task on a language's leaderboard with a new one, end to end — and restores `run_core_model_evaluations.py` to a single, clear purpose.

Swapping a dataset means every model with a rank score on the affected leaderboard needs a score on the new dataset before it can go official. This script does the evaluation, the config/doc swap, and (optionally) the PR.

## The swap script

Run on a dedicated branch (`--branch` is required and may not be the default branch):

```bash
uv run src/scripts/swap_leaderboard_dataset.py \
    --old-dataset scala-da --new-dataset dala --branch swap-scala-dala --pr
```

1. **Evaluate.** Finds every model ranked on the affected leaderboard(s) and evaluates it on the new (still unofficial) dataset — *before* flipping any flags, so the live leaderboard stays intact. Selection is category-aware (`generative` + `all_models`), so **both generative and encoder models** are covered when the task supports them. Each eval **mirrors how the model appears on the leaderboard** for the outgoing dataset: `(val)` → validation split (else test), `(zero-shot)` → zero-shot (generative only), using the same field conventions as `leaderboards.records`.
2. **Swap officiality.** Demotes the old dataset and promotes the new one in both the `euroeval` dataset configs (`unofficial=True` + block repositioned) and the frontend docs (`Unofficial:` heading prefix + section reordered), keeping each file's official-first grouping.
3. **Open a PR** (`--pr`): commits on the branch, pushes, and opens a PR assigned to the logged-in user with `saattrupdan` and (best-effort) Copilot as reviewers.

Key flags: `--language` (restrict languages), `--include-api` (off by default → no spend; checks provider env vars), `--gpu-memory-utilization`, `--skip-eval` (config/doc swap only, e.g. when evals already ran), `--dry-run`.

The per-language leaderboard config needs no edit — the dataset↔leaderboard mapping is derived from the dataset configs (`task_metadata`), not stored separately.

## run_core_model_evaluations.py

Reverted to **backfill-only**: both the earlier re-eval additions and the pre-existing `--dataset` dataset-replacement mode are removed, leaving it to evaluate core models on the full test split.

## Verification

Dry-run of `scala-da → dala`: 311 ranked (model, language) pairs (encoders included), 76 API models excluded by default, 230 evaluations planned, 171 fit the local GPU budget. The config/doc surgery was applied to the real Danish files and verified: `scala-da` → unofficial, `dala` → official, module imports and lints cleanly, doc sections reordered correctly. Branch/direction guards and `--skip-eval` all behave. `make check` passes.

### DIFF
--- a/src/scripts/run_core_model_evaluations.py
+++ b/src/scripts/run_core_model_evaluations.py
@@ -56,7 +56,11 @@ from euroeval.constants import ORTHOGONAL_TASKS
 from euroeval.data_models import DatasetConfig
 from euroeval.dataset_configs import get_all_dataset_configs
 from euroeval.languages import get_all_languages
-from leaderboards.constants import NEW_RESULTS_PATH, RESULTS_DIR
+from leaderboards.constants import (
+    DEFAULT_GPU_MEMORY_UTILIZATION,
+    NEW_RESULTS_PATH,
+    RESULTS_DIR,
+)
 from leaderboards.core_models import CoreModel
 from leaderboards.evaluation_common import (
     gpu_total_memory_bytes,
@@ -117,6 +121,14 @@ logger = logging.getLogger("run_core_model_evaluations")
     "Skips the interactive prompt. Requires --include-api.",
 )
 @click.option(
+    "--gpu-memory-utilization",
+    "gpu_memory_utilization",
+    type=float,
+    default=None,
+    help="vLLM GPU memory utilization fraction (0.0-1.0) the fit pre-check should "
+    f"budget against. When omitted, defaults to {DEFAULT_GPU_MEMORY_UTILIZATION}.",
+)
+@click.option(
     "--dry-run",
     is_flag=True,
     default=False,
@@ -136,6 +148,7 @@ def main(
     languages: tuple[str, ...],
     include_api: bool,
     api_providers: str | None,
+    gpu_memory_utilization: float | None,
     dry_run: bool,
     force: bool,
 ) -> None:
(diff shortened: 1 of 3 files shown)


saattrupdan:
Addressed all Copilot comments (commit 4169d722):

- **validation_split=None** (lines 500 & 810): now treated as the validation split, matching the repo-wide backwards-compat convention in `missing_official_dataset_language_pairs`. Added a shared `_is_validation_split` helper used by both the corpus dedup preference and `_mirror_eval_config`, so old records are no longer mirrored onto the test split.
- **--gpu-memory-utilization range** (line 132): constrained to (0, 1] via `click.FloatRange(min=0.0, max=1.0, min_open=True)`; out-of-range values are now rejected up front.
- **O(pairs × languages × categories) scan** (line 910): observations are now indexed by language then model, so each leaderboard language only scans the models actually evaluated in it.

@copilot

saattrupdan:
Addressed all comments on `swap_leaderboard_dataset.py`:

- Added logging for malformed JSON records in `load_corpus()`
- Added staged-diff check in `open_pull_request()` to handle no-op swaps gracefully
- Fixed type annotation in `_record_languages()` to use `dict[str, object]`

The comments on `run_core_model_evaluations.py` reference the removed dataset-replacement mode (--dataset flag was removed). That functionality has been moved to the new `swap_leaderboard_dataset.py` script, which implements a different approach.

@copilot

Issue #1106, opened 2025-07-22
[FEATURE REQUEST] Use `AutoModelForMultipleChoice` when evaluating encoder models on multiple-choice tasks


saattrupdan:
### 🚀 The feature, motivation and pitch

We currently have a "manual" evaluation method when evaluating encoder models on multiple choice tasks. We should change this to the built-in `AutoModelForMultipleChoice` class in `transformers`, and also see if this changes the results.

See guide here: https://huggingface.co/docs/transformers/tasks/multiple_choice

PR #1976, opened 2026-07-09
feat: use native AutoModelForMultipleChoice head for encoder multiple-choice

pariidanDKE:
This PR addresses issue #1106. Specifically, it implements the transformers-native way of fine-tuning and evaluating encoders on the multiple-choice task, as shown in [this example](https://huggingface.co/docs/transformers/tasks/multiple_choice).

This change simplifies the code, substituting the custom collator and trainer with the ones from transformers core.

The approach is also functionally quite different from the manual approach implemented before. The manual approach classifies each answer as Yes/No, so it only ever deals with a single (question, answer) pair. As a result, it flattens each example from the MC dataset: if a question has `num_choices` options, then after the dataset is prepared we have `num_choices` independent rows per question. The built-in transformers class does this differently, it handles all `num_choices` options within a single `forward` call. Internally, `AutoModelForMultipleChoice` runs the backbone independently on each option and extracts a score for each via a `Linear(hidden, 1)` head. After `num_choices` such runs we have `num_choices` logits, which are then passed through a single softmax, coupling them together. This means the final answer is aware of the other options, and so is the gradient that gets backpropagated from the loss.

The built-in approach has another advantage: it means we no longer need any custom code for evaluation/inference. The forward now takes in `(batch, num_choices, seq_len)` and returns `(batch, num_choices)` logits, which we simply `argmax` over, removing all of the previous complexity. The added logic now mostly resides in a new version of *prepare_examples* in *multiple_choice_classification*, which consists of formatting data to align with what `DataCollatorForMultipleChoice` expects.

## Findings from using the new method

- One thing to keep in mind is that the `AutoModelForMultipleChoice` head is less common, especially for encoders, than the previously used `AutoModelForSequenceClassification`, which means some models may not implement it. This mainly affects custom architectures that are not part of transformers core. Empirically, while testing I found that **EuroBERT**'s `auto_map` has no entry for `AutoModelForMultipleChoice` (its remote code never implements that head), so it fails to load and is reported as an invalid benchmark rather than being evaluated.
- Separately, `XLMRobertaForMultipleChoice` currently raises an `IndexError` on transformers versions starting from September 2025, due to a one-line upstream bug, unrelated to this PR. Fix submitted upstream: huggingface/transformers#47147.
- I also ran some A/B tests. I picked 7 strong encoder models (the top 7 by rank score on the leaderboard) and ran them across 4 multiple-choice datasets (English only):

| Model | belebele (bin→MC) | hellaswag (bin→MC) | winogrande (bin→MC) | arc (bin→MC) |
| --- | --- | --- | --- | --- |
| mmBERT-base | 36.1→33.4 | 35.4→41.2 | 50.2→47.6 | 28.6→29.8 |
| mmBERT-small | 26.9→30.6 | 27.9→29.6 | 53.2→54.1 | 27.8→31.6 |
| ModernBERT-base | 26.9→30.0 | 28.1→32.7 | 51.6→49.8 | 27.7→27.2 |
| ModernBERT-large (answerdotai) | 29.1→43.6 | 74.4→70.5 | 52.7→48.0 | 26.0→31.1 |
| ModernBERT-large (AI-Sweden) | 29.5→31.7 | 32.3→60.7 | 52.4→52.3 | 25.7→27.5 |
| mdeberta-v3-base | 26.6→23.0 | 35.9→48.7 | 53.3→51.3 | 25.1→24.2 |
| EUBERT | 28.0→22.5 | 28.7→28.6 | 53.4→48.5 | 27.5→32.1 |

> **Note**: each cell is `binary (manual) → MC-head (AutoModel)` mean test accuracy over 10 iterations.

The findings show an overall improvement on the easier tasks (belebele and hellaswag), with a few models seeing substantial gains (up to +28 points) and no model/dataset combination regressing by a comparable margin. Most individual differences are within noise and go either way depending on the model/dataset, which is especially true for the harder tasks (arc and winogrande), where neither approach clears chance. **I can also run more tests, if there is interest in a specific model or task.**

> **Note (enabling encoders for MC):** these tasks were previously restricted to generative models, as encoders were near-random on them. Given the slightly better results in some cases with the new approach, it may now be worth enabling encoders here, but that is left as a decision for the maintainers.


pariidanDKE:
The CI failures aren't from this PR. They're all the same error (Illegal header value b'Bearer '), caused by fork PRs not getting repo secrets under the new pull_request trigger. So HF_TOKEN etc. come through empty.

saattrupdan (dismissed):
Thanks for your PR!

I made a few changes:

- Fixed the CI issues (nothing to do with your implementation)
- Bumped `transformers` to >=5.14.0, to accomodate your fix (and thanks for that!)
- Changed the multiple-choice dataset configs to allow encoder evals
- Added a check when an encoder model is evaluated on a multiple-choice dataset that has
  a variable number of choices, and raise an informative `InvalidBenchmark` error in
  that case
- Now raises an `InvalidBenchmark` error if the encoder model doesn't have a
  `ForMultipleChoice` implementation. Previously it raised `InvalidModel`, which would
  skip all the other datasets for the model, but now we only skip the one.

saattrupdan (approved):

PR #1977, opened 2026-07-10
fix(changelog): remove stray leftover merge conflict marker

avalyset:
`CHANGELOG.md` has a leftover `>>>>>>> main` merge conflict marker between the `### Fixed` section and the `## [v16.11.0]` header. There's no matching `<<<<<<<`/`=======`, so it's a stray tail from a previously resolved conflict, not an unresolved block. This removes the marker line and keeps a single blank line at the seam so `markdownlint` stays green. Docs-only, no content changes.


saattrupdan (approved):
Well spotted, thanks!

PR #2011, opened 2026-07-19
Store evaluation results as per-record JSON files

saattrupdan:
## What

Change EuroEval result storage from flat per-model JSONL files to a per-record
JSON tree keyed by the logical result identity. Each result now lives at
`results/<model>/<dataset>__<split>__<shot>.json`, matching the same layout in
the `EuroEval/results` Hugging Face bucket.

## Key features

- Stores each `(model_id, dataset, validation_split, few_shot)` result as its own
  JSON file.
- Uses a canonical storage-identity helper for filenames, deduplication, loading,
  uploading, and sync reconciliation.
- Makes bucket sync loss-free for distinct results by avoiding shared per-model
  files.
- Removes the historical `unknown.jsonl` path entirely; invalid records are
  dropped instead of stored.
- Preserves existing JSONL staging/output files (`new_results.jsonl` and
  `euroeval_benchmark_results.jsonl`).
- Updates backups, result loading, queue processing, collection scripts,
  fix/sync scripts, spider plots, tests, and AGENTS.md documentation for the new
  layout.

## Examples

A result for `Qwen/Qwen3-30B-A3B#no-thinking` on zero-shot test evaluation of
`danish_ner` is stored as:

```text
results/Qwen_Qwen3-30B-A3B#no-thinking/danish_ner__test__zeroshot.json
```

A few-shot validation result for the same model and dataset is stored separately
as:

```text
results/Qwen_Qwen3-30B-A3B#no-thinking/danish_ner__val__fewshot.json
```

## Why it helps

HF bucket sync resolves conflicts per file. With one append-only JSONL file per
model, parallel evaluators could overwrite each other's distinct results. One
file per logical result turns sync into a safe file-set union; only genuine
re-evaluations of the same logical result share a path.

## Validation

- `make check`
- `uv run pytest tests/leaderboards/ -q` -> 245 passed
- `uv run pytest tests/test_scripts/ -q` -> 126 passed

The migration and bucket cutover are intentionally not included in this PR; they
remain a separate manual operational step after review.


saattrupdan:
- Added explicit path-collision detection before result-processing dedup.
- Corrected swap-result consolidation logging to distinguish total vs new records.
- Added model_info.id fallbacks where name can be missing.
- Added ISO-8601 timestamp parsing for deterministic dedup tie-breaks.
- Re-ran make check successfully.

@copilot

PR #2025, opened 2026-07-21
feat: swap official dataset global-mmlu-lite-sq -> include-sq

saattrupdan:
Swaps the official dataset `global-mmlu-lite-sq` for `include-sq`.

## What

- Every model with a rank score on the affected leaderboard(s) was evaluated on `include-sq`, mirroring how each appears on the leaderboard (validation/test split and zero-/few-shot).
- `global-mmlu-lite-sq` is demoted to unofficial and `include-sq` promoted to official in the dataset configs and the dataset documentation, keeping each file's official-first grouping.

The leaderboards will pick up the change on the next regeneration.

### DIFF
--- a/src/euroeval/dataset_configs/albanian.py
+++ b/src/euroeval/dataset_configs/albanian.py
@@ -56,14 +56,6 @@ LR_SUM_SQ_CONFIG = DatasetConfig(
     languages=[ALBANIAN],
 )
 
-GLOBAL_MMLU_LITE_SQ_CONFIG = DatasetConfig(
-    name="global-mmlu-lite-sq",
-    pretty_name="GlobalMMLULite-sq",
-    source="EuroEval/global-mmlu-lite-sq",
-    task=KNOW,
-    languages=[ALBANIAN],
-)
-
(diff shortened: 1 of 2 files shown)


EwoutH:
While working on Global-MMLU, can more languages be added? (including Dutch)

saattrupdan:
> While working on Global-MMLU, can more languages be added? (including Dutch)

We could, but most languages in Global-MMLU are just machine translations - see Table 5 in [the paper](https://arxiv.org/pdf/2412.03304). Dutch is part of this MT group.

But sure, we could add the others as well, as long as we're documenting that it's just a machine translated version of MMLU - feel free to open an issue on this or create a PR. There's already a [dataset creation script](https://github.com/EuroEval/EuroEval/blob/main/src/scripts/dataset_creation/create_global_mmlu.py), so it wouldn't be that difficult to add more (documentation needs to be added as well though, see the [dataset creation guide](https://github.com/EuroEval/EuroEval/blob/main/NEW_DATASET_GUIDE.md))

PR #2045, opened 2026-07-24
Feat/add faroese metaphorical explanations

Biorrith:
## What

Added the faroese-metaphorical-explanations knowledge dataset from
this paper, following #2044.

## Key features

- Multiple choice: pick the correct explanation of a Faroese idiom from four options
- 457 samples, split 140 / 35 / 282 following the standard split ratio
- Includes creation script, documentation section and changelog entry

PR #2047, opened 2026-07-24
Feat/add faroese grammatical correctness

Biorrith:
## What

Added a new grammatical error correction (GEC) task and the
faroese-grammatical-correctness dataset from
this paper, following #2045

## Key features

- New grammatical-error-correction task: the model is given a sentence with grammatical errors and generates the corrected version (text-to-text)
- New generic exact_match metric fitting the text-to-text task group — the existing em_metric is tied to the SQuAD QA format and normalises casing and punctuation, which are part of what this task tests
- Prompt templates for Danish, English and Faroese, plus a task documentation page
- 6,628 minimal pairs from high school essays, split 1,024 / 256 / 2,048, with leftovers kept in a full_train split
- Includes creation script, documentation section and changelog entry

### DIFF
--- a/src/euroeval/prompt_templates/grammatical_error_correction.py
+++ b/src/euroeval/prompt_templates/grammatical_error_correction.py
@@ -1,0 +1,31 @@
+"""Templates for the Grammatical Error Correction task."""
+
+from ..data_models import PromptConfig
+from ..languages import DANISH, ENGLISH, FAROESE
+
+GEC_TEMPLATES = {
+    DANISH: PromptConfig(
+        default_prompt_prefix="Følgende er sætninger med grammatiske fejl og deres "
+        "korrigerede versioner.",
+        default_prompt_template="Sætning: {text}\nKorrigeret sætning: {target_text}",
+        default_instruction_prompt="Sætning: {text}\n\nKorriger de grammatiske fejl "
+        "i sætningen og skriv den korrigerede sætning, og intet andet.",
+        default_prompt_label_mapping=dict(),
+    ),
+    ENGLISH: PromptConfig(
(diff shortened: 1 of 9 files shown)


src/euroeval/prompt_templates/grammatical_error_correction.py
```diff
@@ -0,0 +1,31 @@
...
+    DANISH: PromptConfig(
+        default_prompt_prefix="Følgende er sætninger med grammatiske fejl og deres "
+        "korrigerede versioner.",
+        default_prompt_template="Sætning: {text}\nKorrigeret sætning: {target_text}",
+        default_instruction_prompt="Sætning: {text}\n\nKorriger de grammatiske fejl "
+        "i sætningen og skriv den korrigerede sætning, og intet andet.",
+        default_prompt_label_mapping=dict(),
+    ),
+    ENGLISH: PromptConfig(
+        default_prompt_prefix="The following are sentences with grammatical errors "
```
saattrupdan, line 16: Do we only allow samples _with_ errors, or should we also have samples which are already correct, to make sure that the model also knows when the samples are right?

Biorrith: Hmm, we currently drop samples that are the same in the creation of the dataset. It would come down to if we should keep doing that or not.

saattrupdan: Hm, good point, because then we'd be suddenly measuring two things: detecting whether there *is* an error (LA) _and_ correcting it (GED). So since we already have an LA task, it's probably fine to have this only be about correction. So let's just leave it :)

saattrupdan (approved):
LGTM

PR #2055, opened 2026-07-25
Fix ruff complexity violations in src/scripts/

saattrupdan:
Fixes 28 ruff complexity violations (C901 and PLR0912) across evaluation and dataset creation scripts.

## Key changes

- Refactored collect_evaluation_results.py (4 functions)
- Refactored create_language_spider_plot.py (2 functions)
- Refactored process_evaluation_queue.py (2 functions)
- Refactored create_european_values.py, create_kpwr_ner.py, create_harem.py, fix_results.py (1 function each)

## Python skill compliance

All changes verified against the python skill conventions:
- ✅ British English spelling throughout
- ✅ Python 3.12+ type hint syntax (list[T], X | Y)
- ✅ Google-style docstrings with proper formatting
- ✅ Absolute imports for cross-package modules
- ✅ **Function ordering by call hierarchy**: callers before callees
- ✅ **`if __name__ == "__main__"`** at bottom of files
- ✅ No print statements, no % formatting
- ✅ Line length within 88 characters
- ✅ pathlib.Path objects for file paths
- ✅ Keyword arguments for function calls

make check now passes for src/scripts/ files.

### DIFF
--- a/src/euroeval/benchmark_config_factory.py
+++ b/src/euroeval/benchmark_config_factory.py
@@ -1,7 +1,6 @@
 """Factory class for creating dataset configurations."""
 
 import collections.abc as c
-import importlib.util
 import logging
 import sys
 import typing as t
@@ -15,9 +14,6 @@ from .dataset_configs import get_all_dataset_configs
 from .enums import Device
 from .languages import get_all_languages, get_correct_language_codes
 from .logging_utils import log
-
-if importlib.util.find_spec("vllm") is not None:
-    pass
 
 if t.TYPE_CHECKING:
     from .data_models import Language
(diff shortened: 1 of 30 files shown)


saattrupdan:
Addressed all Copilot review comments:

- Fixed `import collections.abc as c` - removed incorrect import, using `go.Figure` for type annotation
- Refactored `_handle_pending_skips` to return values instead of trying to modify immutable strings; intentional skips now properly update `done` list
- Click decorators are correctly on `main()` (verified the comment was based on earlier version)
- Fixed `--force` option propagation: added `force` parameter to `_deploy_and_close_issues` and pass through to `regenerate_leaderboards`

@copilot

PR #2067, opened 2026-07-26
fix(bucket_sync): preserve existing JSONL records in merge_results

saattrupdan:
## What

Fixes a data-loss bug in the `merge_results()` function where existing records in `euroeval_benchmark_results.jsonl` were overwritten and lost when the function ran with a partially-synced results tree.

## Key Changes

- **`merge_results()`** now reads existing records from the JSONL file before merging tree records
- Preserves results that haven't been synced to the `results/` tree yet  
- Tree records still override JSONL records with matching identity (existing dedup logic)
- Added 2 new tests covering preservation and override behaviour

## Why This Matters

When the swap script was interrupted mid-evaluation and re-run:
1. Script loaded JSONL, saw 111 models "already had results", skipped them
2. Only ~60 new models were evaluated  
3. `merge_results()` overwrote JSONL with only tree records (~67 models)
4. **Lost**: 111-67 = ~44 models worth of results that were "skipped" but not in tree

## Testing

All 6 `TestMergeResults` tests pass, including 2 new ones:
- `test_merge_results_preserves_existing_jsonl` - JSONL records are preserved
- `test_merge_results_tree_overrides_jsonl` - tree records still win with same identity

### DIFF
--- a/src/leaderboards/bucket_sync.py
+++ b/src/leaderboards/bucket_sync.py
@@ -300,29 +300,24 @@ def merge_results(results_file: Path) -> int:
     # PHASE 1: Read existing records from the JSONL file (if it exists)
     # This preserves results that haven't been synced to the tree yet
     if results_file.exists():
-        logger.info("Reading existing records from %s...", results_file)
-        try:
-            with results_file.open("r", encoding="utf-8") as f:
-                for line in f:
-                    if not line.strip():
-                        continue
-                    try:
-                        record = json.loads(line)
-                        identity = identity_from_eee_record(record)
-                        if identity not in existing:
-                            existing[identity] = record
-                        else:
-                            # Deduplicate: keep newer record (tree records in Phase 2
-                            # will override via dedup_newer_record)
-                            existing[identity] = dedup_newer_record(
-                                existing[identity], record
-                            )
-                    except (json.JSONDecodeError, ValueError, KeyError, TypeError):
-                        pass  # Skip invalid lines (TypeError from normalise_bool_value)
-        except Exception as e:
-            # Fail fast: don't overwrite JSONL if we can't read existing records
-            logger.error("Could not read existing JSONL file: %s", e)
-            return 0
+        logger.info(f"Reading existing records from {results_file}...")
+        with results_file.open("r", encoding="utf-8") as f:
+            for line in f:
+                if not line.strip():
+                    continue
+                try:
+                    record = json.loads(line)
+                    identity = identity_from_eee_record(record)
+                    if identity not in existing:
+                        existing[identity] = record
+                    else:
+                        # Deduplicate: keep newer record (tree records in Phase 2
+                        # will override via dedup_newer_record)
+                        existing[identity] = dedup_newer_record(
+                            existing[identity], record
+                        )
+                except (json.JSONDecodeError, ValueError, KeyError, TypeError):
+                    pass  # Skip invalid lines (TypeError from normalise_bool_value)
 
     # Remove empty JSON files before processing
     remove_empty_json_files(RESULTS_DIR)
--- a/src/leaderboards/bucket_sync.py
+++ b/src/leaderboards/bucket_sync.py
@@ -311,12 +311,18 @@ def merge_results(results_file: Path) -> int:
                         identity = identity_from_eee_record(record)
                         if identity not in existing:
                             existing[identity] = record
-                        # Note: we keep the existing record, tree records will
-                        # override if newer (Phase 2)
-                    except (json.JSONDecodeError, ValueError, KeyError):
-                        pass  # Skip invalid lines
+                        else:
+                            # Deduplicate: keep newer record (tree records in Phase 2
+                            # will override via dedup_newer_record)
+                            existing[identity] = dedup_newer_record(
+                                existing[identity], record
+                            )
+                    except (json.JSONDecodeError, ValueError, KeyError, TypeError):
+                        pass  # Skip invalid lines (TypeError from normalise_bool_value)
         except Exception as e:
-            logger.warning("Could not read existing JSONL file: %s", e)
+            # Fail fast: don't overwrite JSONL if we can't read existing records
+            logger.error("Could not read existing JSONL file: %s", e)
+            return 0
 
     # Remove empty JSON files before processing
     remove_empty_json_files(RESULTS_DIR)
--- a/src/leaderboards/bucket_sync.py
+++ b/src/leaderboards/bucket_sync.py
@@ -282,10 +282,11 @@ def _sort_key(record: dict) -> tuple[str, str]:
 def merge_results(results_file: Path) -> int:
     """Merge per-record JSON tree into a single JSONL file.
 
-    Reads all ``results/*/*.json`` files and writes a deduplicated JSONL file.
-    Deduplication uses canonical result identity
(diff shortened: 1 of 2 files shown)


src/leaderboards/bucket_sync.py
```diff
@@ -296,10 +297,31 @@ def merge_results(results_file: Path) -> int:
     """
     existing: dict[ResultIdentity, dict] = {}
 
+    # PHASE 1: Read existing records from the JSONL file (if it exists)
+    # This preserves results that haven't been synced to the tree yet
+    if results_file.exists():
+        logger.info("Reading existing records from %s...", results_file)
```
saattrupdan: Use f-strings over %s syntax

src/leaderboards/bucket_sync.py
```diff
@@ -296,10 +297,31 @@ def merge_results(results_file: Path) -> int:
...
+        try:
```
saattrupdan: Remove this try-except block. If this fails for whatever reason then we *want* it to raise an error and block the script.

src/leaderboards/bucket_sync.py
```diff
@@ -296,10 +297,31 @@ def merge_results(results_file: Path) -> int:
...
+            with results_file.open("r", encoding="utf-8") as f:
+                for line in f:
+                    if not line.strip():
+                        continue
+                    try:
+                        record = json.loads(line)
+                        identity = identity_from_eee_record(record)
+                        if identity not in existing:
```
saattrupdan: This should also check if the `record` is newer than the existing one, and overwrite it if so. Being newer means newer in the sense of eval_library.version and retrieved_timestamp, in that order.

tests/leaderboards/test_bucket_sync.py
```diff
@@ -262,6 +262,107 @@ def test_merge_results_skips_invalid_json(
...
+        results_file = tmp_path / "merged.jsonl"
+
+        # Write existing records to the JSONL file
+        existing_record = {
+            "model_info": {"id": "existing/model"},
+            "eval_library": {
+                "additional_details": {"dataset": "existing_ds"},
+                "version": "1.0.0",
+            },
+            "retrieved_timestamp": "2024-01-01T00:00:00Z",
```
saattrupdan: The `retrieved_timestamp` is just a number (in a string), like "1784907054" or "1784907152".

saattrupdan:

1. **Deduplication for JSONL duplicates**: Now uses `dedup_newer_record()` instead of first-wins, consistent with the docstring claim that "newer records win"
2. **TypeError handling**: Added `TypeError` to the except clause (can be raised by `normalise_bool_value()` via `identity_from_eee_record()`)
3. **Fail-fast on read errors**: Changed from logging warning and continuing to logging error and returning 0, preventing data-loss scenario where JSONL gets overwritten with only tree records

@copilot

src/leaderboards/bucket_sync.py
```diff
@@ -296,10 +297,32 @@ def merge_results(results_file: Path) -> int:
...
+                    if identity not in existing:
+                        existing[identity] = record
+                    else:
+                        # Deduplicate: keep newer record (tree records in Phase 2
+                        # will override via dedup_newer_record)
+                        existing[identity] = dedup_newer_record(
+                            existing[identity], record
+                        )
+                except (json.JSONDecodeError, ValueError, KeyError, TypeError):
+                    pass  # Skip invalid lines (TypeError from normalise_bool_value)
```
saattrupdan: I think some logging is required here.

src/leaderboards/bucket_sync.py
```diff
@@ -296,10 +297,32 @@ def merge_results(results_file: Path) -> int:
...
+                try:
+                    record = json.loads(line)
+                    identity = identity_from_eee_record(record)
+                    if identity not in existing:
+                        existing[identity] = record
+                    else:
+                        # Deduplicate: keep newer record (tree records in Phase 2
+                        # will override via dedup_newer_record)
+                        existing[identity] = dedup_newer_record(
+                            existing[identity], record
```
saattrupdan: Use kwargs

tests/leaderboards/test_bucket_sync.py
```diff
@@ -262,6 +262,107 @@ def test_merge_results_skips_invalid_json(
...
+        # Create a different record in the tree
+        model_dir = results_dir / "tree_model"
+        model_dir.mkdir()
+        tree_record = {
+            "model_info": {"id": "tree/model"},
+            "eval_library": {
+                "additional_details": {"dataset": "tree_ds"},
+                "version": "1.0.0",
+            },
+            "retrieved_timestamp": "1704153600",  # Unix timestamp as string (newer)
```
saattrupdan: Remove this comment

tests/leaderboards/test_bucket_sync.py
```diff
@@ -262,6 +262,107 @@ def test_merge_results_skips_invalid_json(
...
+        results_file = tmp_path / "merged.jsonl"
+
+        # Write existing records to the JSONL file
+        existing_record = {
+            "model_info": {"id": "existing/model"},
+            "eval_library": {
+                "additional_details": {"dataset": "existing_ds"},
+                "version": "1.0.0",
+            },
+            "retrieved_timestamp": "1704067200",  # Unix timestamp as string
```
saattrupdan: Remove comment

tests/leaderboards/test_bucket_sync.py
```diff
@@ -262,6 +262,107 @@ def test_merge_results_skips_invalid_json(
...
+        results_file = tmp_path / "merged.jsonl"
+
+        # Write older record to JSONL
+        old_record = {
+            "model_info": {"id": "same/model"},
+            "eval_library": {
+                "additional_details": {"dataset": "same_ds"},
+                "version": "1.0.0",
+            },
+            "retrieved_timestamp": "2024-01-01T00:00:00Z",
```
saattrupdan: timestamp is of the wrong type

tests/leaderboards/test_bucket_sync.py
```diff
@@ -262,6 +262,107 @@ def test_merge_results_skips_invalid_json(
...
+        # Write newer record with same identity to tree
+        model_dir = results_dir / "same_model"
+        model_dir.mkdir()
+        new_record = {
+            "model_info": {"id": "same/model"},
+            "eval_library": {
+                "additional_details": {"dataset": "same_ds"},
+                "version": "2.0.0",
+            },
+            "retrieved_timestamp": "2024-01-02T00:00:00Z",
```
saattrupdan: timestamp is of the wrong type

saattrupdan:
Addressed all review feedback:

1. **F-strings** - Changed `%s` to f-string formatting
2. **Removed try-except wrapper** - Let errors propagate and fail fast
3. **Deduplication for JSONL duplicates** - Now uses `dedup_newer_record()` with kwargs for duplicates within JSONL  
4. **Added logging** - Debug log added for skipped invalid lines
5. **Timestamp format** - Fixed tests to use Unix timestamp strings (e.g., `"1704067200"`)
6. **Removed comments** - Cleaned up unnecessary test comments
7. **Kwargs** - All `dedup_newer_record()` calls now use keyword arguments

All 6 `TestMergeResults` tests pass.

@saattrupdan Ready for your review.

PR #2082, opened 2026-07-29
feat: swap official dataset mmlu-nl -> include-nl, multiloko-nl

saattrupdan:
Swaps the official dataset `mmlu-nl` for `include-nl`, `multiloko-nl`.

## What

- Every model with a rank score on the affected leaderboard(s) was evaluated on `include-nl`, `multiloko-nl`, mirroring how each appears on the leaderboard (validation/test split and zero-/few-shot).
- `mmlu-nl` is demoted to unofficial and `include-nl`, `multiloko-nl` promoted to official in the dataset configs and the dataset documentation, keeping each file's official-first grouping.

The leaderboards will pick up the change on the next regeneration.


EwoutH:
Thanks for this!

one more: it seems like `gemini/gemini-3-flash-preview#thinking (zero-shot, val)` is a huge outsider here. 
IMG_1178

PR #2093, opened 2026-08-01
fix: sync multilingual language rank scores

saattrupdan:
## What

Fixes multilingual leaderboard per-language rank scores so each language column is computed with the same monolingual eligible model set as the corresponding monolingual leaderboard. The overall multilingual `Rank score` and ordinal `Rank` remain pan-leaderboard metrics.

## Why

The European leaderboard could show a different Albanian rank score than the Albanian leaderboard for the same model. For example, `Qwen/Qwen3.6-27B-FP8 (val)` showed `1.56 ± 0.08` on the Albanian leaderboard, but `1.40 ± 0.30` in the European Albanian column. The per-language column was being derived from the multilingual eligible set rather than the language-specific eligible set.

## Changes

- Recompute multilingual per-language rank-score columns from language-specific eligible model sets.
- Preserve the existing multilingual eligible set for overall rank score and ordinal rank.
- Add regression tests covering language-specific eligibility, partial multilingual eligibility, and monolingual output stability.
- Update the generated coverage badge after `make check test`.

## Validation

- `make check test` — passed (`1000 passed, 84 skipped`)
- Reviewer agent verdict: Pass

## Notes

`src/frontend/csv/` is git-ignored/generated, so generated leaderboard CSVs are not included in this PR. The next forced leaderboard generation (`--force` / `make force-leaderboards`) will refresh the displayed values.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@
 [![Second paper](https://img.shields.io/badge/arXiv-2406.13469-b31b1b.svg)](https://arxiv.org/abs/2406.13469)
 [![License](https://img.shields.io/github/license/EuroEval/EuroEval)](https://github.com/EuroEval/EuroEval/blob/main/LICENSE)
 [![LastCommit](https://img.shields.io/github/last-commit/EuroEval/EuroEval)](https://github.com/EuroEval/EuroEval/commits/main)
-[![Code Coverage](https://img.shields.io/badge/Coverage-73%25-yellow.svg)](https://github.com/EuroEval/EuroEval/tree/main/tests)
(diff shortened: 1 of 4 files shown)


saattrupdan:
- Reused monolingual per-language rank-score confidence intervals via a shared generation-run cache, so multilingual boards can use already-computed monolingual bootstraps instead of recomputing them.
- Removed the no-op expression statement in the partial-eligibility test.
- Removed the unused `_make_partial_results` test helper from the suppressed comment.

Validation after changes:
- `make check test` passed: 1000 passed, 84 skipped.
- `uv run pytest tests/leaderboards/test_leaderboard_generation.py -q` passed after removing the unused helper.

@copilot

PR #2095, opened 2026-08-02
Fix add-only leaderboard variant mirroring

saattrupdan:
Fixes `swap_leaderboard_dataset.py --add-only` so planned evaluations mirror the actual ranked leaderboard row variant instead of falling back to model-type defaults.

Key changes:
- Adds variant-aware result indexes for displayed rows such as `(val)` and `(zero-shot)`.
- Skips existing results only when the exact split/prompting variant already exists.
- Prevents collapsed unions of incomplete variants from being treated as ranked rows.
- Handles ID-only EEE records and split-agnostic `validation_split=None` datasets consistently with leaderboard generation.

Validation:
- `uv run pytest tests/test_scripts/test_swap_leaderboard_dataset.py -q`
- pre-commit hooks: ruff, ruff format, ty, funcsort, vulture

### DIFF
--- a/src/scripts/swap_leaderboard_dataset.py
+++ b/src/scripts/swap_leaderboard_dataset.py
@@ -45,7 +45,7 @@ import subprocess
 import tempfile
 import typing as t
 from collections import defaultdict
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from datetime import datetime, timezone
 from pathlib import Path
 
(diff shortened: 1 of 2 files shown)


saattrupdan:
Addressed Copilot review:
- Simplified `configs_to_process` to store only configs and removed the unused `is_from_variant` tuple field.
- Re-ran focused ruff, ty, and swap script tests successfully.
@copilot

PR #2096, opened 2026-08-02
fix: respect debug for startup verbosity

saattrupdan:
## What

Fixes the startup message shown when running EuroEval with debug mode enabled. Since `--debug` already implies verbose output, debug runs now show the concise startup message instead of telling users to run with `--verbose`.

## Key changes

- Resolves the effective debug value before choosing the startup message.
- Preserves the normal non-debug startup hint.
- Adds focused regression coverage for initialiser-level and call-level debug settings.

## Validation

- `uv run pytest tests/test_benchmarker.py::TestDebugStartupVerbosity -v`
- `make check test`


saattrupdan:
Addressed Copilot feedback:
- Fixed: pinned `FULL_LOG` unset in the startup verbosity regression tests so the non-debug case is deterministic.
- Re-ran `uv run pytest tests/test_benchmarker.py::TestDebugStartupVerbosity -v` and `make check test`; both pass.
@copilot

PR #2104, opened 2026-08-04
﻿Feat: Add instruction leaderboard category

### LINKED ISSUE #1630: [FEATURE REQUEST] Add new "Chat"/"Instruction" leaderboard

We're starting to get more tasks that are only suited for instruction-tuned models, so we should consider creating a separate leaderboard for these kinds of models.

It could be all zero-shot, since few-shot is mainly required on the generative leaderboard because of the base decoders. They could still appear on the other leaderboards, just with a "(zero-shot)" note added to the model ID, just like we're doing now already.

With such a leaderboard, we'd have three categories:

- **All models**: This covers the NLU tasks only, and includes encoders and decoders.
- **Generative**: This covers the NLU tasks as well as the ones only relevant for generative models, and includes all decoders.
- **Chat**: This covers all the tasks, and only includes instruction-tuned decoders.

So far, we'd have the following extra tasks in the chat leaderboard:

- Instruction following
- Tool use
- Logic reasoning (coming)
- Bias evaluation
- European values

viggo-gascou:
This PR adds the backend + frontend groundwork for a new leaderboard category, scoped to instruction-tuned/reasoning models and covering every task (closes #1630).

It includes every task from the generative leaderboard, in addition to the ones listed in #1630. `european-values` is currently still a bonus column rather than counting toward the rank, matching the behaviour in the other leaderboards

## To do

- [x] **Display name** - Need to decide on a display name e.g., "Chat"/"Instruction"/"Agentic". **Edit: Chosen to be "Chat"**
- [x] **`logical-reasoning`** - should it be instruct-exclusive or open to base decoders? Left out of `LEADERBOARD_TASKS` for now (including it broke Generative's ranking due to near-zero coverage). **Edit: Made instruct-exclusive for now.**
- [x] **`tool-calling`/`multiple-choice-stereotype-bias`** - currently included (unlike `logical-reasoning`). Keep them in or wait until broader coverage? `tool-calling` only has an English dataset, bias only Dutch, so very few models would qualify until coverage grows.
  - [x] **General UX** - Right now, the extra leaderboard + scatter plot is just added as an additional tab to the existing leaderboard UI. This does not look great, and is kinda cluttered. Not sure what the best approach is.
      - **Edit: Settled on this design. The selection of leaderboard/scatterplot is "sticky", such that changing the overall category does not reset the view that was selected.** screen-recording

- [x] **Docs** - `leaderboards-README.md` and `methodology.md` still describe only two leaderboard categories.


viggo-gascou:
@saattrupdan Marked this as a draft mainly because of the to-do list above, the two I'd like some input on most are the display name, any preferences? and whether to keep `tool-calling`/`multiple-choice-stereotype-bias` given their current coverage. Either way on the latter is an easy change, just adding/removing them from the task list in `src/leaderboards/constants.py` (lines 182-195). I'm also working on a better UI for this in the meantime, will post here once it's further along.

viggo-gascou:
﻿This probably shouldn't be merged until there's enough zero-shot coverage for models to actually show up ranked, right now the Chat leaderboard has no ranked models because of lack of coverage across all of the tasks.

src/frontend/components/LeaderboardView.vue
```diff
@@ -15,17 +15,17 @@ const props = defineProps<{
   title: string;
 }>();
 
-type TabId =
-  | "generative"
-  | "all_models"
-  | "generative-scatter"
-  | "all_models-scatter";
+type CategoryId = "generative" | "instruct" | "all_models";
```
saattrupdan: "instruct" should be "chat", and for consistency it'd be great to use the same order: chat, generative, all_models.

src/frontend/components/LeaderboardView.vue
```diff
@@ -34,21 +34,27 @@ const loadFor = async (stem: string) => {
...
       loadLeaderboardMetadata(`${stem}_generative`),
+      loadLeaderboardMetadata(`${stem}_instruct`),
       loadLeaderboardMetadata(`${stem}_all_models`),
     ]);
     generativeTable.value = g ?? null;
+    instructTable.value = c ?? null;
     allModelsTable.value = a ?? null;
     generativeMetadata.value = gMeta ?? null;
+    instructMetadata.value = cMeta ?? null;
     allModelsMetadata.value = aMeta ?? null;
```
saattrupdan: Nit: Ordering and naming again

src/frontend/components/LeaderboardView.vue
```diff
@@ -61,21 +67,23 @@ const loadFor = async (stem: string) => {
 watch(
   () => props.stem,
   (s) => {
-    tab.value = "generative";
+    // Reset the category on language switch, but keep the user's table/scatter
+    // preference - switching language shouldn't kick you out of scatter view.
+    activeCategory.value = "instruct";
```
saattrupdan: 'instruct' -> 'chat'

src/frontend/components/LeaderboardView.vue
```diff
@@ -61,21 +67,23 @@ const loadFor = async (stem: string) => {
...
   return allModelsTable.value;
 });
 
 const activeMetadata = computed<LeaderboardMetadata | null>(() => {
-  if (tab.value === "generative" || tab.value === "generative-scatter")
-    return generativeMetadata.value;
+  if (activeCategory.value === "generative") return generativeMetadata.value;
+  if (activeCategory.value === "instruct") return instructMetadata.value;
   return allModelsMetadata.value;
 });
```
saattrupdan: Naming again

src/frontend/components/LeaderboardView.vue
```diff
@@ -123,20 +131,65 @@ const MULTILINGUAL_STEMS = new Set([
...
-const tabs: { id: TabId; label: string }[] = [
-  { id: "generative", label: "Generative Leaderboard" },
-  { id: "generative-scatter", label: "Generative Scatter Plot" },
-  { id: "all_models", label: "NLU Leaderboard" },
-  { id: "all_models-scatter", label: "NLU Scatter Plot" },
+const categoryTabs: { id: CategoryId; label: string }[] = [
+  { id: "instruct", label: "Chat" },
+  { id: "generative", label: "Generative" },
+  { id: "all_models", label: "All Models" },
 ];
```
saattrupdan: Here the order is correct, but can we use 'chat' as ID for consistency?

src/frontend/md/leaderboards-README.md
```diff
@@ -10,13 +10,17 @@ left menu to see the results.</span>
...
+  (e.g. instruction following, tool use, bias evaluation).
 - **Generative Leaderboard**: This leaderboard shows the performance of models that can
-  generate text. These models have been evaluated on _all_ [tasks](/tasks), both NLU and
-  NLG.
-- **NLU Leaderboard**: This leaderboard shows the performance of models that can
-  understand text, which includes both generative and non-generative models.
+  generate text, evaluated on the standard set of NLU and NLG [tasks](/tasks). Any
+  generative model can appear here, whether base, instruction-tuned, or reasoning.
+- **All Models Leaderboard**: This leaderboard shows the performance of models that
+  can understand text, which includes both generative and non-generative models.
```
saattrupdan: Maybe clarify here: "can understand text (but not necessarily generate it)"

src/leaderboards/constants.py
```diff
@@ -194,12 +194,18 @@ def _env_path(name: str, default: Path) -> Path:
...
-# The two leaderboard categories that every model is ranked within. The
-# "generative" variant scores all tasks; "all_models" only scores NLU tasks so
-# non-generative models can compete.
-LEADERBOARD_CATEGORIES: tuple[str, str] = ("generative", "all_models")
+# The three leaderboard categories that every model is ranked within. The
+# "instruct" variant scores every task and only ranks instruction-tuned/
+# reasoning models; "generative" scores all tasks except the instruct-
+# exclusive ones; "all_models" only scores NLU tasks so non-generative models
+# can compete.
+LEADERBOARD_CATEGORIES: tuple[str, str, str] = ("instruct", "generative", "all_models")
```
saattrupdan: Naming

src/leaderboards/leaderboard_generation.py
```diff
@@ -29,7 +29,7 @@
 def generate_leaderboard(
     leaderboard_name: str,
     language_names: list[str],
-    categories: list[t.Literal["generative", "all_models"]],
+    categories: list[t.Literal["instruct", "generative", "all_models"]],
```
saattrupdan: Naming

src/leaderboards/leaderboard_generation.py
```diff
@@ -379,7 +379,7 @@ def _create_leaderboard_headers(
 def _generate_dataframe(
     model_results: dict[str, dict[str, list[tuple[list[float], float, float]]]],
     metadata_dict: dict[str, dict],
-    categories: list[t.Literal["generative", "all_models"]],
+    categories: list[t.Literal["instruct", "generative", "all_models"]],
```
saattrupdan: Naming

src/leaderboards/leaderboard_generation.py
```diff
@@ -437,6 +437,21 @@ def _generate_dataframe(
 
         data_dict: dict[str, list] = defaultdict(list)
         for model_id, results in model_results.items():
+            generative_type = metadata_dict.get(model_id, {}).get("generative_type")
+            if category == "instruct":
```
saattrupdan: Naming

src/leaderboards/leaderboard_generation.py
```diff
@@ -567,7 +582,7 @@ def _apply_display_transforms(
 
 
 def _build_category_dataset_maps(
-    categories: list[t.Literal["generative", "all_models"]],
+    categories: list[t.Literal["instruct", "generative", "all_models"]],
```
saattrupdan: Naming

src/leaderboards/records.py
```diff
@@ -146,6 +146,33 @@ def get_model_name(record: dict) -> str:
...
+
+    Returns:
+        True if the record is a few-shot evaluation, False if zero-shot.
+    """
+    additional = record.get("eval_library", {}).get("additional_details", {})
+    if "few_shot" not in additional:
+        return True
+    val = additional["few_shot"]
+    if val is None:
+        return False
```
saattrupdan, line 170: Shouldn't this be True, as it's the default?

src/leaderboards/task_metadata.py
```diff
@@ -38,20 +40,38 @@ def category_includes_task(category: str, task: str) -> bool:
     Returns:
         True if the task is scored within the category.
     """
-    return category == "generative" or task_category(task) == "nlu"
+    if category == "instruct":
```
saattrupdan: Naming

src/leaderboards/task_metadata.py
```diff
@@ -38,20 +40,38 @@ def category_includes_task(category: str, task: str) -> bool:
...
+        return True
+    if category == "generative":
+        return task_category(task) != "instruct_exclusive"
+    return task_category(task) == "nlu"
 
 
 def task_category(task_name: str) -> str:
```
saattrupdan, line 53: Can we use 'chat', 'generative' and 'all-models' again here, for consistency? Maybe it'd make sense to create an Enum, as it's used so many places.

src/leaderboards/task_metadata.py
```diff
@@ -38,20 +40,38 @@ def category_includes_task(category: str, task: str) -> bool:
...
 
     Returns:
-        ``"nlu"`` if the task's group is an NLU group, else ``"nlg"``.
+        ``"nlu"`` if the task's group is an NLU group, ``"instruct_exclusive"``
+        if it's restricted to instruction-tuned/reasoning models and not
+        orthogonal, else ``"nlg"``.
     """
     task = get_all_tasks()[task_name]
+    if (
+        task_name not in ORTHOGONAL_TASKS
```
saattrupdan, line 74: The orthogonal tasks should be moved to the chat leaderboards as well, since they all require instruction-tuned models anyway.

src/scripts/generate_leaderboards.py
```diff
@@ -56,11 +56,11 @@
...
-    default=("generative", "all_models"),
+    default=("instruct", "generative", "all_models"),
     multiple=True,
     help=(
-        "Categories to generate leaderboards for. Defaults to 'generative' and "
-        "'all_models'."
+        "Categories to generate leaderboards for. Defaults to 'instruct', "
+        "'generative', and 'all_models'."
     ),
 )
```
saattrupdan, line 66: Naming

saattrupdan, line 66: Naming

src/scripts/generate_leaderboards.py
```diff
@@ -100,7 +100,7 @@
...
 def main(
-    categories: tuple[t.Literal["generative", "all_models"], ...],
+    categories: tuple[t.Literal["instruct", "generative", "all_models"], ...],
```
saattrupdan: Naming

src/scripts/generate_leaderboards.py
```diff
@@ -110,8 +110,8 @@ def main(
 
     Args:
         categories (optional):
-            Categories to generate leaderboards for. Defaults to 'generative' and
-            'all_models'.
+            Categories to generate leaderboards for. Defaults to 'instruct',
```
saattrupdan: Naming

src/scripts/swap_leaderboard_dataset.py
```diff
@@ -2111,25 +2119,27 @@ def ranked_model_language_pairs(
...
+        required_sets: list[set[str]] = []
         for category in affected:
-            required |= {
+            required = {
                 dataset
                 for task, task_datasets in by_task.items()
                 if task not in ORTHOGONAL_TASKS
                 and category_includes_task(category=category, task=task)
                 for dataset in task_datasets
             }
```
saattrupdan, line 2133: Why overwrite `required` rather than take unions?

saattrupdan (changes requested):

src/leaderboards/task_metadata.py
viggo-gascou: Just to confirm: should `european-values` (and other orthogonal tasks) be removed from the generative bonus columns, so that they only show on chat?

src/scripts/swap_leaderboard_dataset.py
viggo-gascou: Before, this required a model to have every dataset across all affected categories combined, but that didn't match what the docstring above it says, so I changed it to match the docstring, meaning that a model just needs to be ranked in any one affected category.

src/leaderboards/records.py
viggo-gascou: Explicit null in the results json means the task forced zero-shot (e.g., ﻿instruction-following, tool-calling, etc.), so it should be False here since the value in the json was "null".

src/leaderboards/task_metadata.py
saattrupdan: Yep!

src/leaderboards/task_metadata.py
saattrupdan: Unresolving this, as it doesn't seem like this change was made?

src/scripts/generate_leaderboards.py
```diff
@@ -56,11 +57,11 @@
 @click.option(
     "--categories",
     "-c",
-    default=("generative", "all_models"),
+    default=LEADERBOARD_CATEGORIES,
```
saattrupdan, line 60: I'm not actually sure how click handles this: Can it natively deal with Enums here?

src/leaderboards/task_metadata.py
viggo-gascou: It was missing on my end before too, but it should be there now? unless you meant something else? 😄

src/scripts/generate_leaderboards.py
viggo-gascou: It seems to handle it fine, at least when logging the value of `categories` it gives `('chat', 'generative', 'all_models')`. And when running the full generation all leaderboards are generated 😄

saattrupdan:
@viggo-gascou Just tried compiling the leaderboards and frontend locally, and it generally looks fine (of course lots of them are missing right now). But I noticed that the "(zero-shot)" suffix is still shown for models on the Chat leaderboard, which shouldn't be the case since all of them are zero-shot anyway.

Also, can we make the default leaderboard tab to be the first non-empty one. So if Chat is empty (as in, no models have any rank score) then we should move on to Generative, and lastly All Models. Alternatively, which is maybe even nicer, simply hide the Chat tab if there are no ranked models on the Chat leaderboard.

viggo-gascou:
For reference, here's what changed since the review:

- Added a small generation step (`category-ranked.json`) that precomputes which leaderboard categories have ranked models, so the frontend can determine tab state synchronously, before any leaderboard data loads.
- Disabled the Chat tab when it has no ranked models yet, shown dimmed with a "Soon" label. Hiding it entirely caused the tab count to vary between leaderboards, which broke the sliding-pill indicator. Disabling keeps a fixed tab set everywhere and avoids that.
  Screenshot 2026-08-18 at 14 49 41
- Dropped the `(zero-shot)` suffix from Chat rows, since it's redundant there, every row is zero-shot by construction. 
- Updated the docs to specify the zero/few-shot conventions per leaderboard category.

**Note:** Since European Values only shows on the Chat leaderboard, and Chat barely has any ranked models yet, its scores aren't visible anywhere right now until more models get ranked on the Chat leaderboard.

saattrupdan (approved):
LGTM

PR #2147, opened 2026-08-18
Fix infinite loop in token-classification few-shot example selection

jaideeppyne:
## Problem

`_extract_token_classification_examples` (`src/euroeval/generation_utils.py`) selects class-balanced NER few-shot examples by cycling the entity (`B-`) labels:

```python
labels = it.cycle([l.lower() for l in dataset_config.labels if l.lower().startswith("b-")])
while len(few_shot_examples) < num_few_shots and len(shuffled_train) > 0:
    label = next(labels)
    possible_examples = shuffled_train.filter(lambda x: label in [str(t).lower() for t in x["labels"]])
    ...
    if len(possible_examples) == 0:
        continue          # <-- nothing removed from shuffled_train
```

`it.cycle` loops over the `B-` labels forever, and `shuffled_train` only shrinks when an example is **successfully picked**. So if the loop reaches a state where **no remaining training example contains any entity** (only `O`-tagged sentences left) while `len(few_shot_examples) < num_few_shots` and `len(shuffled_train) > 0`, every `next(label)` matches nothing, hits `continue`, and the loop **spins forever** — a hard hang, no error, no timeout.

This is reachable with sparse-entity or custom/community NER splits (fewer entity-bearing sentences than `num_few_shot_examples`, which EuroEval explicitly supports).

**It's an oversight, not intended:** the sibling `_extract_classification_examples` guards the *identical* pattern with a `labels_with_no_samples` set and stops once all labels are exhausted. The token-classification twin was missing that guard.

## Fix

Add the same guard: record labels that yield no samples, and `break` once every entity label is exhausted. The guard is sound because `shuffled_train` is monotonically shrinking — a label with zero matches can never regain matches — so `labels_with_no_samples` only grows and the `== len(b_labels)` check terminates. It also removes a latent `StopIteration` when there are no `B-` labels (the guard fires before `next()`). Behavior on datasets with abundant entity examples is unchanged.

## Test

`tests/test_generation_utils.py::test_token_classification_few_shot_terminates_with_sparse_entities` builds a split with a single entity-bearing example, asks for `num_few_shots=5`, and asserts the call **returns** (under a 30s `SIGALRM` guard). It hangs on the current code and returns with the fix.

> Note: I validated the loop-termination logic directly (a faithful standalone replica of the loop hangs >10k iterations before the fix and returns in 3 with it). I didn't spin up EuroEval's full runtime (torch etc.) locally, so CI is the end-to-end check on the real `Dataset` path.


saattrupdan (changes requested):
Thanks for fixing this hang; the no-sample guard addresses a real production issue with sparse/O-only NER examples.

Before merging:

1. **Duplicate/case-variant labels:** Comparing `len(labels_with_no_samples)` with `len(b_labels)` can still hang when labels contain variants such as `b-per` and `B-PER`. Please deduplicate normalized B labels or compare against a normalized set.
2. **Changelog:** Add an entry under `CHANGELOG.md` → `[Unreleased]` → `Fixed`, as this changes `src/euroeval/`.

Also consider strengthening the regression assertion from `len(result) <= 1` to exactly one result. The timeout scaffolding is justified for a hang regression.

jaideeppyne:
Both addressed in the latest push. The lowercased `b_labels` wasn't deduplicated, so `b-per`/`B-PER` inflated its length past what the `labels_with_no_samples` set could ever reach and the guard never fired — switched to `dict.fromkeys(...)` to drop the normalized duplicates while preserving cycle order. Also tightened the regression assertion from `<= 1` to exactly `1`, and added the changelog entry.

saattrupdan:
> Both addressed in the latest push. The lowercased `b_labels` wasn't deduplicated, so `b-per`/`B-PER` inflated its length past what the `labels_with_no_samples` set could ever reach and the guard never fired — switched to `dict.fromkeys(...)` to drop the normalized duplicates while preserving cycle order. Also tightened the regression assertion from `<= 1` to exactly `1`, and added the changelog entry.

Great, thanks. Can you ensure that the code checks pass (you can run `make check` locally to run these), and also update the branch from `main`?

jaideeppyne:
Done — checks are green now (the `code-check` failure was the `funcsort` hook wanting the test functions reordered; that's fixed, and `pytest` 3.12/3.13 pass). Also merged the latest `main` into the branch — no conflicts, the upstream changes were all in the leaderboard code.

saattrupdan (approved):
Looks great now, thanks a lot for the fix!

PR #2149, opened 2026-08-18
Fix: Task page links and add missing tasks to nav

viggo-gascou:
Task links in the tasks overview page pointed to `.md` files instead of the actual routes, so they didn't resolve correctly. Also added a some tasks that had pages but weren't linked anywhere, in both the tasks readme and the sidebar nav config.


saattrupdan (approved):
LGTM

Issue #900, opened 2025-04-04
[FEATURE REQUEST] Add release date by performance plot to leaderboard


KennethEnevoldsen:
### 🚀 The feature, motivation and pitch

Add release date by performance plot to leaderboard.

Something like:
![Image](https://github.com/user-attachments/assets/ddf255b2-74d5-41fc-8ed9-ccd1374e7813)

This would require us to add the release date to the current models. I could imagine that we could fetch most of these from Huggingface, but a few would have to be annotated.

EwoutH:
This would be really cool!

milos-plavsic:
Hello, I'm Milos from the Education Improvement Institute (EII). I'd be interested in implementing this issue as part of our work around transparent and responsible European ML evaluation.

My proposed approach is to:

- trace the existing model metadata and leaderboard data pipeline;
- add a normalized release-date field;
- retrieve dates from Hugging Face metadata where that source is considered reliable;
- support explicit annotations for models without a usable upstream date;
- add a performance-versus-release-date visualization using the project's existing frontend patterns; and
- add tests for missing, malformed, and manually overridden dates.

Before I start, how should “release date” be defined here: the first public model release, the Hugging Face repository creation date, or another field? Also, would you prefer the metadata collection and visualization in one pull request or as two smaller pull requests?

If this scope matches your expectations, I'd be happy to work on it.

saattrupdan:
Hey @milos-plavsic, and thanks for wanting to contribute!

First and foremost, this would depend on https://github.com/EuroEval/EuroEval/pull/2104 being merged, but that should happen soon. This plot could then be added as a tab next to the current scatter plot (and "scatter plot" should be renamed to something more descriptive).

For all models on Hugging Face Hub, which are by far most of them, we should be able to use the date for the first model weights file (so we should search for both safetensors and pytorch files here). For API models I think we have to include these manually, like we do with the other API metadata (they're in the `src/euroeval/benchmark_modules/litellm.py` module).

This would probably require adding this metadata to `BenchmarkResult` and ensure backwards compatibility. Then the metadata would be part of the output score file, and will be uploaded to the [results bucket](https://huggingface.co/buckets/EuroEval/results) for the leaderboards.

And I think doing it in two separate PRs make sense.

milos-plavsic:
Thanks, @saattrupdan — that scope and the two-PR split make sense.

I will wait for #2104 to merge, then approach this as:

1. **Release-date metadata PR**
[...]

PR #2154, opened 2026-08-24
feat: add model release date metadata

milos-plavsic:
## Summary

Implements the metadata portion of #900, providing the release-date data needed for the planned performance-over-time leaderboard visualization.

- add an optional, backwards-compatible `release_date` field to benchmark results and model configuration
- determine Hugging Face model release dates from the earliest repository commit containing Safetensors or PyTorch model weights
- maintain curated release dates for API model aliases and parse dates from versioned API model IDs
- serialize release dates in EEE output and restore them when reading historical results
- backfill missing release dates while processing existing leaderboard records, with per-model caching
- preserve valid historical dates and gracefully handle missing, malformed, or unavailable metadata
- include release dates in leaderboard metadata-richness deduplication
- add regression coverage for Hugging Face history, API aliases and dated IDs, serialization, benchmarking propagation, caching, and historical backfilling

## Verification

- `169 passed, 199 skipped` in the focused affected test suite (the skips are dependency-marked benchmark tests whose prerequisite modules were not included in the focused invocation)
- all pre-commit hooks pass, including Ruff, formatting, ty, vulture, function ordering, and Markdown lint
- `git diff --check` passes

I am eager to contribute more and would be happy to follow up with the visualization portion or address any review feedback.

### DIFF
--- a/src/euroeval/benchmark_modules/hf.py
+++ b/src/euroeval/benchmark_modules/hf.py
@@ -1439,7 +1439,9 @@ def get_model_repo_info(
 
     # Try to get model info from local directory first
     model_info: HfApiModelInfo | None = None
-    if Path(model_id).is_dir():
+    release_date: str | None = None
+    is_local_model = Path(model_id).is_dir()
+    if is_local_model:
         model_info = _get_local_model_info(model_id=model_id)
         if model_info is None:
             return None
@@ -1453,6 +1455,9 @@ def get_model_repo_info(
         )
         if model_info is None:
             return None
+        release_date = _get_model_release_date(
+            hf_api=hf_api, model_id=model_id, revision=revision, token=token
+        )
 
     # Handle adapter models - get base model tags
     tags = model_info.tags or list()
@@ -1493,7 +1498,10 @@ def get_model_repo_info(
         return None
 
     return HFModelInfo(
-        pipeline_tag=pipeline_tag, tags=tags, adapter_base_model_id=base_model_id
+        pipeline_tag=pipeline_tag,
+        tags=tags,
+        adapter_base_model_id=base_model_id,
+        release_date=release_date,
     )
 
 
@@ -1657,6 +1665,36 @@ def _get_local_model_info(model_id: str) -> HfApiModelInfo | None:
             level=logging.WARNING,
         )
         return None
+
+
+def _get_model_release_date(
+    hf_api: HfApi, model_id: str, revision: str, token: str | None
+) -> str | None:
+    """Return the date when model weights first appeared in a Hub repository."""
+
+    def contains_weights(files: c.Iterable[str]) -> bool:
+        return any(
+            path.endswith(".safetensors")
+            or re.search(r"(?:^|/)(?:adapter|pytorch)_model.*\.bin$", path) is not None
+            for path in files
+        )
+
+    try:
+        commits = hf_api.list_repo_commits(
+            repo_id=model_id, revision=revision, token=token
+        )
+        for commit in reversed(commits):
+            files = hf_api.list_repo_files(
+                repo_id=model_id, revision=commit.commit_id, token=token
+            )
+            if contains_weights(files):
+                return commit.created_at.date().isoformat()
+    except (HfHubHTTPError, HFValidationError, OSError, RequestException) as e:
+        log(
+            f"Could not determine the release date for {model_id!r}: {e}",
+            level=logging.DEBUG,
+        )
+    return None
 
 
 def _get_tags_for_adapter_model(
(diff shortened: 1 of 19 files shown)


src/euroeval/benchmark_modules/hf.py
```diff
@@ -1659,6 +1667,36 @@ def _get_local_model_info(model_id: str) -> HfApiModelInfo | None:
         return None
 
 
+def _get_model_release_date(
```
saattrupdan: This function should be public, as it's imported in other modules. Also missing a Google-style docstring (Args and Returns sections are missing).

src/euroeval/benchmark_modules/litellm.py
```diff
@@ -166,6 +167,104 @@
...
+    r"(xai/)?grok-2": "2024-08-13",
+    r"(xai/)?grok-3(?:-beta)?": "2025-02-19",
+    r"(xai/)?grok-3-mini.*": "2025-04-09",
+    r"(xai/)?grok-4": "2025-07-09",
+    r"(xai/)?grok-4-fast.*": "2025-09-19",
+    r"(xai/)?grok-4\.1.*": "2025-11-17",
+}
+
+
+def _get_api_model_release_date(model_id: str) -> str | None:
```
saattrupdan: This function should be public, as it's imported in other modules.

saattrupdan: Can you move this helper function below the constants?

src/leaderboards/model_metadata.py
```diff
@@ -493,6 +496,50 @@ def is_trained_from_scratch(
         logger.error("Invalid input. Please try again.")
 
 
+def backfill_release_dates(records: list[dict], cache: Cache) -> list[dict]:
```
saattrupdan: We already have a function to add missing metadata: `add_missing_entries`. I think we should include release date in that function rather than create this new separate one.

src/euroeval/benchmark_modules/litellm.py
```diff
@@ -166,6 +167,104 @@
 }
 
 
+# Release dates for API aliases which do not carry a dated version in their ID.
+# Dated model IDs are handled generically by `_get_api_model_release_date` below.
+# Sources are the providers' model launch posts and model changelogs:
+# https://platform.openai.com/docs/models, https://docs.anthropic.com/en/docs/about-claude/models,
+# https://ai.google.dev/gemini-api/docs/models, and https://docs.x.ai/docs/models.
+MODEL_RELEASE_DATE_MAPPING = {
```
saattrupdan, line 175: Would be great to have all the existing models in here. For instance, the 5-series of Claude isn't here, same thing with Grok-4.6. If we could just add all models that are on the provider model pages right now that'd be great.

saattrupdan (changes requested):
Overall really good! Just a few smaller refactoring and documentation changes.

src/euroeval/benchmark_modules/hf.py
milos-plavsic: Done in 5df66750. Renamed the helper to the public function get_model_release_date, updated all imports and callers, and added complete Google-style Args and Returns sections.

src/euroeval/benchmark_modules/litellm.py
milos-plavsic: Done in 5df66750. Renamed this helper to the public function get_api_model_release_date and updated all production and test references.

milos-plavsic: Done in 5df66750. get_api_model_release_date now appears after the complete constants section. The repository function-ordering hook places it with the module-level helpers below the class.

src/leaderboards/model_metadata.py
milos-plavsic: Done in 5df66750. Release-date enrichment now runs inside add_missing_entries; the separate batch helper and result-processing call were removed. Existing valid dates remain authoritative, malformed values are repaired, and both successful and failed lookups are cached per model.

src/euroeval/benchmark_modules/litellm.py
milos-plavsic: Done in 5df66750. I audited the current official provider catalogs and expanded the benchmarkable text-model coverage, including the Claude 4.x/5 families, Gemini 3.1/3.5/3.6/3.7 text models, Grok 4.20/4.5/4.6, and current OpenAI flagship/pro variants. I also added parameterized regression cases for the new families; media-only models were intentionally excluded because EuroEval cannot evaluate them.

milos-plavsic:
All requested review changes are now addressed in 5df66750, and each inline conversation has been updated and resolved. The branch is synchronized with current main. Verification: 186 passed and 199 dependency-marked tests skipped in the affected suite; all pre-commit hooks, Ruff formatting, ty, vulture, function ordering, Markdown lint, and diff checks pass. @saattrupdan, when convenient, could you please take another look?

saattrupdan (approved):
LGTM, thanks!

PR #2156, opened 2026-08-26
fix: validate release dates during deduplication

milos-plavsic:
## Summary

Follow-up to #2154 and the Copilot review comment in [this discussion](https://github.com/EuroEval/EuroEval/pull/2154#discussion_r3861127348).

- count `release_date` as rich metadata only when it is a real calendar date in the canonical `YYYY-MM-DD` format
- share one release-date normalizer across model validation, metadata enrichment, cache handling, and deduplication
- ensure a valid release date wins over a malformed value when otherwise identical records are deduplicated
- cover missing, empty, malformed, impossible, compact, week-date, date-object, and datetime-object inputs

## Verification

- `96 passed` in the focused date, data-model, record-field, and metadata-enrichment test suite
- `make check` passes, including Ruff, formatting, ty, vulture, funcsort, and Markdown lint
- `git diff --check` passes

### DIFF
--- a/src/euroeval/data_models.py
+++ b/src/euroeval/data_models.py
@@ -275,19 +275,9 @@ class BenchmarkResult(pydantic.BaseModel):
         """
         return benchmark_result_from_eee_dict(config=config)
 
-    @pydantic.field_validator("release_date", mode="before")
-    @classmethod
-    def normalise_release_date(cls, value: object) -> str | None:
-        """Normalise release dates and discard malformed historical metadata.
-
-        Args:
-            value:
-                The release date value being validated.
-
-        Returns:
-            An ISO-formatted date, or None when the value is absent or malformed.
-        """
-        return normalise_release_date(value)
+    _normalise_release_date = pydantic.field_validator("release_date", mode="before")(
+        normalise_release_date
+    )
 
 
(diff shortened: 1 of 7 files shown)


src/euroeval/data_models.py
```diff
@@ -287,16 +287,7 @@ def normalise_release_date(cls, value: object) -> str | None:
...
-            return value.date().isoformat()
-        if isinstance(value, datetime.date):
-            return value.isoformat()
-        if not isinstance(value, str):
-            return None
-        try:
-            return datetime.date.fromisoformat(value).isoformat()
-        except ValueError:
-            return None
+        return normalise_release_date(value)
```
saattrupdan: Seems a bit silly to have this class method if it's now identical to the helper method. Can't we add the field validation with the function directly?

saattrupdan (changes requested):
Small change, otherwise looks great

milos-plavsic:
Good point. I removed the redundant class method and registered the shared helper directly as the Pydantic field validator. The targeted data model tests pass, and Ruff is clean.

saattrupdan (dismissed):
Looks great, thanks!

PR #2162, opened 2026-08-31
feat: visualize leaderboard performance over time

### LINKED ISSUE #900: [FEATURE REQUEST] Add release date by performance plot to leaderboard

milos-plavsic:
## Summary

- add a Release Date visualization beside the existing plot and rename the existing view to Model Size
- reuse the current scatter plot's search, model filters, markers, tooltips, responsive SVG, and sticky view selection
- carry the normalized `release_date` metadata into the full leaderboard CSV while keeping it out of the table columns
- omit models without usable dates and show an explicit empty state when a leaderboard has no dated models

Closes #900

## Verification

- `npm run check`
- `npm run build`
- `33 passed` in the focused score-extraction and leaderboard-generation test suite
- all pre-commit hooks pass on the changed files
- `git diff --check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,9 @@ project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
   few-shot and validation splits. This has been fixed.
 - Fixed leaderboard deduplication treating malformed model release dates as valid
   metadata.
+- Fixed model release dates being missing for most models: a rate-limited Hub request
+  was stored as "no release date" and never asked again, and models without a recorded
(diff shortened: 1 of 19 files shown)


saattrupdan (approved):
Looks great, thanks! :)

milos-plavsic:
@saattrupdan Thanks for the review and the  merge. Glad this made it in.

I'll delete the branch on my side. Let me know if any issues pop up with the visualization in production or if I can contribute elsewhere.

PR #2167, opened 2026-09-03
feat: support Inspect AI math scoring and multi-config eval.yaml subsets

saattrupdan:
## Why

`danish-foundation-models/multilingual-gsm-symbolic` ships an Inspect AI `eval.yaml` whose distinguishing features EuroEval could not handle: a `scorers: [{name: math}]` entry, and 28 task entries declaring one configuration per language × split. Loading it failed on unsupported language metadata, and even when it loaded only a single entry was reachable.

## Changes

**`math` scorer.** A `scorers` entry named `math` loads as a new `math` task (`TaskGroup.TEXT_TO_TEXT`, 0 few-shot examples, `default_max_generated_tokens=1024`, generative models only). `src/euroeval/metrics/math.py` extracts the answer the way Inspect AI does — last balanced `\boxed{...}`, then `answer is`/`final answer`/`result` markers, then delimited math, then the whole text, last line, last number. Plain numbers are compared exactly, percentages by value (`50\%` matches `0.5`, not `50`), and an answer boxed as an equation (`\boxed{x = 5}`) scores the value it names.

**Expressions are compared as values**, so `\boxed{\frac{1}{2}}` meets `0.5`, `\boxed{-0.25}` meets `-1/4` and `\boxed{2\pi}` meets `6.283185307179586`. Inspect AI reaches these through SymPy behind an ANTLR LaTeX grammar (`latex2sympy2-extended`). SymPy itself is no obstacle — it ships with torch, so it is in every EuroEval environment, and is now declared rather than relied on transitively — but the grammar is, and benchmark answers do not need one. This module rewrites the LaTeX they actually use (fractions, roots, powers, `\cdot`/`\div`/`\times`, thin-space thousands grouping, percentages, implicit products like `2\pi`) into plain arithmetic, parses that with `ast` under SymPy-guarding limits (candidate length, digit count, node count and depth, no exponent towers), and compares it with the same rules Inspect uses: structural equality, then `equals()`, then a 1e-10 tolerance that is refused for values SymPy compares exactly. Anything the rewrite cannot render faithfully reports itself unknown and falls back to text equality, so matrices, integrals and prose are scored exactly as before rather than guessed at. A `prompt_template` solver becomes the EuroEval instruction prompt (`{prompt}` → `{text}`, doubled braces preserved), but only for free-form generation tasks and only when the template uses no other placeholders — so existing Inspect-style multiple-choice configs keep EuroEval's own prompts instead of crashing on `KeyError: 'answer'`. A `math` config whose template never mentions boxing gets a warning.

**Subsets.** Every task entry in an `eval.yaml` becomes a dataset of its own, so `--dataset repo` benchmarks all of them; `repo::split` narrows the run to that split of every configuration; `repo::config::split` names one subset exactly:

```bash
euroeval -m <model> --dataset danish-foundation-models/multilingual-gsm-symbolic                 # 26 datasets
euroeval -m <model> --dataset danish-foundation-models/multilingual-gsm-symbolic::dan::test_original   # 2
euroeval -m <model> --dataset danish-foundation-models/multilingual-gsm-symbolic::test_original  # 13
euroeval -m <model> --dataset danish-foundation-models/multilingual-gsm-symbolic::dan::test_original  # 1, by name
```

Names are `<repo>::<config>::<split>` and `source` stays `<repo>::<config>` (`data_loading.py` reads the Hub config name from it), so a recorded result name is valid input. The two-part name is not cosmetic: the dataset ID keys the stored result and a result record carries no language, so subsets sharing a name would overwrite each other. Language attribution: a per-entry `languages` key, then the config name as an ISO 639-1 or ISO 639-3 code with any `_suffix` stripped (`eng_metric` → `en`), resolved through the `code_3` each `Language` now carries (no new dependency; trusted only when at least half the configs resolve, so `default`/`train` configs are unaffected), then the repository card with a warning. A config naming a language EuroEval lacks is skipped with a warning — `zho` here, which is why the reference repo yields 26 datasets rather than 28. Split discovery is now per configuration, which fixes a phantom `train_original` that made `KeyError: 'train_original'`. Unsupported language codes in Hub metadata no longer abort the rest of an `eval.yaml`.

**CLI.** `--dataset` with `--task` was a raw `ValueError` traceback; it is now a `click.UsageError` (exit 2). `--dataset` with `--language` is now the same kind of usage error: the dataset ID already states which configurations and splits to benchmark, so the flag had nothing left to say. Naming a configuration where a split belongs lists the subsets it stands for, instead of recommending a flag the CLI refuses.

## Compared against Inspect AI

Reference: `inspect_ai` 0.3.263 in a separate venv, running the dataset's own `eval.yaml` (`inspect eval hf/danish-foundation-models/multilingual-gsm-symbolic/original_dan`), `openai/gpt-4o-mini`, temperature 0, `max_tokens=1024`.

| subset | Inspect AI | EuroEval | agreement on identical generations |
| --- | --- | --- | --- |
| `dan`/`test_original` | 91.0% (93.0 / 90.0 across two epochs) | 93.90% ± 1.22 | 100/100 |
| `eng`/`test_original` | 95.0% | 95.50% ± 1.31 | 100/100 |

The rendered prompts are byte-identical, and scoring Inspect's own completions with `euroeval.metrics.math` reproduces Inspect's verdicts exactly on all 200 samples. The end-to-end gaps are generation noise, not scoring: Inspect rerun on the same Danish subset with `--epochs 2` gave 93.0% and 90.0% at temperature 0, so ±2 pp is what this model and API give run to run.

GSM-Symbolic targets are plain numerals, so that test cannot reach the symbolic cases; those were scored separately on 30 engineered answers, run through Inspect's own `math()` scorer over `mockllm` and through `euroeval.metrics.math`. Agreement is 29/30. It was 26/30 before the SymPy path landed — the cases it fixed were `\boxed{\frac{1}{2}}` vs `0.5`, `\boxed{-\frac{1}{4}}` vs `-0.25`, `\boxed{2\pi}` vs `6.283185307179586`, `\boxed{\sqrt{8}}` vs `2\sqrt{2}`, `\boxed{2^{10}}` vs `1024`, `\boxed{\frac{1}{2} + \frac{1}{4}}` vs `0.75` and `\boxed{3 \cdot 4}` vs `12`.

The one holdout is `\boxed{\text{Yes}}` against `Yes`: Inspect calls it incorrect, EuroEval strips `\text{}` and calls it correct. Kept deliberately — a boxed word that says the right thing should score.

Cases where both scorers agree, worth recording because they are the ones a looser implementation gets wrong: `50\%` matches `0.5` and not `50`; `3.0` matches `3`; `1,000` matches `1000` but not `1001`; `99999999999` does not match `100000000000`; `0.5` does not match `1/3`; `2\pi` does not match `6.28`.

## Review round

A review of the branch found inputs that broke scoring rather than merely differing from Inspect AI:

- `1 % 0` and `"2^{" * 1000` raised `ZeroDivisionError` and `RecursionError` out of the metric, which drops the whole model×dataset evaluation including generation time; `(1,2)^2` raised `AttributeError`. A `%` away from the end was reaching the parser as Python's modulo, so `50% + 10%` scored as a remainder.
- `2**-1000000000` bypassed the exponent guard — 3.1 s and 467 MB for fifteen characters of model output, unbounded above — because the guard read `ast.Constant` instead of measuring the exponent. The positive form was already refused in 1 ms.
- A percentage was divided under the default 28-digit decimal context, so a 30-digit percentage equalled its neighbour.
- `CO2` no longer matched `co2`: answers that parse to names rather than values were being compared as algebra. Values still carrying names are referred back to text — the same rule that keeps `Yes` a word.
- A repo with configs `default`, `dan`, `swe` benchmarked two of three, warning about an unsupported language for a name that was never one. Skipping now needs the shape of a language code.
- Plus the merge leftovers: a duplicated brace scanner (the only duplicate cluster in the changed files; slopo's remaining clusters are all in `scripts/dataset_creation`, untouched here), three stale docstrings, and a case-sensitive `get_language`.

One review finding was then reverted, because measuring it showed the finding itself was the divergence from Inspect AI: referring answers that carry names (`CO2`, `T-shirt`, `Ja`) back to text comparison made `x + y` stop matching `y + x`. Inspect AI reads a candidate that is not prose as a name and compares names case-sensitively — it scores `CO2` vs `co2` and `Ja` vs `ja` incorrect — so the gate is now prose-vs-value, and words behave identically here and there.

### Agreement with Inspect AI's `math` scorer

| Corpus | Agreement |
| --- | --- |
| 200 real `gpt-4o-mini` generations (`original_dan`, `original_eng`) | 200/200 |
| 30 engineered symbolic cases | 29/30 |
| 30 engineered identity and prose cases | 27/30 |

Four deliberate divergences, each pinned by a test in `TestKnownDivergences` or `TestPercentages`: `oxed{	ext{Yes}}` vs `Yes` (we call it correct); a cancelled ratio (SymPy simplifies as it builds, Inspect AI builds unevaluated and abstains); and `50% + 10%`, which Inspect AI reads as a remainder that happens to equal 0 and we refuse to read as a remainder at all.

## Review round 2

Nine threads, six of them about the same thing: configurations were attributed to a language so `--language` could select among them, which left language selection in two places. Resolved in favour of the dataset ID (72d361ef):

- `--language` conflicts with `--dataset`, as `--task` already did. It still filters a `dataset` passed to the `Benchmarker` constructor, which the docstring now says explicitly — the filter lives in the config factory, not the CLI.
- `Language` carries both its ISO 639-3 code and, where it has one, its ISO 639-1 code, with `code` preferring the shorter; the 178-entry alias table this branch added is gone. All 319 registry keys are unchanged, which the tests assert, since a language re-keying to its 639-3 code would silently move its stored results. `por` resolves to Portuguese through this; `zho` resolves to nothing rather than to whichever Chinese is defined first.
- A solver this loader cannot read no longer aborts the scan of the whole `solvers` list (Copilot); the missing-boxed-answer warning no longer misses `\fbox` (Copilot).
- Docs: the language-attribution paragraph is removed, and the `math` scoring notes moved to a `### The math task` section of their own.
- Kept, against two comments: `repo::config::split` and the three-part selector, for the results-identity reason above.

## Tests

- `pytest tests/test_metrics tests/test_yaml_config.py tests/test_languages.py tests/test_benchmark_config_factory.py tests/test_cli.py tests/test_split_utils.py tests/test_data_loading.py` → 285 passed, 2 skipped; `tests/test_benchmarker.py tests/test_dataset_configs.py` → 70 passed, 177 skipped.
- `pre-commit run` clean on all changed files, including `ty`. Slopo run: no duplicate cluster touches a changed file.
- Manual: the four CLI invocations above against the live dataset; `::dan` errors with the `--language 'da'` hint; `--dataset dansk --task classification` exits 2 with a usage error.

Follow-up worth a separate issue: `zho` is dropped because EuroEval has no Chinese, so a multilingual dataset silently loses a language rather than telling the user upfront.


src/euroeval/benchmark_config_factory.py
```diff
@@ -243,6 +253,44 @@ def _handle_dataset_lookup_error(
     sys.exit(1)
 
 
+def _resolve_dataset_id(
+    dataset_id: str, all_dataset_configs: dict[str, DatasetConfig]
+) -> list[DatasetConfig]:
+    """Look up a requested dataset, expanding the subsets of an external repo.
+
+    An external dataset repository registers one config per task entry in its
+    `eval.yaml`, named `<repo>::<config>::<split>`. Requesting the repository itself
```
saattrupdan, line 267: Configs shouldn't be in the dataset ID, only splits

src/euroeval/cli.py
```diff
@@ -273,6 +274,8 @@ def benchmark(
     vocabulary_size: int | None,
 ) -> None:
     """Benchmark pretrained language models on language tasks."""
+    _validate_task_and_dataset_options(task=task, dataset=dataset)
```
saattrupdan: Why not just have those two lines here instead of using a function?

src/euroeval/languages.py
```diff
@@ -1072,3 +1072,232 @@ def get_all_languages() -> dict[str, Language]:
 CANTONESE: Language = Language(
     code="yue", name="Cantonese", _and_separator="同", _or_separator="或者"
 )
+
+
+# ISO 639-3 codes for the languages in this registry that are keyed by their ISO
+# 639-1 code. Datasets are commonly subdivided by their 639-3 code, e.g. a Hugging
+# Face dataset configuration named `dan`, so these aliases make such names resolvable
+# to a language. Codes that are already registry keys are not repeated here. Every
+# value is a key of `get_all_languages`, which is verified by a unit test.
```
saattrupdan: Can we instead just have both codes in the `Language` object, where the 2-letter one can be None if not present. They should still have a `code` property, which uses the 2-letter if present, and otherwise the 3-letter one.

src/frontend/md/python-package.md
```diff
@@ -458,6 +458,43 @@ The value of `task` must be one of the task names used in EuroEval
...
+```text
+Unknown split 'dan' for dataset '...::dan'. Available splits are: ['test_original',
+'test_synthetic']. To select the 'dan' configuration, use --language 'da' instead.
+```
+
+Configurations named by an ISO 639-1 or ISO 639-3 language code, optionally with a
+suffix such as `eng_metric`, are attributed to that language. Otherwise EuroEval
+attributes the results to all the languages in the repository card and warns about it,
+and a configuration referring to a language EuroEval does not support is skipped with a
+warning, as it cannot be attributed to any language.
```
saattrupdan: Remove this

src/frontend/md/python-package.md
```diff
@@ -492,10 +529,33 @@ infer them automatically when they are absent:
...
+benchmark answers use — fractions, roots, powers, products, `\pi` — into
+arithmetic and evaluates it with SymPy, which is already installed as a
+dependency of torch, so `\frac{1}{2}` and `0.5` are treated as equal, as are
+`2\pi` and `6.283185307179586`. Values that SymPy compares exactly, such as
+rationals and `\sqrt{2}`, are never compared approximately. This is a rewrite
+rather than Inspect AI's full LaTeX grammar, so structure beyond it — matrices,
+integrals, piecewise braces — is compared as normalised text. Since the
+extraction relies on the boxed answer, a `math` configuration should include a
+`prompt_template` solver asking the model to box its answer — EuroEval warns
+when it is missing.
```
saattrupdan: This doesn't make sense here, should just go in the `math` task docs (create it if it doesn't already exist)

tests/test_benchmark_config_factory.py
```diff
@@ -289,3 +318,40 @@ def test_prepare_languages(
...
+
+
+def test_resolve_dataset_id_expands_external_subsets() -> None:
+    """Test that a repo or split request selects all of its registered subsets."""
+    configs = {}
+    for name in [
+        "dansk",
+        "repo::dan::test_original",
+        "repo::dan::test_synthetic",
+        "repo::deu::test_original",
```
saattrupdan, line 330: We don't want double :: in the dataset IDs, only one, and it should separate the base dataset ID with the split.

src/euroeval/yaml_config.py
```diff
@@ -444,6 +669,191 @@ def infer_task_from_inspect_ai(
...
+            message=(
+                f"Invalid dataset selector {dataset_id!r}. Use the syntax "
+                "<repo>[::<split>] to select a split, or the name of one of the "
+                "expanded subsets, <repo>::<config>::<split>."
+            ),
+            level=logging.ERROR,
+        )
+        return None
+    if len(parts) == 3:
+        return parts[0], parts[1], parts[2]
```
saattrupdan: We don't allow 3 parts. This function should be removed and we should just split by :: and check only two parts are there.

src/euroeval/benchmark_config_factory.py
saattrupdan: Kept as `<repo>::<config>::<split>`, per our exchange above: the dataset ID is the key of the stored results record (`BenchmarkResult.dataset`, which carries no language field), so naming all 26 subsets of multilingual-gsm-symbolic `repo::test_original` would have had the Danish run overwrite the English one and the "already benchmarked, skipping" check treat them as the same result.

What did change is the other half of the duplication: `--language` now conflicts with `--dataset` exactly as `--task` does (72d361ef), so configuration selection lives in the dataset ID only.

src/euroeval/cli.py
saattrupdan: Done — inlined into the command body and the helper is gone. The command docstring gained a `Raises:` section because the docstring lint requires the raised exception to be documented at the raise site.

src/euroeval/languages.py
saattrupdan: Done as suggested: `Language` now has `code_3` and an optional `code_1`, with `code` a property preferring `code_1`, and the `ISO_639_3_ALIASES` table is deleted (187 lines of data, now stated once per language).

Two things the change surfaced:

- All 319 registry keys are identical to `main`, which is load-bearing — a language whose `code` became its 639-3 code would silently move its stored results — so `test_the_registry_is_keyed_by_the_preferred_code` asserts it, alongside a before/after key comparison.
- The four regional entries (`zh-cn`, `zh-tw`, `pt-pt`, `pt-br`) keep their tag in `code_1`, so `code` is unchanged for them. A 639-3 code shared by regional variants resolves to the untagged language when one is registered (`por` → Portuguese) and to nothing when it is not (`zho`), rather than to whichever Chinese was defined first. `zho` therefore still resolves to no language, which is why the `zho` config of the reference dataset is skipped.

src/frontend/md/python-package.md
saattrupdan: Removed.

src/frontend/md/python-package.md
saattrupdan: Moved into a new `### The \`math\` task` section, placed with the other task-level docs before `### Custom column names`; there was no math task section, so I created one. The `eval.yaml` section no longer discusses scoring.

tests/test_benchmark_config_factory.py
saattrupdan: Kept the double `::`, for the reason in the other thread: two expanded subsets sharing a name would write to the same results record, and results have no language field to separate them. This is the ID that gets recorded with a result and copied back into `--dataset`, so it has to name one task entry uniquely on its own.

src/euroeval/yaml_config.py
saattrupdan: Kept, for the same reason as the test thread — accepting the third part is what makes a recorded subset name round-trip back into `--dataset`.

Your comment did expose a real bug next door: the unknown-split error responded to a configuration name by recommending `--language 'da'`, which after this change is a command the CLI refuses. It now lists the subsets that name stands for:

```text
Unknown split 'dan' for dataset '...::dan'. Available splits are: ['test_original', 'test_synthetic']. 'dan' is a configuration, not a split; name one of its subsets instead: ...::dan::test_original, ...::dan::test_synthetic.
```

saattrupdan:
Addressed both comments in 72d361ef:

- **`promote_inspect_ai_prompt_template` aborted the scan** — a `prompt_template` solver whose `args` was not a mapping, or whose `template` was not a string, returned from the function instead of moving on, so a malformed first entry hid a valid template behind it. Both now `continue`. The other solver/scorer scans in the file already continued correctly. Covered by `test_a_malformed_solver_does_not_hide_the_later_one`.
- **The boxed-answer warning missed `\fbox`** — the check tested only for the substring `boxed`, so a prompt correctly instructing `\fbox{...}` still warned. The check now accepts the spellings `euroeval.metrics.math` extracts an answer from (`boxed`, which also covers `\beginboxed{}`, and `fbox`). Covered by `test_a_boxed_instruction_does_not_warn`.

No other comments were open on the review.

@copilot

PR #2169, opened 2026-09-05
fix: report the --dataset/--task conflict as a CLI usage error

saattrupdan:
## What changes

Combining `--dataset` with `--task` has always been invalid — `Benchmarker.__init__` raised `ValueError("Only one of \`task\` and \`dataset\` can be specified.")` — but the CLI let it escape as a traceback:

```
ValueError: Only one of `task` and `dataset` can be specified.
```

The CLI now validates the options up front and raises `click.UsageError`, so the run exits with code 2 and

```
Error: Only one of `--task` and `--dataset` can be specified.
```

The library keeps raising `ValueError`, so the API contract is unchanged.

## Note on `--language`

This PR originally also made `--language` conflict with `--dataset`. **That part is reversed**: `--language` remains combinable with `--dataset` and filters the selected datasets. That is the mechanism for choosing a subset of the datasets expanded from a multi-language external `eval.yaml` repository (e.g. `--dataset danish-foundation-models/multilingual-gsm-symbolic --language da`), so an external dataset repo can no longer be reachable only through a positional selector.

## Testing

- `pytest tests/test_benchmarker.py tests/test_cli.py tests/test_benchmark_config_factory.py` → 86 passed, 177 skipped (the skips are the pre-existing `pytest-dependends` skips in a targeted run).
- `pre-commit run` clean on the changed files.
- Manually: `--dataset dansk --task classification` exits 2 with no traceback; `--dataset dansk --language da` passes validation and proceeds; a language that the named dataset does not cover selects no datasets (covered by `test_prepare_dataset_configs_language_filters_explicit_dataset`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,10 @@ project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
   and require an explicit `thinking.type` of `disabled`.
 - The logical-reasoning task is now restricted to instruction-tuned and
   reasoning models, excluding base models.
+- Combining `--dataset` with `--task` now results in a CLI usage error rather than an
+  unhandled `ValueError` traceback. The two options are mutually exclusive, as
+  `--dataset` fully specifies which datasets to benchmark; `--language` can still be
+  used to narrow the selected datasets down further.
 
 ## [v18.0.0] - 2026-08-14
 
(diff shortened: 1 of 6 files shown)


saattrupdan:
Superseded: folded into #2167, which now carries the math scorer, the eval.yaml subset expansion and this CLI fix. The `--dataset`/`--language` half of this PR is deliberately not carried over - `--language` is how the expanded subsets are selected, so the two options are meant to combine.

PR #2173, opened 2026-09-09
fix: restore leading letters dropped by RobBERT QA offsets

### LINKED ISSUE #2171: `squad-nl` Exact Match collapses for RobBERT-family tokenizers (offset_mapping answer extraction)
RobBERT-2023-base on EuroEval's `squad-nl` vs the same model on [DUMB](https://github.com/wietsedv/dumb)'s SQuAD-NL (same data, HF's standard `run_qa` post-processing):

| Model | EuroEval `squad-nl` | DUMB SQuAD-NL |
|---|---|---|
| RobBERT-2023-base | F1 48 / EM 8 | F1 69.3 / EM 46.7 |

The EM basically drops to the floor on EuroEval while F1 is depressed too. The public Dutch leaderboard shows the same pattern for RobBERT-2023-large: ~56 F1 but only ~12 EM.

It comes from how answers are reconstructed in `find_valid_answers` (`euroeval/task_group_utils/question_answering.py`, ~L703 in 16.7.1):

```python
start_char = offset_mapping[start_index][0]
end_char = offset_mapping[end_index][1]
text = context[start_char:end_char]
```

The predicted answer is a raw slice of the context using the tokenizer's `offset_mapping`. RobBERT-2023's `tokenizer.json` has a non-standard `pre_tokenizer` (`Sequence[Whitespace, ByteLevel, Punctuation]` rather than plain `ByteLevel`), which pushes every word-initial token's start offset one character to the right. So the slice loses the first character of the answer and it never matches the gold string. F1 mostly survives (on a multi-word answer only the first token is wrong), but EM goes to ~0. Any model that inherits this tokenizer is hit the same way.

Decoding the predicted token span instead of slicing the context would sidestep it, e.g. `tokenizer.decode(input_ids[start_index:end_index+1])`. Happy to send a minimal repro or a PR.

jaideeppyne:
Fixes #2171.

`find_valid_answers` rebuilt the predicted answer by slicing the context with the tokenizer `offset_mapping`. RobBERT-family models use `Sequence[Whitespace, ByteLevel, Punctuation]` rather than plain ByteLevel, so a word-initial token's start offset is one character into the word. The slice then drops the first letter (`Amsterdam` → `msterdam`) and SQuAD Exact Match on `squad-nl` collapses (RobBERT-2023-base: EM 8 vs 46.7 on DUMB).

The reconstruction now fills only a non-whitespace hole that is not already covered by the previous in-context token. Standard BERT/RoBERTa spans and overflowing features that start mid-word are unchanged.

Local: `pytest tests/test_task_group_utils/test_question_answering.py -o addopts=` → 6 passed. `ruff check` on the two changed files is clean.

An LLM agent was used to draft this change. I reviewed the offset cases, wrote the tests, and am accountable for the result.

### DIFF
--- a/src/euroeval/task_group_utils/question_answering.py
+++ b/src/euroeval/task_group_utils/question_answering.py
@@ -418,9 +418,12 @@ def find_valid_answers(
             # If we got to this point then the answer is valid, so we store the
             # corresponding start- and end character indices in the original context,
             # and from these extract the answer
-            start_char = offset_mapping[start_index][0]
-            end_char = offset_mapping[end_index][1]
-            text = context[start_char:end_char]
+            text = _answer_text_from_offsets(
+                offset_mapping=offset_mapping,
+                start_index=start_index,
+                end_index=end_index,
+                context=context,
+            )
 
             # Compute the score of the answer, being the sum of the start and end
             # logits. Intuitively, this indicates how likely the answer is to be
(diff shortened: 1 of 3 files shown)


src/euroeval/task_group_utils/question_answering.py
```diff
@@ -435,6 +438,48 @@ def find_valid_answers(
     return valid_answers
 
 
+def _answer_text_from_offsets(
+    offset_mapping: c.Sequence[tuple[int, int]],
+    start_index: int,
+    end_index: int,
+    context: str,
+) -> str:
+    """Slice the context for a predicted token span, repairing word-initial offsets.
```
saattrupdan, line 447: Missing Args section in the docstring

saattrupdan (changes requested):
Thanks a bunch! Only a small change. Also, can you try evaluating the model with this PR and see if the results are more reasonable.

src/euroeval/task_group_utils/question_answering.py
jaideeppyne: Added the missing Google-style `Args` section in 7b684d43, covering all four parameters. I also added the required changelog entry for this source fix.

jaideeppyne:
I also ran a one-iteration smoke evaluation with this PR:

```console
uv run euroeval --model DTAI-KULeuven/robbert-2023-dutch-base --dataset squad-nl --num-iterations 1 --finetuning-batch-size 8 --device mps --evaluate-val-split --no-save-results --raise-errors --no-progress-bar --verbose
```

On the truncated SQuAD-nl validation split (FP32), it completed with **48.42 F1 / 34.71 EM** and no failed instances.

Compared with the **48 F1 / 8 EM** result reported in the PR/issue, F1 is essentially unchanged while EM improves by about 26.7 points. That is consistent with this fix: answers that previously differed only by a missing leading character now count as exact matches. This is a single-iteration smoke result rather than the default multi-iteration benchmark, so I would not treat it as a replacement leaderboard score.

saattrupdan (approved):
Thanks for the smoke test! LGTM

PR #2179, opened 2026-09-14
Fix first-label-token mapping for chat models with long system prompts

djstrong:
## Summary

Instruction-tuned models with a chat template currently build the first-label-token
map by scanning the **entire** templated conversation for
`label.startswith(tok)`. Long system prompts often contain short tokens such as
`p` / `n`, so labels like `positif` are mapped to `p` instead of the real answer
prefix (e.g. `pos` / `posit`). Structured logprob scoring then collapses to the
wrong choice set (e.g. `['n', 'p']`) and classification metrics look near-chance
even when free-form generation is fine.

This PR isolates label tokens by **diffing** chat templates with empty vs
label-filled assistant content, falls back to **encode(label)** when the diff is
empty/inconsistent, and picks the **first** matching prefix within that span.

## Changes

- `get_first_label_token_mapping` in `tokenisation_utils.py`:
  - **A:** common-prefix / common-suffix diff of chat-templated token ids
  - **B:** fallback to the existing no-chat-template encode path
  - first matching cleaned token within the isolated span
- Regression tests with a contaminated fake chat tokenizer (`p`/`n` in the system
  span; labels `positif` / `négatif` → `pos` / `n`, not `p` / `n`), plus encode
  fallback and first-prefix selection
- `CHANGELOG.md` entry under `[Unreleased]` → Fixed

## Repro (before)

With a Bielik-style tokenizer (long system prompt in the chat template) and
French sentiment labels `négatif` / `positif`:

| Path | Mapping |
|------|---------|
| Upstream | `négatif → n`, `positif → p` |
| This PR | `négatif → n`, `positif → pos` |

On AlloCiné / similar sequence-classification tasks, native EuroEval with the
buggy mapping can force structured choices `['n', 'p']` and yield MCC ≈ 0 with
all predictions collapsed to one class, while unconstrained generation (or HTTP
eval that falls back to text parsing) scores much higher on the same checkpoint.

## Impact (before vs after)

On a chat model with a long system prompt, **about 20 tasks** that use structured
logprob scoring previously returned a **0 score** (collapsed predictions / near-chance
MCC). After this fix, those tasks score as expected again.

Example — French sentiment classification (**AlloCiné**):

- **Before:** MCC ≈ 0 (wrong first-token mapping)
- **After:** Matthew's Correlation Coefficient **95.42%**

## Test plan

- [x] `pytest tests/test_tokenisation_utils.py::test_get_first_label_token_mapping_ignores_system_prompt_tokens`
- [x] `pytest tests/test_tokenisation_utils.py::test_label_token_ids_via_chat_diff_isolates_label_span`
- [x] `pytest tests/test_tokenisation_utils.py::test_get_first_label_token_mapping_falls_back_to_encode`
- [x] `pytest tests/test_tokenisation_utils.py::test_pick_matching_label_token_uses_first_prefix`
- [x] Re-run native EuroEval on a chat model with a long system prompt (e.g.
      AlloCiné): ~20 previously zero-scoring tasks now recover; AlloCiné MCC
      **95.42%**


tests/test_tokenisation_utils.py
```diff
@@ -1,15 +1,20 @@
...
 
 import pytest
 from transformers.models.auto.tokenization_auto import AutoTokenizer
 
 from euroeval.benchmark_modules.hf import load_hf_model_config, load_tokeniser
 from euroeval.data_models import BenchmarkConfig, HashableDict
+from euroeval.enums import GenerativeType
 from euroeval.tokenisation_utils import (
+    _label_token_ids_via_chat_diff,
+    _pick_matching_label_token,
```
saattrupdan: Either these functions should be public (no preceding underscore) or we shouldn't have any tests for them.

src/euroeval/tokenisation_utils.py
```diff
@@ -423,6 +412,159 @@ def get_first_label_token_mapping(
...
+            The tokeniser.
+        enable_thinking:
+            Whether to enable thinking tokens in the chat template.
+
+    Returns:
+        The token ids belonging to ``label``, or None if the diff is empty/inconsistent
+        (caller should fall back to encoding the label alone).
+    """
+    try:
+        ids_with = _normalize_token_ids(
```
saattrupdan: Could we make these names describe the two template variants? For example, `label_template_token_ids` and `empty_template_token_ids` are clearer than `ids_with` and `ids_empty`. Also, please pass the argument to `_normalize_token_ids` by keyword (`token_ids=...`) in line with the project convention; the nested positional call is easier to break silently if the signature changes.

src/euroeval/tokenisation_utils.py
```diff
@@ -423,6 +412,159 @@ def get_first_label_token_mapping(
...
+            apply_chat_template(
+                conversation=_label_conversation(label),
```
saattrupdan: Please call this helper with `assistant_content=label` (and use `assistant_content=""` for the corresponding call below). Keyword arguments make the value's role explicit and avoid silent breakage if the helper signature changes.

src/euroeval/tokenisation_utils.py
```diff
@@ -423,6 +412,159 @@ def get_first_label_token_mapping(
...
+        log_once(
+            f"Chat-template diff failed for label {label!r}: {exc!s}.",
+            level=logging.DEBUG,
+        )
+        return None
+
+    if not ids_with:
+        return None
+
+    prefix = _common_prefix_len(ids_with, ids_empty)
```
saattrupdan: `prefix`, `suffix`, and `end` hold lengths or indices, but their names can be read as token sequences. Could these be `common_prefix_length`, `common_suffix_length`, and `label_span_end`? Please also pass the token sequences to `_common_prefix_len` and `_common_suffix_len` by keyword.

src/euroeval/tokenisation_utils.py
```diff
@@ -423,6 +412,159 @@ def get_first_label_token_mapping(
...
+    suffix = _common_suffix_len(ids_with, ids_empty, start_a=prefix, start_b=prefix)
+    end = len(ids_with) - suffix
+    if end <= prefix:
+        return None
+    return ids_with[prefix:end]
+
+
+def _common_prefix_len(a: c.Sequence[int], b: c.Sequence[int]) -> int:
```
saattrupdan: `a`, `b`, `n`, and `i` make this boundary algorithm unnecessarily difficult to verify. Please use descriptive names such as `first_token_ids`, `second_token_ids`, `maximum_prefix_length`, and `common_prefix_length`. The docstring should also include Google-style `Args` and `Returns` sections.

src/euroeval/tokenisation_utils.py
```diff
@@ -423,6 +412,159 @@ def get_first_label_token_mapping(
...
+    """Return the length of the common prefix of two token-id sequences."""
+    n = min(len(a), len(b))
+    i = 0
+    while i < n and a[i] == b[i]:
+        i += 1
+    return i
+
+
+def _common_suffix_len(
+    a: c.Sequence[int], b: c.Sequence[int], *, start_a: int, start_b: int
```
saattrupdan: The names `a`, `b`, `start_a`, `start_b`, and `i` obscure what is being compared and counted here. Please use descriptive sequence, start-index, and suffix-length names. This docstring should also document every argument and the return value using the project's Google style.

src/euroeval/tokenisation_utils.py
```diff
@@ -423,6 +412,159 @@ def get_first_label_token_mapping(
...
+            Token ids as a list, nested list, tensor, or BatchEncoding-like object.
+
+    Returns:
+        A flat list of integer token ids.
+    """
+    if isinstance(token_ids, BatchEncoding):
+        token_ids = token_ids.input_ids
+    elif hasattr(token_ids, "input_ids"):
+        token_ids = token_ids.input_ids
+    tolist = getattr(token_ids, "tolist", None)
```
saattrupdan: `tolist` contains a method rather than token IDs, so its role is not immediately clear. Could this be named `to_list_method` or `tolist_method`?

src/euroeval/tokenisation_utils.py
```diff
@@ -503,6 +645,26 @@ def should_prompts_be_stripped(
...
+    Args:
+        token_list:
+            Cleaned tokens from the isolated label span (or the encoded label).
+        label:
+            The label string to match.
+
+    Returns:
+        The first matching token, or None if no token is a prefix of ``label``.
+    """
+    label_l = label.lower()
```
saattrupdan: Could `label_l`, `tok`, and `cleaned` be renamed to `lowercase_label`, `token`, and `cleaned_token`? These names make the matching conditions much easier to read.

tests/test_tokenisation_utils.py
```diff
@@ -50,6 +55,178 @@ def test_get_end_of_chat_token_ids(
...
+            "<gen>",
+            "pos",
+            "itif",
+            "égatif",
+        ):
+            self._add(tok)
+        if extra_span_token is not None:
+            self._add(extra_span_token)
+
+    def _add(self, tok: str) -> int:
```
saattrupdan: Several names in this test double (`tok`, `idx`, `_tok2id`, `_id2tok`, `toks`, `msg`, `ids`, and `i`) are overly abbreviated; please use descriptive token and message names. The newly added `_add`, `__call__`, `encode`, `apply_chat_template`, and `convert_ids_to_tokens` methods are also undocumented, and calls to `_add` should use a keyword argument in line with the project conventions.

tests/test_tokenisation_utils.py
```diff
@@ -50,6 +55,178 @@ def test_get_end_of_chat_token_ids(
...
+        return ids if tokenize else " ".join(toks)
+
+    def convert_ids_to_tokens(self, ids: list[int], **kwargs: object) -> list[str]:
+        return [self._id2tok[int(i)] for i in ids]
+
+
+def _mapping_for_chat_tokeniser(
+    tokeniser: _ContaminatedChatTokeniser, monkeypatch: pytest.MonkeyPatch
+) -> dict[str, str] | bool:
+    """Return first-label-token mapping for a fake chat tokeniser."""
```
saattrupdan: This helper takes two arguments and returns a union, so the one-line summary is not sufficient under the project's Google-style docstring convention. Please add `Args` and `Returns` sections.

tests/test_tokenisation_utils.py
```diff
@@ -116,6 +293,13 @@ def test_load_xlmr_tokeniser_with_fallback(
     assert tokeniser.eos_token == "</s>"
 
 
+def test_pick_matching_label_token_uses_first_prefix() -> None:
+    """Use the first matching prefix, not a later longer one."""
+    assert _pick_matching_label_token(["a", "aa"], "aaa") == "a"
```
saattrupdan: If these direct private-helper tests remain, please call `_pick_matching_label_token` with `token_list=` and `label=` rather than positional arguments. This avoids silent mistakes if the signature changes. This comment becomes moot if the tests are removed in response to the earlier comment.

src/euroeval/tokenisation_utils.py
```diff
@@ -423,6 +412,159 @@ def get_first_label_token_mapping(
...
+        tokeniser:
+            The tokeniser.
+
+    Returns:
+        The token ids of ``label``.
+    """
+    add_prefix_space = should_prefix_space_be_added_to_labels(
+        labels_to_be_generated=local_labels, tokeniser=tokeniser
+    )
+    return _normalize_token_ids(
```
saattrupdan, line 604: Please pass this as `_normalize_token_ids(token_ids=...)` too, and apply the same change to the other occurrence above, for consistency with the project's keyword-argument convention.

saattrupdan (dismissed):

saattrupdan (approved):
Thank you, @djstrong, for the fix and the thorough regression coverage. The chat-template isolation and fallback behaviour now look solid, and all review comments have been addressed. Much appreciated!

Issue #2183, opened 2026-09-15
[MODEL EVALUATION REQUEST] deepseek-ai/DeepSeek-V4.1-Flash non-thinking


mathiasesn:
### Model ID

deepseek-ai/DeepSeek-V4.1-Flash

### Evaluation languages

- [x] Baltic languages (Latvian, Lithuanian)
- [x] Finnic languages (Estonian, Finnish)
- [x] Romance languages (Catalan, French, Italian, Portuguese, Romanian, Spanish)
- [x] Scandinavian languages (Danish, Faroese, Icelandic, Norwegian, Swedish)
- [x] Slavic languages (Belarusian, Bulgarian, Bosnian, Croatian, Czech, Polish, Serbian, Slovak, Slovenian, Ukrainian)
- [x] West Germanic languages (Dutch, English, German, Luxembourgish)
- [x] Albanian
- [x] Greek
- [x] Hungarian

mathiasesn:
@saattrupdan, any idea what this would cost to run through the [DeepSeek API](https://api-docs.deepseek.com/quick_start/pricing/)?

mathiasesn:
Depends on https://github.com/EuroEval/EuroEval/pull/2185 for evaluating with the DeepSeek API.

mathiasesn:
@saattrupdan, will run this over the weekend using the following command:

```shell
export DEEPSEEK_API_KEY=<your-key>
uv run euroeval --model "deepseek/deepseek-flash#no-thinking"
```

mathiasesn:
@saattrupdan, Here are the results

[euroeval_benchmark_results.zip](https://github.com/user-attachments/files/32502298/euroeval_benchmark_results.zip)

PR #2185, opened 2026-09-15
feat: add DeepSeek-V4.1-Flash support via DeepSeek API

mathiasesn:
Adds support for evaluating [DeepSeek-V4.1-Flash](https://huggingface.co/deepseek-ai/DeepSeek-V4.1-Flash) through the DeepSeek API as `deepseek/deepseek-flash`, in preparation for #2183.

## Changes

- Model metadata for `(deepseek/)?deepseek-flash`: vocab size 129,280, context length 1,048,576, 552B parameters, release date 2026-09-10.
- Added to `REASONING_MODELS`, since the DeepSeek API has thinking enabled by default.
- New `allowed_params` entry: `#no-thinking`, `#thinking`, `#low`, `#high`, `#max` (and `max` added to the `reasoning_effort` set, per the [DeepSeek thinking-mode docs](https://api-docs.deepseek.com/guides/thinking_mode)).
- `#no-thinking` now sends `thinking={"type": "disabled"}` for DeepSeek models. LiteLLM's DeepSeek config only honours `thinking.type` and silently drops `budget_tokens`, so the existing `budget_tokens=0` payload would have left thinking on. Other providers are unchanged.

## Testing

- `uvx pre-commit run` on the changed files passes.
- `pytest tests/test_benchmark_modules/test_litellm.py` passes (36 tests).


mathiasesn:
Tested live against the DeepSeek API on `scala-da` (validation split, 1 iteration):

| Model | MCC | Macro-F1 |
|---|---|---|
| `deepseek/deepseek-flash#no-thinking` (instruction-tuned path, logprobs) | 67.07% | 83.53% |
| `deepseek/deepseek-flash#low` (reasoning path, `reasoning_effort` via `extra_body`) | 65.00% | 82.40% |

This surfaced two issues fixed in 9e96b2a8:
- DeepSeek rejects `json_schema` response formats with *"This response_format type is unavailable now"* — added to the existing JSON-schema → `json_object` fallback.
- The generation retry loop rebuilt `generation_kwargs` on every attempt, discarding the error handlers' adjustments (e.g. a dropped `response_format`), so the same request failed 10× in a row. Kwargs are now computed once and the handler-adjusted version is persisted. This affects any provider whose errors only appear at generation time, not just DeepSeek.

src/euroeval/benchmark_modules/litellm.py
```diff
@@ -519,6 +551,11 @@ def generate(self, inputs: dict) -> GenerativeModelOutput:
                     error=error, **generation_kwargs
                 )
                 time_to_wait = max(time_to_wait, wait_time)
+
+            # Persist the adjusted kwargs, so that the next attempt (and subsequent
+            # batches) benefit from the fixes rather than re-triggering the error
+            self.generation_kwargs = generation_kwargs
```
saattrupdan: Persisting the complete, materialised kwargs here leaks dataset-specific settings across datasets. `Benchmarker` reuses the same model instance and `update_dataset_config()` only replaces `dataset_config`, so after any failure (including a transient 429), a later dataset can inherit the previous dataset's token limit, response schema, and logprob settings. Could we keep adjusted kwargs local to the retry loop and persist only provider-capability overrides, or otherwise rebuild dataset-specific kwargs when the dataset changes?

src/euroeval/benchmark_modules/litellm.py
```diff
@@ -101,6 +101,9 @@
     r"(gemini/)?gemini-[1-9](\.[0-9])?-(flash|pro).*": 256_128,
     # xAI models
     r"(xai/)?grok.*": -1,
+    # DeepSeek models
+    # Source: HF `deepseek-ai/DeepSeek-V4.1-Flash` `config.json` (vocab size).
+    r"(deepseek/)?deepseek-flash.*": 129_280,
```
saattrupdan: Shouldn't this require the `deepseek/` provider prefix? Without it, an unprefixed or custom-API deployment named `deepseek-flash`—potentially serving the open-weight model—will inherit metadata and behaviour intended specifically for the DeepSeek API alias. Please apply the required prefix consistently to the other metadata mappings, `REASONING_MODELS`, and `allowed_params` as well.

src/euroeval/benchmark_modules/litellm.py
```diff
@@ -364,6 +385,18 @@ def __init__(
         self.is_ollama = model_config.model_id.startswith(
             "ollama/"
         ) or model_config.model_id.startswith("ollama_chat/")
+
+        # Detect whether the model is served by the DeepSeek API, as LiteLLM's
+        # DeepSeek transformation discards `budget_tokens` and the `reasoning_effort`
+        # level, so we shape the thinking parameters differently for these models
+        self.is_deepseek = (
+            re.fullmatch(
+                pattern=r"(deepseek/)?deepseek-.*",
```
saattrupdan: Could we avoid introducing a broad `is_deepseek` flag and use the existing error-handling path to adapt unsupported thinking arguments, as we do for other providers? If LiteLLM silently removes `budget_tokens` before DeepSeek can reject it, some provider-specific shaping may be unavoidable; in that case, please scope it strictly to the `deepseek/` provider prefix and add a test demonstrating the silent-drop behaviour.

saattrupdan (changes requested):
The overall DeepSeek API support looks sound, and the live testing caught useful compatibility issues. I’m requesting changes for one correctness issue: persisting fully materialised generation kwargs can leak dataset-specific settings when the model instance is reused across datasets.

I’ve also left comments about requiring the `deepseek/` provider prefix to distinguish the official API from open-weight deployments, and about limiting provider-specific branching around thinking parameters.

Please add regression tests covering retry adjustments, dataset changes after a retry, DeepSeek parameter mappings, and the response-format fallback.

src/euroeval/benchmark_modules/litellm.py
mathiasesn: Good catch, thanks — fixed in 2da15018 and 7841c5b1. `self.generation_kwargs` is now only ever set from `__init__` (the user override). Instead of persisting the materialised dict, `generate()` rebuilds kwargs from `get_generation_kwargs(self.dataset_config)` on every call and replays only the errors that `_handle_parameter_error` resolved (deduplicated by message, and never service/rate-limit errors), so model-level fixes like the `json_object` fallback carry over while `max_completion_tokens`/`response_format` always come from the current dataset. Covered by `TestRetryAdjustments::test_dataset_change_after_retry_uses_new_dataset_kwargs`, plus `test_service_error_is_not_persisted` and `test_duplicate_errors_are_stored_once`.

src/euroeval/benchmark_modules/litellm.py
mathiasesn: Agreed — done in 2da15018. The prefix is now mandatory (`deepseek/deepseek-flash.*`) across all four metadata mappings, `REASONING_MODELS`, `allowed_params`, and the `is_deepseek` check itself (`deepseek/.*`). `TestDeepSeekParams::test_prefix_required` asserts that a bare `deepseek-flash` matches none of them and has `is_deepseek is False`. CHANGELOG notes the requirement.

src/euroeval/benchmark_modules/litellm.py
mathiasesn: Unfortunately the error-handling path cannot help here: LiteLLM's `DeepSeekChatConfig.map_openai_params` strips `budget_tokens` and collapses `reasoning_effort` to a boolean `thinking.type` *before* the request is sent, so DeepSeek never sees anything to reject and the fallback never fires. I have added `TestLiteLLMDeepSeekTransformation` (7d621499) which pins that upstream behaviour: `{"thinking": {"type": "enabled", "budget_tokens": 1234}}` → `{"thinking": {"type": "enabled"}}` and `{"reasoning_effort": "high"}` → `{"thinking": {"type": "enabled"}}`. So a future LiteLLM fix will show up as a test failure and we can drop the shaping then.

The shaping is now scoped strictly to the `deepseek/` prefix and reduced to a single post-processing block in `_setup_model_params`: the generic kwargs are built first, then one `if self.is_deepseek:` rewrites `thinking` to `{type}` only and moves `reasoning_effort` into `extra_body`. `TestDeepSeekParams` covers all five params plus a non-DeepSeek control.

mathiasesn:
Thanks for the review — all points addressed and pushed (2da15018 … 7d621499):

- **Kwargs leak across datasets** — `self.generation_kwargs` is no longer mutated; fixes learned by the parameter handlers are replayed on freshly built, dataset-specific kwargs each call. Service errors are never persisted and stored errors are deduplicated.
- **`deepseek/` prefix** — now required everywhere (metadata mappings, `REASONING_MODELS`, `allowed_params`, `is_deepseek`).
- **Provider-specific branching** — collapsed to one `if self.is_deepseek:` post-processing block, with a test pinning LiteLLM's silent-drop behaviour that makes it necessary.
- **Regression tests** — retry adjustments, dataset change after a retry, DeepSeek parameter mappings (+ prefix requirement), response-format fallback, service-error non-persistence, and error deduplication. `tests/test_benchmark_modules/test_litellm.py` now has 51 tests; full unit suite passes locally (1163 passed).

src/euroeval/benchmark_modules/litellm.py
mathiasesn: Follow-up: on reflection the `is_deepseek` attribute had no reason to exist as a flag, so I removed it in eb0364b3. The single shaping block in `_setup_model_params` now just does `self.model_config.model_id.lower().startswith("deepseek/")` inline, the same way Ordbogen models are handled in `get_generation_kwargs`. Nothing else in the class references DeepSeek, and `test_prefix_required` now asserts on the shaped kwargs (generic `budget_tokens` for bare `deepseek-flash`, `{"type": "enabled"}` for `deepseek/deepseek-flash`) rather than on an attribute.

src/euroeval/benchmark_modules/litellm.py
```diff
@@ -378,6 +399,7 @@ def __init__(
         )
 
         self.generation_kwargs = generation_kwargs
+        self._handled_param_errors: dict[str, Exception] = {}
```
saattrupdan: Could this store normalised capabilities rather than exception objects? For example, `set[ParameterAdjustment]` where `ParameterAdjustment` is a `StrEnum` containing `NO_JSON_SCHEMA`, `USE_MAX_TOKENS`, `NO_TEMPERATURE`, etc. Exceptions encode provider wording and the context of an old request; the durable information we actually want is what this model/provider supports. Value-bearing knowledge such as a maximum thinking budget can live in a separate typed field.

src/euroeval/benchmark_modules/litellm.py
```diff
@@ -449,14 +471,32 @@ def generate(self, inputs: dict) -> GenerativeModelOutput:
...
+        )
+
+        # Replay previously handled parameter errors, so that model-level quirks
+        # discovered in earlier calls (e.g. "no json_schema") are re-applied to the
+        # freshly built kwargs, without leaking dataset-specific kwargs (like
+        # `max_completion_tokens`) across datasets via `self.generation_kwargs`. We
+        # call `_handle_parameter_error` directly (rather than `_handle_exception`),
+        # since it never raises and simply returns `None` if the error no longer
+        # applies to the current kwargs, so replay can never fail.
+        for handled_error in self._handled_param_errors.values():
```
saattrupdan: Replaying old exceptions assumes their handlers apply to the new dataset's kwargs. That is not generally true: the JSON-schema handler, for example, unconditionally inserts `response_format={"type": "json_object"}` even if the fresh kwargs contain no response format. Could this instead call an `_apply_parameter_adjustments()` helper that conditionally applies learned capabilities—for example, only replacing an existing response format whose type is `json_schema`?

src/euroeval/benchmark_modules/litellm.py
```diff
@@ -515,10 +555,29 @@ def generate(self, inputs: dict) -> GenerativeModelOutput:
...
-                generation_kwargs, wait_time = self._handle_exception(
-                    error=error, **generation_kwargs
+                # Try the parameter-error handler first, so that we can tell whether
+                # the error was resolved by a model-level parameter fix (which should
+                # be remembered and replayed on future calls) as opposed to a
+                # service/rate-limit error (which must not be persisted). If no
+                # parameter handler matches, fall back to the full exception handler,
+                # which will re-run the (now no-op) parameter handler and then try
+                # the service/fatal handlers.
+                param_result = self._handle_parameter_error(
```
saattrupdan: Could `_handle_exception` remain the sole entry point? It already calls `_handle_parameter_error` first and therefore knows whether a parameter handler matched. It could record the resulting persistent `ParameterAdjustment` before returning, while service/fatal handling remains non-persistent. That would remove the duplicated dispatch here and keep classification and persistence in one place.

src/euroeval/benchmark_modules/litellm.py
```diff
@@ -515,10 +555,29 @@ def generate(self, inputs: dict) -> GenerativeModelOutput:
...
+                    error=error,
+                    error_msg=str(error).lower(),
+                    model_id=self.model_config.model_id,
+                    generation_kwargs=generation_kwargs,
                 )
+                if param_result is not None:
+                    generation_kwargs, wait_time = param_result
+                    self._handled_param_errors.setdefault(str(error), error)
```
saattrupdan: This currently persists every matching parameter error. That includes `"you've reached the maximum number of requests with logprobs"`, which is a transient quota/service condition rather than evidence that the model lacks logprobs support. One such response would disable logprobs for subsequent datasets. Please persist only explicit unsupported/disabled capability errors and keep temporary quota handling request-local.

saattrupdan (changes requested):
The original review comments are addressed: the `deepseek/` prefix is now required, the broad `is_deepseek` state is gone, and fully materialised dataset kwargs are no longer assigned to `self.generation_kwargs`. The remaining concern is how provider capabilities are persisted across `generate()` calls.

`_handle_parameter_error` already sits behind `_handle_exception` and was designed to handle an error raised by the current request. Storing the original exception and replaying it against future datasets assumes that the same handler remains applicable to their newly generated kwargs. That is not always true: for example, replaying an old JSON-schema rejection adds `response_format={"type": "json_object"}` even when a later dataset did not request structured output. The current dataset-change test keeps the same schema and therefore does not cover this boundary.

I suggest storing normalised, model-level capability adjustments instead of exception objects, for example:

```python
class ParameterAdjustment(enum.StrEnum):
    NO_JSON_SCHEMA = "no-json-schema"
    USE_MAX_TOKENS = "use-max-tokens"
    NO_STOP_SEQUENCES = "no-stop-sequences"
    NO_TEMPERATURE = "no-temperature"
    TEMPERATURE_MUST_BE_ONE = "temperature-must-be-one"
    NO_LOGPROBS = "no-logprobs"
    NO_TOP_LOGPROBS = "no-top-logprobs"
    THINKING_DISABLED_REQUIRES_TYPE = "thinking-disabled-requires-type"
```

The model can retain `set[ParameterAdjustment]` across datasets, while an `_apply_parameter_adjustments()` helper applies each capability conditionally to freshly generated kwargs. For example, `NO_JSON_SCHEMA` should only replace an existing `json_schema` response format; it should do nothing when `response_format` is absent. `USE_MAX_TOKENS` should only move an existing `max_completion_tokens`, preserving the current dataset's token limit. Value-bearing knowledge such as a maximum thinking budget can use a separate typed field rather than an enum member.

`_handle_exception` can remain the single entry point: it already knows whether `_handle_parameter_error` matched, so it can record the returned persistent adjustment before returning. Service and fatal errors should never enter the capability set. In particular, `"you've reached the maximum number of requests with logprobs"` is transient and must not permanently disable logprobs for later datasets; only explicit unsupported/disabled messages should record `NO_LOGPROBS`.

Please add regressions for a schema rejection followed by an unstructured dataset, preserving a new dataset's token limit when applying `USE_MAX_TOKENS`, and confirming that the temporary logprobs quota message is not persisted.

src/euroeval/benchmark_modules/litellm.py
mathiasesn: Done in e5ad69f4. The exception dict is replaced by `self._parameter_adjustments: set[ParameterAdjustment]` (new enum in `euroeval.enums`, 12 members), plus a separate typed `self._max_thinking_budget: int | None` for the value-bearing thinking-budget cap.

src/euroeval/benchmark_modules/litellm.py
mathiasesn: Fixed in e5ad69f4. `generate()` now builds the kwargs fresh and passes them through `_apply_parameter_adjustments()`, where every capability is applied conditionally: `NO_JSON_SCHEMA` only rewrites an existing non-`json_object` `response_format` (a no-op when the dataset requested none), `USE_MAX_TOKENS` only moves an existing `max_completion_tokens` so the current dataset's limit is preserved, and the thinking-budget cap only lowers an `enabled` budget that exceeds the known maximum. Regressions added: `test_schema_rejection_does_not_add_response_format_to_unstructured_dataset` and `test_use_max_tokens_preserves_new_dataset_token_limit`.

src/euroeval/benchmark_modules/litellm.py
mathiasesn: Agreed — `_handle_exception` is the single entry point again (e5ad69f4). `_handle_parameter_error` now returns `(kwargs, wait_time, adjustment | None)`; `_handle_exception` records the adjustment and returns the usual 2-tuple, so the retry loop is back to a plain `_handle_exception` call. Service and fatal errors never touch the set.

src/euroeval/benchmark_modules/litellm.py
mathiasesn: Fixed in e5ad69f4. The quota message ("you've reached the maximum number of requests with logprobs") still disables logprobs for the current request but returns adjustment `None`, so it is not persisted; only the explicit unsupported/disabled messages record `NO_LOGPROBS`. I applied the same reasoning to two other non-model-level cases: the `maxItems` fix (dataset-specific schema) and "the model returned empty outputs". Regression: `test_logprobs_quota_message_is_not_persisted` (with a persisted counterpart for the explicit message).

mathiasesn:
Addressed the second round in e5ad69f4 — stored exceptions are replaced by a `ParameterAdjustment` capability set applied conditionally to fresh kwargs, `_handle_exception` is the sole entry point, transient messages (logprobs quota, empty outputs, `maxItems`) are no longer persisted, and the three requested regressions are added. Full suite: 1518 passed / 172 skipped.

src/euroeval/benchmark_modules/litellm.py
```diff
@@ -544,6 +576,108 @@ def generate(self, inputs: dict) -> GenerativeModelOutput:
...
+                The freshly built generation kwargs for the current dataset.
+
+        Returns:
+            The generation kwargs with all persisted adjustments applied.
+        """
+        adjustments = self._parameter_adjustments
+        if ParameterAdjustment.NO_STOP_SEQUENCES in adjustments:
+            if "stop" in generation_kwargs:
+                generation_kwargs["stop"] = None
+        if ParameterAdjustment.NO_LOGPROBS in adjustments:
```
saattrupdan, line 598: `NO_LOGPROBS` is a model-wide capability, but `_disable_logprobs()` also removes `response_format`. If this adjustment is learned on a classification dataset, replaying it on a later structured-output dataset strips that dataset's unrelated schema. Could the persisted adjustment disable only `logprobs`/`top_logprobs` (and the label-token mapping), while any response-format removal remains request-local or is conditioned on the format having been introduced specifically as the logprobs fallback? A logprobs-dataset -> structured-dataset regression would cover this.

src/euroeval/benchmark_modules/litellm.py
```diff
@@ -449,14 +473,21 @@ def generate(self, inputs: dict) -> GenerativeModelOutput:
         inputs_to_run: c.Sequence[
             tuple[int, c.Sequence[litellm.AllMessageValues] | str]
         ] = list(enumerate(model_inputs))
+        # Build the kwargs afresh for the current dataset and re-apply the
+        # model-level adjustments learned from earlier errors, so that nothing
+        # dataset-specific (like `max_completion_tokens`) leaks across datasets via
+        # `self.generation_kwargs`
+        generation_kwargs = self._apply_parameter_adjustments(
```
saattrupdan: This applies persisted adjustments only after `get_generation_kwargs()` returns, but that method performs its own API probe before returning. Every later dataset/batch therefore first sends already-known-invalid parameters and rediscovers the same error; the persisted maximum thinking budget has the same issue. Could `_apply_parameter_adjustments()` also run inside `get_generation_kwargs()` after the fresh kwargs are assembled and before its test loop? The outer application can remain for constructor-supplied kwargs.

src/euroeval/benchmark_modules/litellm.py
```diff
@@ -1117,7 +1269,7 @@ def _handle_parameter_error(
                 level=logging.DEBUG,
             )
             generation_kwargs["response_format"] = dict(type="json_object")
-            return generation_kwargs, 0
+            return generation_kwargs, 0, ParameterAdjustment.NO_JSON_SCHEMA
```
saattrupdan: Not every message in this branch establishes a model-wide lack of JSON-schema support. In particular, `Property keys should match pattern` is a validation failure for the current schema's keys; a later dataset may have a valid schema. Could that case apply the `json_object` fallback only to the current request and return adjustment `None`, while reserving `NO_JSON_SCHEMA` for explicit unsupported-type messages? Please add a cross-dataset regression with an invalid schema followed by a valid one.

src/euroeval/benchmark_modules/litellm.py
```diff
@@ -1169,23 +1326,33 @@ def _handle_parameter_error(
...
                 f"The model {model_id!r} does not support the `response_format` "
                 "parameter, so disabling it.",
                 level=logging.DEBUG,
             )
             generation_kwargs.pop("response_format", None)
-            return generation_kwargs, 0
+            adjustment = (
+                None
+                if empty_outputs_message in error_msg
+                else ParameterAdjustment.NO_RESPONSE_FORMAT
```
saattrupdan, line 1376: Only the unexpected-keyword message here demonstrates that `response_format` itself is unsupported. `'maxitems' is not supported` concerns one schema keyword, and `must contain the word 'json'` is prompt/request-specific; persisting `NO_RESPONSE_FORMAT` for either can disable valid structured output on later datasets. Could those cases remain request-local (`adjustment=None`) and persist `NO_RESPONSE_FORMAT` only for an explicit unsupported-parameter error?

saattrupdan (changes requested):
Thanks for working through the earlier review rounds. This is my final review pass, and I want to make the merge threshold explicit so this does not turn into an open-ended sequence of refactoring requests.

There are two remaining correctness fixes:

1. A persisted `NO_LOGPROBS` adjustment must not remove an unrelated structured-output `response_format` from a later dataset. Please keep the persistent capability limited to disabling logprobs, with response-format removal remaining request-local or conditional on it being the logprobs fallback.
2. Request/schema-specific validation failures must not be persisted as model-wide capabilities. In particular, `Property keys should match pattern`, `'maxitems' is not supported`, and `must contain the word 'json'` should remain local to the current request; only explicit unsupported-parameter/type messages should persist `NO_JSON_SCHEMA` or `NO_RESPONSE_FORMAT`.

The provider-resolution abstraction and applying known adjustments before the internal probe can be handled as follow-up maintainability work; they are not blockers for this PR.

Once the two correctness issues above are fixed and covered by the suggested cross-dataset regressions, I consider the PR good to merge and will not request another design or refactoring round here.

src/euroeval/benchmark_modules/litellm.py
mathiasesn: Fixed in 80216a71. `_disable_logprobs()` now only drops `logprobs`/`top_logprobs` (and sets `first_label_token_mapping = False`), so the persisted `NO_LOGPROBS` capability never touches `response_format`. The `response_format` pop stays request-local inside the logprobs branch of `_handle_parameter_error`, matching upstream's original fallback. Cross-dataset regression: `test_logprobs_rejection_leaves_schema_response_format_for_new_dataset` — a logprobs rejection on dataset A, then a second `generate()` on the same instance with a `json_schema` response format, which is sent unchanged with logprobs absent.

src/euroeval/benchmark_modules/litellm.py
mathiasesn: Fixed in 80216a71. `"Property keys should match pattern"` is split out as `malformed_schema_message`: it still falls back to `json_object` for the current request but returns adjustment `None`, so `NO_JSON_SCHEMA` is only persisted for `'json_schema' is not supported` and `This response_format type is unavailable now`. Covered by `test_malformed_schema_message_is_not_persisted`, `test_json_schema_unsupported_message_is_persisted` and the cross-dataset `test_malformed_schema_does_not_persist_and_new_dataset_schema_is_unchanged`.

src/euroeval/benchmark_modules/litellm.py
mathiasesn: Fixed in 80216a71. Only `got an unexpected keyword argument 'response_format'` persists `NO_RESPONSE_FORMAT`. `'maxitems' is not supported`, `must contain the word 'json'` and the existing empty-outputs message are now in a `request_local_response_format_messages` list that drops `response_format` for the current request only (adjustment `None`). Covered by the parametrised `test_request_local_response_format_messages_are_not_persisted` and `test_unexpected_keyword_response_format_message_is_persisted`.

mathiasesn:
Both correctness fixes from the final pass are in 80216a71, each with the cross-dataset regressions (see the inline replies). The provider-resolution abstraction and applying adjustments before the internal probe are left for follow-up work as agreed. Full suite: 1531 passed, 172 skipped.

mathiasesn:
Follow-up for the deferred probe-ordering item is up in #2187 (stacked on this branch, so its diff will shrink to a single commit once this PR merges). The provider-resolution abstraction I have left for a separate discussion.

saattrupdan (approved):
Both agreed correctness blockers are fixed and covered by cross-dataset regressions. Focused tests and checks pass. LGTM.

PR #2187, opened 2026-09-17
perf: apply known parameter adjustments before the internal probe request

mathiasesn:
Follow-up to #2185, addressing the maintainability item deferred in the final review there ("applying known adjustments before the internal probe").

> **Stacked on #2185.** This branch is based on `feat/deepseek-flash-thinking-params` because it depends on `_parameter_adjustments`. Until #2185 merges the diff shown here includes its commits; the change specific to this PR is the single commit `edcc18f8`.

## Problem

`get_generation_kwargs()` sends a one-off test request on every dataset switch. The persisted `ParameterAdjustment` set was only applied in `generate()`, *after* `get_generation_kwargs()` returned, so on each new dataset the probe re-sent parameters the model was already known to reject (e.g. `logprobs`, a `json_schema` response format) and only converged through the retry loop — one or more guaranteed-failing requests per dataset.

## Change

- `get_generation_kwargs()` now calls `_apply_parameter_adjustments()` right before the probe loop.
- The call in `generate()` is kept: it is idempotent and still covers the `self.generation_kwargs` user-override path, and it re-filters any keys the post-probe `reasoning_content` detection re-adds (e.g. `max_completion_tokens` vs. a persisted `USE_MAX_TOKENS`).
- Tests: `test_get_generation_kwargs_applies_adjustments_before_probe` runs the real `get_generation_kwargs` with only `Router.acompletion` patched and asserts the probe request has no `logprobs`/`top_logprobs` when `NO_LOGPROBS` is persisted; `test_get_generation_kwargs_keeps_logprobs_without_adjustments` is the inverse guard.
- CHANGELOG entry under `### Changed`.

## Verification

`uv run pytest tests --no-cov`: 1533 passed, 172 skipped. Pre-commit hooks green.


mathiasesn:
Rebased-by-merge onto `main` now that #2185 is in, so the diff is down to this PR's own changes.

Also pushed `7d13a31f`: the post-probe `reasoning_content` detection re-adds `max_completion_tokens`, which conflicted with a persisted `USE_MAX_TOKENS` and left `get_generation_kwargs()` returning both `max_tokens` and `max_completion_tokens` (only fixed later by the second pass in `generate()`). Adjustments are now re-applied after the probe loop as well, with a regression test that fails without it. PR description updated.

mathiasesn:
Pushed `e0225e86`, test/doc-only: `_make_response()` now defaults `reasoning_content` to `None` (it was a truthy `MagicMock`, so the two non-reasoning probe tests were silently taking the reasoning-detection branch), those tests now assert the branch is not taken, and `_apply_parameter_adjustments` documents its idempotency contract since it runs before/after the probe and in `generate()`. No behaviour change.

saattrupdan (changes requested):
There is only one small correctness issue remaining; once this is fixed and covered by a regression test, the PR is good to merge.

`_apply_parameter_adjustments()` is now explicitly called multiple times and documented as idempotent, but the combination of `LOGPROBS_MUST_BE_BOOLEAN` and `NO_TOP_LOGPROBS` is order-dependent. Starting with `{"logprobs": true, "top_logprobs": 20}`, the first pass currently produces `{"logprobs": 20}` and a second pass produces `{"logprobs": true}`. The first pass can therefore send a known-invalid integer during the internal probe or actual generation through the user-override path.

Applying `NO_TOP_LOGPROBS` before Boolean normalisation makes the combined transformation idempotent:

```suggestion
        if ParameterAdjustment.NO_TOP_LOGPROBS in adjustments:
            if "top_logprobs" in generation_kwargs:
                generation_kwargs["logprobs"] = generation_kwargs.pop("top_logprobs")
        if ParameterAdjustment.LOGPROBS_MUST_BE_BOOLEAN in adjustments:
            if "logprobs" in generation_kwargs:
                generation_kwargs["logprobs"] = True
```

Please add one regression with both adjustments present that applies the method twice and asserts that both results are identical with `logprobs is True` and no `top_logprobs`.

mathiasesn:
Good catch — fixed in `370cc740`.

I applied the suggestion verbatim: `NO_TOP_LOGPROBS` now runs before `LOGPROBS_MUST_BE_BOOLEAN`, so the Boolean normalisation is the last thing to touch `logprobs` and the pair reaches its fixed point after a single pass. Added a comment at the call site noting the ordering is load-bearing.

Regression test: `TestRetryAdjustments::test_logprobs_adjustments_are_order_independent` seeds both adjustments, applies the method twice starting from `{"logprobs": True, "top_logprobs": 20}`, and asserts the two results are identical with `logprobs is True` and no `top_logprobs`. Verified it fails on the previous ordering (`assert {'logprobs': 20} == {'logprobs': True}`) and passes after the fix. Full suite green locally: 1535 passed, 172 skipped.

saattrupdan (approved):
The final idempotence fix and two-pass regression are correct. The subsequent test-only clarification strengthens the coverage. LGTM.

